PR #3, opened 2016-06-13
Foreign Endianess

flanfly:
Extends the `from_fd` functions to consider the byte order of all integers in the ELF file. Adds a dependency to the `byteorder` crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "goblin"
 version = "0.0.1"
-authors = ["m4b <m4b.github.io@gmail.com>"]
+authors = ["m4b <m4b.github.io@gmail.com>", "seu <seu@panopticon.re>"]
 readme = "README.md"
 keywords = ["binary", "elf", "mach", "pe", "cross-platform"]
 repository = "https://github.com/m4b/goblin"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "goblin"
-version = "0.0.0"
+version = "0.0.1"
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "goblin"
-version = "0.0.0"
+version = "0.0.1"
 authors = ["m4b <m4b.github.io@gmail.com>"]
```
m4b: you forgot to add your name to the authors array :)

m4b:
Wow, awesome work, thank you!

LGTM, I'm wondering how we can add a unit test in say, `program_header.rs`, for the endianness?  Or will this be too brittle, since we need to provide it an fd to some hardcoded path, which may or may not exist on some machine which is doing the testing?

m4b:
Otherwise modulo adding your name to the authors, and what you think about the unit test, I'll merge this tonight.

flanfly:
I initially had a big endinan ELF file in the tree as test vector. I decided against it, but if you're ok with binaries in the tree I can put it back in.

m4b:
Hmm, maybe we can hold off on test binaries in the tree for now.(How big is it though?)

I also worry about license issues, like where did it come from, is their accompanying source, etc, for random binaries.  I'm thinking later have a more principled approach with elf/mach/pe samples, and accompanying source, and a make command to generate them.

Anyway, @flanfly if you can add your name to the authors array, I'll merge this, and publish the newest version to cargo later tonight (or maybe next day, sorry very busy right now!).

Once again, thanks for the PR, great work! 😃

flanfly:
Done

PR #6, opened 2016-08-19
Mostly style improvements & criticism

ticki:
This tries to make the style more aligned to idiomatic Rust.

----

Very nice crate, indeed (Redox's ELF parser was getting unmaintainable anyway)!

If you don't mind, I have a some criticism, mostly idiomatic ones.

1. So, you seem to be writing Rust the same way you would write C. You would make everything much cleaner if you made more use of Rust's type system and other abstractions.

   1. Use the enum! You make use of the constant integer approach. Unfortunately, this makes code somewhat hairy. You could eliminate classes of bugs as well as writing more clean and idiomatic Rust code, if you made more use of enums. One particularly handy features is the ability to define the discriminant and then cast it to an integer (`my_enum_value as u8`). Instead of the 100 lines of constant declaration, make an enum.

   2. Newtyping. Byte strings are naturally untyped, newtyping makes sure that you do not accidentally do bad stuff. Furthermore, you are then able to define the needed methods directly on the type.

   3. Traits. So, there are a lot of places where you use standalone functions to implement functionality which could be provided by a trait implementation (for example, all of those `a_to_str` methods are unneeded in favor of the `Display` or `From` trait).

   4. Less lonely functions. Lonely functions are generally considered an anti-pattern in Rust, especially if they ties to some specific type, which it seems that yours often do.

   5. Macros. You reimplement a lot of the same stuff, which could be reduced to only a single implementation through macros.

   6. Generics. I see afew places where generics are useful.

2. Your documentation is pretty good. The only catch is the way you document it. Instead of using `//` comments, you should use `///` comments, which allows it to be rendered in the API docs.

3. Certain style choices in this crate are inconsistent with conventional Rust:

   1. I recommend running the `rustfmt` utility over the codebase, it makes the style more uniform.

   2. Some modules are unnecessarily big without any strong relation between the primitives provided. These could be broken down.

    3. `inline(always)` is generally considered bad, since `inline` gives the compiler more freedom, which means better performance.

    4. Long lists of variables. I noticed a few places where a lot of variables were declared, despite the fact that all of those could be packed into a struct. Things like `derive(Default)` are very useful here.

    5. `clippy` is fairly good at detecting anti-patterns.

I hope that wasn't too nitpicky.

Nice crate, and keep it up.

### DIFF
--- a/src/elf/_32/mod.rs
+++ b/src/elf/_32/mod.rs
@@ -12,5 +12,5 @@ pub use self::impure::*;
 
 #[cfg(not(feature = "pure"))]
 mod impure {
-    elf_from_fd!(::std::u32::MAX);
+    elf_from_fd!(!0);
 }
--- a/src/elf/mod.rs
+++ b/src/elf/mod.rs
@@ -1,36 +1,59 @@
-//! Access ELF constants, other helper functions, which are independent of ELF bithood.
-//! Also provides simple parser which returns an Elf64 or Elf32 "pre-built" binary.
-//! **WARNING**: to use the automagic ELF datatype union parser, you _must_ enable both elf and elf32 features - i.e., do not use `no_elf` **NOR** `no_elf32`, otherwise you'll get obscure errors about [goblin::elf::from_fd](fn.from_fd.html) missing.
+//! Access ELF constants, other helper functions, which are independent of ELF bithood.  Also
+//! provides simple parser which returns an Elf64 or Elf32 "pre-built" binary.
+//!
+//! **WARNING**: to use the automagic ELF datatype union parser, you _must_ enable both elf and
+//! elf32 features - i.e., do not use `no_elf` **NOR** `no_elf32`, otherwise you'll get obscure
+//! errors about [goblin::elf::from_fd](fn.from_fd.html) missing.
 
 #[cfg(not(feature = "pure"))]
 pub mod strtab;
 
-// These are shareable values for the 32/64 bit implementations
+// These are shareable values for the 32/64 bit implementations.
+//
 // They are publicly re-exported by the pub-using module
-#[macro_use] pub mod header {
-    pub const ET_NONE: u16 = 0; // No file type
-    pub const ET_REL: u16 = 1; // Relocatable file
-    pub const ET_EXEC: u16 = 2; // Executable file
-    pub const ET_DYN: u16 = 3; // Shared object file
-    pub const ET_CORE: u16 = 4; // Core file
-    pub const ET_NUM: u16 = 5; // Number of defined types
+#[macro_use]
+pub mod header {
+    /// No file type.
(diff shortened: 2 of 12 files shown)


m4b:
@ticki thanks for the PR, comments and criticisms definitely welcome!

I apologize for the delay, I'm actually AFK, and will be for a month or so, so if any new changes or PRs happen may or may not get to them.

That being said, I've looked over the changes and they look great, thanks for them!  There's only a couple places I had some issues but they're minor.

Since I don't want to block this PR since it's good and adds a lot of nice and well needed doc comments, I'm going to merge, though I might modify some things if I weren't on my phone, hehe. I am worried it could have broken dryad, but unfortunately can't test right now (also can't remember just where I had some worries, but I did have some)

As for the generic comments, I agree with most of them. I won't get into some of my responses here, but in particular the constant/enum point was something I fretted over for a while, but decided to go with constant values for the primary reason they seemed easier from an API point of view, in addition to more friendly use cases when dealing with raw headers, etc., as one often just wants to use the raw number and compare it against a constant.

Anyway, we can discuss this some other place, or maybe create an issue.

Thanks again for your work!

src/elf/_32/mod.rs
```diff
@@ -12,5 +12,5 @@ pub use self::impure::*;
 
 #[cfg(not(feature = "pure"))]
 mod impure {
-    elf_from_fd!(::std::u32::MAX);
+    elf_from_fd!(!0);
```
m4b: I don't like this personally because it seems less readable, but since it's going into a macro it's arguably not very readable in the first place.

src/elf/mod.rs
```diff
@@ -571,329 +705,344 @@ pub mod section_header {
...
-                        DT_FINI_ARRAY => fini_array = dyn.d_val.wrapping_add(bias),
-                        DT_FINI_ARRAYSZ => fini_arraysz = dyn.d_val,
-                        DT_NEEDED => needed_count += 1,
-                        DT_FLAGS => flags = dyn.d_val,
-                        DT_FLAGS_1 => flags_1 = dyn.d_val,
-                        DT_SONAME => soname = dyn.d_val,
-                        _ => (),
+            impl DynamicInfo {
+                pub fn new(dynamic: &[Dyn], bias: usize) -> DynamicInfo {
+                    let mut res = DynamicInfo::default();
```
m4b: Derp, nice :)

src/elf/mod.rs
```diff
@@ -571,329 +705,344 @@ pub mod section_header {
...
+
+                    for dyn in dynamic {
+                        match dyn.d_tag as u64 {
+                            DT_RELA => res.rela = dyn.d_val.wrapping_add(bias as _) as usize, // .rela.dyn
```
m4b: Why don't you cast the bias as `$size` like original function though?

PR #7, opened 2016-08-30
fixed some clippy warnings

llogiq:
Just a few readability nits – should be easy to review.


m4b:
Hey @llogiq thanks for the PR!

I feel like the `<'a>` annotations were added because dryad would not compile without them, but I'm probably wrong.

If so I may have to revert those changes once I'm not afk and able to test.

Anyway, LGTM, and thanks for the help!

PR #13, opened 2016-12-21
Typo prevents export of goblin::elf on 32-bit

jdub:


m4b:
Wow, wth, thanks! I swear I was just using this for 32-bit compilation a few weeks ago...

Thanks, good catch, and appreciate the PR :)

PR #19, opened 2017-03-31
RFC: safer unsafe slice casts

le-jzr:
I've noticed that almost all unsafe blocks in elf exist simply to reinterpret/cast slices. The same unsafe work with pointers and lengths is copied over all the places where this is done. To make these a little bit safer and edit-proof, I've encapsulated this slice-casting into a separate module, such that safety in the call site is assured simply because target/source type is C-style data (which can't be ensured by any Trait I'm familiar with, sadly).

While waiting for feedback, I haven't touched anything outside `elf` directory yet.


le-jzr:
Into the rabbit hole I go!
I rediscovered there are actually unsafe traits. So a trait can in fact be used to mark types for which these shenanigans are safe. Also, I realized I need explicit alignment checks for it to be really safe. And then I was like "I can generalize it to non-slice references too".
...
...
...
I'll just publish a crate for it.

src/tools.rs
```diff
@@ -0,0 +1,79 @@
...
+use core::mem::size_of;
+use core::slice::{from_raw_parts, from_raw_parts_mut};
+
+impl<S> Slice for [S] {
+    unsafe fn retype<T>(&self) -> &[T]
+        where T: Copy {
+
+        let byte_len = self.len() * size_of::<S>();
+        let new_len = byte_len / size_of::<T>();
+        from_raw_parts(self.as_ptr() as *const T, new_len)
```
m4b: When `source slice len < destination type len`, this silently drops elements, yes?

Example:

```rust
Foo {
  a: u32,
  b: u32,
}

Bar {
  a: u64,
  b: u16
}
```

Sizeof Foo is 8; sizeof Bar is 10.  Consequently, if we have `let foos: &[Foo] = [foo1, foo2]` and we do `let bars: &[Bar] = foos.retype()`, since `byte_len = 16 = 8 * 2`, `bars.len() == 1`, yes?

src/tools.rs
```diff
@@ -0,0 +1,79 @@
...
+        unsafe {
+            from_raw_parts(self.as_ptr() as *const u8, self.len() * size_of::<S>())
+        }
+    }
+    
+    unsafe fn as_mut_bytes(&mut self) -> &mut [u8] {
+        from_raw_parts_mut(self.as_mut_ptr() as *mut u8, self.len() * size_of::<S>())
+    }
+}
+
```
m4b: I think we need a couple of simple tests; I propose `retype` and `as_bytes` roundtripping validation used on `program_headers` and `section_headers`?

m4b:
@le-jzr Hehe, that's usually what happens with generics :laughing:. I set out to build a simple offset-based endian aware `pread` clone and ended up with the monstrosity of [scroll](https://github.com/m4b/scroll) (although I'm happy about it)

If you do plan on making a crate, let me know; I'm trying to keep goblin as crate-light as possible, but a good `no_std` compatible, safe byte casting crate sounds like a good thing, so I'd consider using it, but I can't make any promises.

Some hard requirements for goblin:

1. must be usable on stable rust. I'm targetting 1.15 and don't plan on bumping that anytime soon.
2. must have `no_std` capability, e.g., with something like a `std` feature flag

Some considerations:
1.  I'm hesitant to add other upstream crates because i'm not in control (:P), and they typically tend to slowly add other upstream crate dependencies, which I don't want. scroll will always be a terminal crate.
2. result based api using a custom error type and moving the asserts and other invalid states into the return type
3. alignment checking would be awesome, a la: https://github.com/m-labs/artiq/blob/0a687b7902c4fb0f27ce2c7fa63bc0b7bde9cf92/artiq/firmware/libdyld/lib.rs#L9-L38

m4b:
Oh and I should add if you do add a `Result` api, I think you can remove the `unsafe` marker from the methods, as a long as the unsafety is pushed into the error state, and proper vetting of the code (this is the hard part ;))

src/tools.rs
le-jzr: Yes indeed. I'm not really sure whether the discrepancy in length should be a concern, but I also can't think of any potential use where the types deliberately don't align currently.

We can possibly put a !debug_assert() there as a lint for possibly unintended layout errors, but any checking in release build is unnecessary as it doesn't violate Rust safety.

src/tools.rs
```diff
@@ -0,0 +1,79 @@
...
+        let new_len = byte_len / size_of::<T>();
+        from_raw_parts(self.as_ptr() as *const T, new_len)
+    }
+
+    unsafe fn retype_with_len<T>(&self, len: usize) -> &[T]
+        where T: Copy {
+
+        let byte_len = self.len() * size_of::<S>();
+        let new_len = byte_len / size_of::<T>();
+        assert!(len <= new_len);
```
le-jzr: Should it really be considered an expected error condition?
As I see it, the failure here is programmer's, and it's his responsibility to ensure that the input slice is long enough, analogously to how indexing a slice will panic if the requested index is out of range.

le-jzr: If you are concerned about `no_std` compatibility, there's no cause for worry - assert is in `core`.

m4b: No, concern is the assert failure and line number with no explanation why.

So, .e.g., you call this function with `shnum` for section headers, which comes directly from the binary. if its a bad value, incorrect, etc., it'll throw an assert instead of allowing caller to return malformed binary, etc.

I suppose its just a matter of viewpoint, but I nowadays tend to prefer explicit error returns

le-jzr:
Here you can see my initial effort: https://github.com/randomites/plain
It doesn't incorporate all the feedback yet, and it needs way more tests and documentation, but it's a start. :-P
Also, it's somewhat more flexible than the original set of functions, but this is as far as it goes. I don't think I can get any more flexible than this without changing the _purpose_ of the thing. To be continued tomorrow.

src/tools.rs
le-jzr: After a lot of thinking about possible use cases, I think I have to agree with you. I'll change it to Result<> based error handling tomorrow (it's 1am here right now :) ).

src/tools.rs
```diff
@@ -0,0 +1,79 @@
+
+pub trait Slice {
```
le-jzr: That piece of code is quite interesting, thanks! Although the functions there should still be marked unsafe, the Copy trait is not quite enough to ensure safety. Hence the unsafe trait I added to my crate (which, incidentally, would be very easy to automatically derive in compiler).

le-jzr:
Opened another pull request with a crated version.

Issue #11, opened 2016-10-16
pull core-only dependent functions and impls out of `impure`


m4b:
This will make them accessible to consumers who use don't use `std` (because why not), and will also remove the warnings for those who use `std` but not `endian_fd` (binary loaders, dryad), and will also make the api more usable in general.

m4b:
I want to do this, but I'm just not motivated.  I think we need to copy alex chrichton's approach in futures: https://github.com/alexcrichton/futures-rs/blob/3c0fbfc3b46fa0b5445459cf6b359ca39d2aaa30/src/future/mod.rs#L78-L114

i think it'll be much readable, friendly, and approachable, but i don't have the energy right now for another goblin internal's refactor

m4b:
I dunno what this issue is even about anymore, heh

PR #20, opened 2017-04-04
Replace unsafe bits using crate `plain`.

le-jzr:
Wrapped in a separate tiny crate this time, because it's just that bit of functionality that totally should be in core. It is possible to go one step further and implement #[derive(Plain)], with static checking of the prescribed requirements, but I'm not yet guru enough to do that.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,9 @@ documentation = "https://docs.rs/goblin"
 
 [lib]
 
+[dependencies]
+plain = "0.0.2"
+
 [dependencies.scroll]
 version = "0.4.0"
 default_features = false
--- a/src/elf/dyn.rs
+++ b/src/elf/dyn.rs
@@ -10,6 +10,9 @@ macro_rules! elf_dyn {
             /// Integer value
             pub d_val: $size,
         }
+
+        use plain;
+        unsafe impl plain::Plain for Dyn {}
     }
 }
 
@@ -435,6 +438,8 @@ macro_rules! elf_dyn_std_impl {
             use elf::dyn::Dyn as ElfDyn;
             use super::*;
 
+            use plain::Methods;
+
             impl From<ElfDyn> for Dyn {
                 fn from(dyn: ElfDyn) -> Self {
                     Dyn {
@@ -497,12 +502,9 @@ macro_rules! elf_dyn_std_impl {
                     if phdr.p_type == PT_DYNAMIC {
                         let filesz = phdr.p_filesz as usize;
                         let dync = filesz / SIZEOF_DYN;
-                        let mut bytes = vec![0u8; filesz];
+                        let mut dyns = vec![Dyn::default(); dync];
                         try!(fd.seek(Start(phdr.p_offset as u64)));
-                        try!(fd.read(&mut bytes));
-                        let bytes = unsafe { slice::from_raw_parts(bytes.as_ptr() as *mut Dyn, dync) };
-                        let mut dyns = Vec::with_capacity(dync);
-                        dyns.extend_from_slice(bytes);
+                        try!(fd.read(dyns.as_mut_bytes()));
                         dyns.dedup();
                         return Ok(Some(dyns));
                     }
--- a/src/elf/header.rs
+++ b/src/elf/header.rs
@@ -36,13 +36,17 @@ macro_rules! elf_header {
             /// Section header string table index
             pub e_shstrndx: u16,
         }
+
+        use plain;
+        // Declare that this is a plain type.
+        unsafe impl plain::Plain for Header {}
+
         impl Header {
             /// Returns the corresponding ELF header from the given byte array.
             pub fn from_bytes(bytes: &[u8; SIZEOF_EHDR]) -> &Header {
-                // This is not unsafe because the header's size is encoded in the function,
-                // although the header can be semantically invalid.
-                let header: &Header = unsafe { ::core::mem::transmute(bytes) };
-                header
+                // FIXME: Length is ensured correct because it's encoded in the type,
+                // but it can still panic due to invalid alignment.
+                plain::from_bytes(bytes).unwrap()
             }
         }
         impl fmt::Debug for Header {
--- a/src/elf/program_header.rs
+++ b/src/elf/program_header.rs
@@ -239,6 +239,8 @@ macro_rules! elf_program_header_std_impl { ($size:ty) => {
         use std::io::{Seek, Read};
         use std::io::SeekFrom::Start;
 
+        use plain::Methods;
+
         impl From<ProgramHeader> for ElfProgramHeader {
             fn from(ph: ProgramHeader) -> Self {
                 ElfProgramHeader {
@@ -295,11 +297,10 @@ macro_rules! elf_program_header_std_impl { ($size:ty) => {
                 Ok(program_headers)
             }
 
-
             pub fn from_bytes(bytes: &[u8], phnum: usize) -> Vec<ProgramHeader> {
-                let bytes = unsafe { slice::from_raw_parts(bytes.as_ptr() as *mut ProgramHeader, phnum) };
-                let mut phdrs = Vec::with_capacity(phnum);
-                phdrs.extend_from_slice(bytes);
+                // FIXME: will panic if length of `bytes` is not equal to phnum * SIZEOF_PHDR.
+                let mut phdrs = vec![ProgramHeader::default(); phnum];
+                phdrs.as_mut_bytes().copy_from_slice(bytes);
                 phdrs
             }
 
@@ -310,10 +311,10 @@ macro_rules! elf_program_header_std_impl { ($size:ty) => {
             }
 
             pub fn from_fd(fd: &mut File, offset: u64, count: usize) -> Result<Vec<ProgramHeader>> {
-                let mut phdrs = vec![0u8; count * SIZEOF_PHDR];
+                let mut phdrs = vec![ProgramHeader::default(); count];
                 try!(fd.seek(Start(offset)));
-                try!(fd.read(&mut phdrs));
-                Ok(ProgramHeader::from_bytes(&phdrs, count))
+                try!(fd.read(phdrs.as_mut_bytes()));
+                Ok(phdrs)
             }
         }
     }
@@ -348,6 +349,10 @@ pub mod program_header32 {
 
     pub const SIZEOF_PHDR: usize = 32;
 
+    use plain;
+    // Declare that this is a plain type.
+    unsafe impl plain::Plain for ProgramHeader {}
+
     elf_program_header_std_impl!(u32);
 }
 
@@ -379,6 +384,10 @@ pub mod program_header64 {
     }
 
     pub const SIZEOF_PHDR: usize = 56;
+
+    use plain;
+    // Declare that this is a plain type.
+    unsafe impl plain::Plain for ProgramHeader {}
 
     elf_program_header_std_impl!(u64);
 }
--- a/src/elf/reloc.rs
+++ b/src/elf/reloc.rs
@@ -71,6 +71,10 @@ macro_rules! elf_reloc {
             /// relocation type and symbol address
             pub r_info: $size,
         }
+        use plain;
+        unsafe impl plain::Plain for Rela {}
+        unsafe impl plain::Plain for Rel {}
+
         impl fmt::Debug for Rela {
             fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
                 let sym = r_sym(self.r_info);
@@ -115,6 +119,8 @@ macro_rules! elf_rela_std_impl { ($size:ident, $isize:ty) => {
             use std::fs::File;
             use std::io::{Read, Seek};
             use std::io::SeekFrom::Start;
+
+            use plain::Methods;
 
             impl From<Rela> for Reloc {
                 fn from(rela: Rela) -> Self {
@@ -183,12 +189,9 @@ macro_rules! elf_rela_std_impl { ($size:ident, $isize:ty) => {
 
             pub fn from_fd(fd: &mut File, offset: usize, size: usize) -> Result<Vec<Rela>> {
                 let count = size / SIZEOF_RELA;
-                let mut bytes = vec![0u8; size];
+                let mut relocs = vec![Rela::default(); count];
                 fd.seek(Start(offset as u64))?;
-                fd.read(&mut bytes)?;
-                let bytes = unsafe { slice::from_raw_parts(bytes.as_ptr() as *mut Rela, count) };
-                let mut relocs = Vec::with_capacity(count);
-                relocs.extend_from_slice(bytes);
+                fd.read(relocs.as_mut_bytes())?;
                 Ok(relocs)
             }
         }
--- a/src/elf/sym.rs
+++ b/src/elf/sym.rs
@@ -126,11 +126,12 @@ macro_rules! elf_sym_std_impl {
             use elf::error::*;
 
             use core::fmt;
-            use core::slice;
 
             use std::fs::File;
             use std::io::{Read, Seek};
             use std::io::SeekFrom::Start;
+
+            use plain::Methods;
 
             impl Sym {
                 /// Checks whether this `Sym` has `STB_GLOBAL`/`STB_WEAK` bind and a `st_value` of 0
@@ -186,20 +187,11 @@ macro_rules! elf_sym_std_impl {
                 }
             }
 
-            pub unsafe fn from_raw<'a>(symp: *const Sym, count: usize) -> &'a [Sym] {
-                slice::from_raw_parts(symp, count)
-            }
-
             pub fn from_fd<'a>(fd: &mut File, offset: usize, count: usize) -> Result<Vec<Sym>> {
                 // TODO: AFAIK this shouldn't work, since i pass in a byte size...
-                // FIX THIS, unecessary allocations + unsafety here
-                let mut bytes = vec![0u8; count * SIZEOF_SYM];
+                let mut syms = vec![Sym::default(); count];
                 try!(fd.seek(Start(offset as u64)));
-                try!(fd.read(&mut bytes));
-                let bytes = unsafe { slice::from_raw_parts(bytes.as_ptr() as *mut Sym, count) };
-                let mut syms = Vec::with_capacity(count);
-                syms.extend_from_slice(bytes);
-                syms.dedup();
+                try!(fd.read(syms.as_mut_bytes()));
                 Ok(syms)
             }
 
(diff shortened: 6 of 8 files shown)


src/elf/dyn.rs
```diff
@@ -10,6 +10,9 @@ macro_rules! elf_dyn {
             /// Integer value
             pub d_val: $size,
         }
+
+        use plain;
+        unsafe impl plain::Plain for Dyn {}
```
le-jzr: These unsafe impls might go away as well, if I learn how to implement derive.

src/elf/dyn.rs
```diff
@@ -497,12 +502,9 @@ macro_rules! elf_dyn_std_impl {
...
                         let dync = filesz / SIZEOF_DYN;
-                        let mut bytes = vec![0u8; filesz];
+                        let mut dyns = vec![Dyn::default(); dync];
                         try!(fd.seek(Start(phdr.p_offset as u64)));
-                        try!(fd.read(&mut bytes));
-                        let bytes = unsafe { slice::from_raw_parts(bytes.as_ptr() as *mut Dyn, dync) };
-                        let mut dyns = Vec::with_capacity(dync);
-                        dyns.extend_from_slice(bytes);
+                        try!(fd.read(dyns.as_mut_bytes()));
                         dyns.dedup();
```
le-jzr: Is this dedup() actually useful? Seems like a waste of time to me, and besides, it makes the parsed state possibly different from what's in the file.

src/elf/header.rs
```diff
@@ -36,13 +36,17 @@ macro_rules! elf_header {
...
         impl Header {
             /// Returns the corresponding ELF header from the given byte array.
             pub fn from_bytes(bytes: &[u8; SIZEOF_EHDR]) -> &Header {
-                // This is not unsafe because the header's size is encoded in the function,
-                // although the header can be semantically invalid.
-                let header: &Header = unsafe { ::core::mem::transmute(bytes) };
-                header
+                // FIXME: Length is ensured correct because it's encoded in the type,
+                // but it can still panic due to invalid alignment.
+                plain::from_bytes(bytes).unwrap()
```
le-jzr: Most things that panic would instead result in undefined behavior with original code.

src/elf/program_header.rs
```diff
@@ -295,11 +297,10 @@ macro_rules! elf_program_header_std_impl { ($size:ty) => {
...
             }
 
-
             pub fn from_bytes(bytes: &[u8], phnum: usize) -> Vec<ProgramHeader> {
-                let bytes = unsafe { slice::from_raw_parts(bytes.as_ptr() as *mut ProgramHeader, phnum) };
-                let mut phdrs = Vec::with_capacity(phnum);
-                phdrs.extend_from_slice(bytes);
+                // FIXME: will panic if length of `bytes` is not equal to phnum * SIZEOF_PHDR.
+                let mut phdrs = vec![ProgramHeader::default(); phnum];
+                phdrs.as_mut_bytes().copy_from_slice(bytes);
```
le-jzr: Can panic now. Originally, you could use this "safe" public function to read out of bounds data.
```
let b = [u8; 0];
let out_of_bounds = from_bytes(&b, 10000);
assert(out_of_bounds.len() == 10000);
```

src/elf/reloc.rs
```diff
@@ -183,12 +189,9 @@ macro_rules! elf_rela_std_impl { ($size:ident, $isize:ty) => {
 
             pub fn from_fd(fd: &mut File, offset: usize, size: usize) -> Result<Vec<Rela>> {
                 let count = size / SIZEOF_RELA;
-                let mut bytes = vec![0u8; size];
+                let mut relocs = vec![Rela::default(); count];
```
le-jzr: All these buffer allocations could be very slightly more efficient if `plain` also provided a function for allocating uninitialized vectors of `Plain` type. Probably not worth the trouble though.

src/elf/sym.rs
```diff
@@ -186,20 +187,11 @@ macro_rules! elf_sym_std_impl {
...
                 // TODO: AFAIK this shouldn't work, since i pass in a byte size...
-                // FIX THIS, unecessary allocations + unsafety here
-                let mut bytes = vec![0u8; count * SIZEOF_SYM];
+                let mut syms = vec![Sym::default(); count];
                 try!(fd.seek(Start(offset as u64)));
-                try!(fd.read(&mut bytes));
-                let bytes = unsafe { slice::from_raw_parts(bytes.as_ptr() as *mut Sym, count) };
-                let mut syms = Vec::with_capacity(count);
-                syms.extend_from_slice(bytes);
-                syms.dedup();
```
le-jzr: Oops, deleted dedup() call by accident here.

le-jzr: Also, from_raw definition above. That one was deliberate. There's no reason to provide such violently unsafe functions, especially if it's just an alias for core function.

Cargo.toml
```diff
@@ -11,6 +11,9 @@ documentation = "https://docs.rs/goblin"
 
 [lib]
 
+[dependencies]
+plain = "0.0.2"
```
le-jzr: It's no_std unconditionally, so there's no need for features.

le-jzr:
Sprinkled in some comments throughout the diff.

src/elf/reloc.rs
le-jzr: Actually, Rust prohibits any such function from being considered safe.
It's [undefined](https://doc.rust-lang.org/reference.html#behavior-considered-undefined) for any code to read uninitialized memory, even though reading an arbitrary bit pattern from `Plain` type is well-defined. So code using uninitialized vectors would always be unsafe "de jure".

le-jzr: A helper function that (unsafely) allocates the vector and instantly reads in content from another slice/file _would_ sidestep that issue.

src/elf/sym.rs
m4b: Haven't had a chance to give this a full review yet, but please don't remove API functions, e.g., this removal breaks dryad here and in 2 other places: https://github.com/m4b/dryad/blob/585189f3986c284aad52701d6de944a06192ab01/src/loader.rs#L197

Yes, I should probably write a safe wrapper for an elf symtable similar to [mach version](https://github.com/m4b/goblin/blob/8188a8a15017cbdadfdbb807e08db968f1c634c9/src/mach/symbols.rs#L164-L172),  and use this in clients, but don't have time for that right now (and its also a different PR).

src/elf/dyn.rs
m4b: So for the dynamic array, the correct way to read them is [to terminate](https://github.com/m4b/goblin/blob/8188a8a15017cbdadfdbb807e08db968f1c634c9/src/elf/dyn.rs#L379) on the first `DT_NULL` tag, which is just `0x0`, which also fills up the remainder of the dynamic array (e.g., `phdr.p_filesz`) sometimes if there is padding, so occasionally there are several `DT_NULL` entries; the `dedup` was me being lazy here, but I don't think this endpoint has been used by any downstream consumer in very likely a long time, so it doesn't really matter. I should probably just remove it next release, as it isn't a very useful function.

src/elf/program_header.rs
m4b: :sob:, good catch

src/elf/reloc.rs
m4b: This is also another dubiously useful function, a holdover before I started memmaping in dryad, iirc

src/elf/dyn.rs
m4b: You can checkout https://github.com/m4b/scroll_derive for a very simple derive implementation, but:

1. I don't know if custom derive works with `core` (probably does, don't see why not)?
2. both `syn` and `quote` will add like 5-10 seconds to compile time

I don't allow the `Pread`, `Pwrite` derives on the `no_std` portion of goblin for this reason, even though they would be useful, as the compile time is almost instantaneous now with a `elf32`, `elf64` feature flags, etc., in addition to adding 2 upstream dependencies that seem to keep profilerating (e.g., recent `syn` now adds some crate called `synom` as a dependency, etc.)

le-jzr: Another problem is that I need to inspect the definition of referenced types as well, otherwise the derive would only work for the most trivial cases. I think I'd have to use unstable rustc interfaces to do that.

m4b:
Sorry for delays, I've been quite busy; my predisposition is to merge, but definitely need tests for the apis this affects and I haven't had the chance to review the crate in detail, though you seem to be better at finding unsafe errors than me so it's probably fine :P

One thing I don't like is the `Plain/Methods` situation; is there no way to implement the methods on `Plain`, so two separate imports aren't required?

Also, FYI, and this is related to #11, this https://github.com/m4b/goblin/blob/8188a8a15017cbdadfdbb807e08db968f1c634c9/src/elf/program_header.rs#L306-L310 function is actually in a `std` cfg wrapper.

So on that note, one thing I'm extremely hesitant of is adding crates into the `core` package, as some will view it as heavyweight, since they just want the struct definitions, etc., so it might be worthwhile to keep all these methods in the `std` package.

But then again, they are very useful conversion functions, so I dunno.  I don't really do much `no_std` dev, and from what I've seen, e.g., in redox, people usually roll their own conversions, etc.

I dunno, what do you think?

le-jzr:
> One thing I don't like is the `Plain/Methods` situation; is there no way to implement the methods on `Plain`, so two separate imports aren't required?

Technically, it's possible. The thing is, I intend to ask Rust devs to add core::marker::Plain that's auto-derived by the compiler. It codifies assumptions that a LOT of unsafe code implicitly depend on for safety. In that scenario, trait Plain would get removed from the crate.

As for the latter parts, IMO if you can provide API such that nobody needs to roll their own conversions, that's absolutely the right thing to do. Every custom conversion of data is a potential bug (even a security hole, in case of unsafe code), so deduplicating them as much as possible is A Good Thing(tm).

On a related note, the current way goblin's code is laid out, it's extremely hard to tell at glance what actually gets compiled in no_std mode. I imagine this alone might prevent many people from using the library for its definitions.

le-jzr:
>> One thing I don't like is the `Plain/Methods` situation; is there no way to implement the methods on `Plain`, so two separate imports aren't required?

>Technically, it's possible.

Wait, scratch that. Actually the original reason for separation was that there is no way (that I know of) to make generic `from_bytes<T>(&[u8]) -> T` that works equally well for cases where T is sized and where T is [S] for some sized S. They need separate implementations.

Alternative solution is to provide two separate functions, e.g. from_bytes() and slice_from_bytes(). Originally, it seemed to me that from the consumer perspective, they work identically (sans generic type argument). Now I'm not that sure, since the documentation does need explicit mention of what happens with length for slices. What's your take?

m4b:
> On a related note, the current way goblin's code is laid out, it's extremely hard to tell at glance what actually gets compiled in no_std mode. I imagine this alone might prevent many people from using the library for its definitions.

Yea I'm very sensitive to this, I've been told this as the exact reason for not being used, but I also don't know what the solution is. (And am open to proposals)

I tried to make it better by removing the old hidden 32/64 modules in order to make the code itself more friendly, and thus (imho) polluting the namespace by two alternative paths to same content.

But I'm not sure what else to do. Goblin has the following important tasks:

1. Core repr C structs, tiny compile time, 32/64 (or both) at your leisure
2. Type punning. This is really important when we have repr c, and no one else does it. Dryad and redox both use it, get two completely separate compiled output without really changing anything, and no macros. See automagic.rs in examples for basic example.  
3. Std mode. This throws in read and write impls, reading from file, convenience allocations, extra methods, etc. This is for clients who can allocate and want to read files off disk. E.g. Dryad and an unnamed crate I'm working on which will output binaries. 
4. Endian_fd. A truly terrible name :laughing: this is for binary analysis like in panopticon which needs to read binaries of foreign endianness, or basis for constructing cross platform foreign architecture binutils, e.g. cargo-sym and bingrep are simple examples of this, but the sky is the limit.

(On hindsight I should put this in the readme or crate level documentation :grin:)

That being said, this is primarily an elf codebase issue, but I am not sure how to make it more friendly. Cfgs suck. What's worse is you have to resort to the private module hack if you don't want a cfg on every item. But perhaps that is better, I really don't know. 

So for someone who seems to have survived elfs goblin impl, what in terms of code structure was hardest for you (and how can we improve this)? (I personally _hate_ the macros.)

m4b:
@le-jzr I just decided to merge, can do tests later or I'll add them :) great work :+1: and thanks for the efforts!

le-jzr:
Thanks! Note that wrt the Plain/Methods separation, I decided to try splitting the function into separate slice and non-slice versions, and grudgingly acknowledge it does seem much cleaner. I'll open another pull request for your consideration once I get around to publishing a new crate version. :)

As you can see, I have a lot of free time on my hands. :D

le-jzr:
FYI, https://internals.rust-lang.org/t/automatic-marker-trait-for-unconditionally-valid-repr-c-types/5054

m4b:
:) well there's plenty to do ! Also re your crate, while you're at it, might I suggest a different name for the trait methods, perhaps like `Reinterpret` or `Cast` or less seriously `Frob`

le-jzr:
If you are thinking of the original `retype`, those are gone, in favor of `as_bytes` and `from_bytes`, since those are the principal use cases and can be trivially composed. In the next version, the plain is to split `from_bytes` into `from_bytes` (for non-slices only) and `slice_from_bytes`.

m4b:
Ah no I meant the trait name `Methods`. It's just me bikeshedding but I like seeing `use plain::Reinterpret` or something similar as it seems more meaningful but YMMV

le-jzr:
Aaaah. Sorry, slow on the uptake. :D
I'll actually just keep the `Plain` with methods, and remove `Methods`, as you suggested before. It's cleaner that way.

m4b:
Ah right, nice then won't need that at all, cool :)

m4b:
@le-jzr what happened to updating to latest plain crate ?

le-jzr:
Me being a terribly distractible person is what happened. :) I'll see to it soon(tm).

m4b:
Ftr it's stuff like this why I was hesitant to add an another upstream dependency :/

le-jzr:
The thing is, I need to make some further changes in `plain` first, and I'm unsure how to go about it.

Padding bytes being currently in this weird place where reading them may be undefined behavior, there are two distinct issues for `plain` to solve (or are there?).

One issue is converting from byte slices to structures, which is the original motivation, and depends just on the property that "all bit patterns are valid for the type". When I originally made it, I thought the opposite direction (structure to bytes) is basically free and can be included without much consideration. Then I realized padding bytes pose a second, independent issue.

So, we have two properties: `Plain`, being valid for all bit patterns, and `Gapless`, not having any implicit padding (name subject to bikeshedding). For `from_bytes()`/`from_mut_bytes()` to be safe, just `Plain` is needed. For `as_bytes`, just `Gapless` is needed.  For `as_mut_bytes()`, both `Plain` and `Gapless` are necessary for safety.

The current version conflates the two properties, which is wrong and inflexible. So my choices boil down to following possibilities.

1. I add `Gapless` trait and update the methods and functions accordingly.
2. I remove `as_bytes()` and `as_mut_bytes()` from the crate. The problem is that the latter leaves a certain gap, where it could have been used to read out a value from memory that isn't always aligned properly. That could be filled with a different method, `read_from_bytes()`, that actually returns a value instead of reference, but it wouldn't work well with slices (or there could be a second version that returns heap-allocated vector, which is even worse hit for conceptual purity).
3. I could also expand the previous point to its logical conclusion and add a full set of reading/writing functionality. Which would obviously infringe on the niche filled by `scroll`.
4. Conversely, we could expand the functionality as in previous point, but merge it into `scroll` instead of staying a separate crate. This has it's own set of disadvantages, not least of which being that the crate name is totally non-obvious and confusing. 😃 

Any thoughts?

m4b:
So `Gapless` is basically `repr(packed)` or `repr(C)` which luckily (or intentionally) happened to have as many fields with the right size as necessary so padding values aren't needed (like ELF structs)?

I don't see `as_bytes` used anywhere in goblin; if i understand correctly this is the safe variant/wrapper over transmute from `struct -> bytes`?

I considered adding native endianness "parsing" to scroll. Not sure how it would work exactly, but the idea was you'd pass a byte slice like `&[u8]` or perhaps even a raw pointer and it would safely transmute the value at the machine endianness to a reference to the struct. I thought of adding it as a `reinterpret` method to `Pread` or whatever.

`dryad` would use something like this, since it only ever uses native endianness. There are other usecases where I want to transmute instead of copying as well.

Some problems I see is that you're pushing semantic burden onto your users by requiring them to understand what Plain and Gapless _mean_, and hence when its safe to use them.  Since these are unsafe traits, I think this semantic overhead should be as low as possible, otherwise people will just impl them and then bad things will happen (I am probably one of these users cause i'm stupid + lazy + I just want to frob some damn bits).

So, for example, it's still unclear to me when one is safe to use/impl, e.g., do I have to pre-check (via println! or something) that the struct has padding added (and how do i know, short of declaring it packed?)

The `as_bytes` seems nice, but as you said, it requires a lot of thought.  Seems like you're going after memory safe, zero-copy, repr/casting/transmutes, which I do think would be a nice niche to fill, especially for severe memory limitations and/or very performance sensitive code where copying say, a record out of a binary data object which has hundreds of millions of them would be prohibitive/unnecessary (especially if you mmap it).

I don't think heap allocation or even copying is a good alternative in 2. At that point, might as well use scroll or some other deserialization method that copies.

I would tentatively be ok with merging something into scroll, like a `reinterpret` method I suggested above, but I'm not sure.  It does feel logical to me, since scroll is about reading/writing/casting bytes, etc., and it would be one less upstream dependency for goblin, which would make me happy.

This would also increase my chances of using it, since most crates I write use scroll, and then i'd have access to safe transmuting.

Dunno :man_shrugging:

le-jzr:
> So `Gapless` is basically `repr(packed)` or `repr(C)` which luckily (or intentionally) happened to have as many fields with the right size as necessary so padding values aren't needed (like ELF structs)?

Exactly. C has pretty simple rules for alignment and padding, so a user familiar with those rules would have no problem understanding it. But the semantic burden concerns me too.

I would rather use custom derive exclusively to implement those traits. For `Plain` it should be simple enough, but for `Gapless`, it would need some runtime testing as well, since type size is not a compile-time constant in Rust (yet)...

m4b:
Derive only route would be an interesting way to compile time ensure the semantics of the Gapless structed, no? Also size_of can then effectively be used as a constant and checked for conformance

le-jzr:
Still working on pushing out new version of `plain`. Sadly, illness is distracting me big time. -_-

m4b:
Oh no! Feel better and don't work too hard...

P.S. Pretty sure I won't use the derive macro because those methods in goblin exist in the no_std world, and decided no proc macros there

PR #30, opened 2017-07-29
pe: Add enough debugging support to locate PDB GUIDs

willglynn:
Contemporary PE files use external debugging information stored in [PDB files](https://github.com/Microsoft/microsoft-pdb). Since these files are separate, it's important to collate them properly, since there are few things more aggravating than debugging with data that's only _slightly_ incorrect.

PEs and PDBs both have a long history in general, and they also have a long history of different solutions to this problem. This PR adds enough support to do things [the modern way](https://llvm.org/docs/PDB/PdbStream.html): the debug table data directory points to a `ImageDebugDirectory`, which points to a `CodeviewPDB70DebugInfo`, which holds a GUID.

There's other fields in each data structure and other data structures possible via different magic numbers, but as far as I can tell, that one value retrieved that one way is really the only thing that matters since about 15 years ago. This GUID corresponds exactly to the one in [PDB stream 1](https://docs.rs/pdb/0.1.5/pdb/struct.PDBInformation.html).

### DIFF
--- a/src/pe/debug.rs
+++ b/src/pe/debug.rs
@@ -56,8 +56,8 @@ pub const IMAGE_DEBUG_TYPE_BORLAND: u32 = 9;
 impl ImageDebugDirectory {
     fn parse(bytes: &[u8], dd: &data_directories::DataDirectory, sections: &[section_table::SectionTable]) -> error::Result<Self> {
         let rva = dd.virtual_address as usize;
-        let mut offset = &mut utils::find_offset(rva, sections).unwrap();
-        let idd: Self = bytes.gread_with(offset, scroll::LE)?;
+        let offset = utils::find_offset(rva, sections).unwrap();
+        let idd: Self = bytes.pread_with(offset, scroll::LE)?;
         Ok (idd)
     }
 }
--- a/src/pe/debug.rs
+++ b/src/pe/debug.rs
@@ -1,0 +1,120 @@
+use scroll::{self, Pread};
+use error;
+
+use pe::section_table;
+use pe::utils;
+use pe::data_directories;
+
+#[derive(Debug, PartialEq, Copy, Clone, Default)]
+pub struct DebugData<'a> {
+    pub image_debug_directory: ImageDebugDirectory,
+    pub codeview_pdb70_debug_info: Option<CodeviewPDB70DebugInfo<'a>>,
+}
+
+impl<'a> DebugData<'a> {
+    pub fn parse(bytes: &'a [u8], dd: &data_directories::DataDirectory, sections: &[section_table::SectionTable]) -> error::Result<Self> {
+        let image_debug_directory = ImageDebugDirectory::parse(bytes, dd, sections)?;
+        let codeview_pdb70_debug_info = CodeviewPDB70DebugInfo::parse(bytes, &image_debug_directory)?;
+
+        Ok(DebugData{
+            image_debug_directory,
+            codeview_pdb70_debug_info
+        })
+    }
+    
+    /// Return this executable's debugging GUID, suitable for matching against a PDB file.
+    pub fn guid(&self) -> Option<[u8; 16]> {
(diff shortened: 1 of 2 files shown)


src/pe/debug.rs
```diff
@@ -0,0 +1,120 @@
...
+pub const IMAGE_DEBUG_TYPE_MISC: u32 = 4;
+pub const IMAGE_DEBUG_TYPE_EXCEPTION: u32 = 5;
+pub const IMAGE_DEBUG_TYPE_FIXUP: u32 = 6;
+pub const IMAGE_DEBUG_TYPE_BORLAND: u32 = 9;
+
+impl ImageDebugDirectory {
+    fn parse(bytes: &[u8], dd: &data_directories::DataDirectory, sections: &[section_table::SectionTable]) -> error::Result<Self> {
+        let rva = dd.virtual_address as usize;
+        let mut offset = &mut utils::find_offset(rva, sections).unwrap();
+        let idd: Self = bytes.gread_with(offset, scroll::LE)?;
```
m4b: If this doesn't need to update offset can use `pread`

src/pe/debug.rs
```diff
@@ -0,0 +1,120 @@
...
+    pub codeview_pdb70_debug_info: Option<CodeviewPDB70DebugInfo<'a>>,
+}
+
+impl<'a> DebugData<'a> {
+    pub fn parse(bytes: &'a [u8], dd: &data_directories::DataDirectory, sections: &[section_table::SectionTable]) -> error::Result<Self> {
+        let image_debug_directory = ImageDebugDirectory::parse(bytes, dd, sections)?;
+        let codeview_pdb70_debug_info = CodeviewPDB70DebugInfo::parse(bytes, &image_debug_directory)?;
+
+        Ok(DebugData{
+            image_debug_directory,
```
m4b: I was trying to keep 1.15 compat; I can't remember when shorthand initializers were introduced. If it was like 1.16 or 1.17 I'm maybe ok with bumping since that's somewhat back and the ergonomics are worth it but I did like the 1.15 commitment...

willglynn:
Testing is a great question, and yep, that's a problem with this PR as it stands right now. I can tell you that it works, but there's nothing in this PR that would ensure it stays that way.

I faced a similar problem in my `pdb` crate, where I decided to make a [`fixtures/`](https://github.com/willglynn/pdb/tree/master/fixtures/self) directory that contains a toy program in its entirety: source, makefile, and EXE/PDB build artifacts. Unit tests do their own thing on `&[u8]`s and such, but integration tests use that PDB, since the PDB format is complex enough that I found it impractical to construct by hand. A single executable likely doesn't go far enough in terms of coverage, but it's a start.

Another approach I've used is to have CI retrieve a test corpus separate from the repository itself. That scales to ludicrous sizes really well -- nobody likes having GBs of cruft in `.git`, and this way you don't need it. This is good at keeping rare regressions away, especially if checking for those regressions requires a large amount of input data that you could easily collect, but it kind of sucks to push something that passed `cargo test` only to have it break to a test you didn't have the ability to run.

One project involved a combination of both approaches: CI would test against a giant corpus, and if it caught something not caught by the other tests, you could run a command to pull a single record out of the CI corpus and add it to the repo. That worked great there, since the equivalent of `cargo test` actually had good coverage and you could easily expand it as needed, but individual tests were a lot smaller than `goblin` executables so it probably doesn't help here.

src/pe/debug.rs
willglynn: Looks like that's 1.17: https://github.com/rust-lang/rust/commit/1b6f1fe9dd29c97fca66a7377f367d0beb327b58

I'll fix.

src/pe/debug.rs
willglynn: Good point. Fixed.

m4b:
LGTM! Thanks so much for this, really awesome work :) :heart:

PR #31, opened 2017-07-29
Update plain to latest version.

le-jzr:
I'm not happy that some unsafe blocks returned, but the API wasn't entirely safe to begin with, so I figure this is better than just sitting on my frustration with the memory model.

Though it would be even better to implement _proper_ reading from files, endianness and all.

### DIFF
--- a/src/elf/reloc.rs
+++ b/src/elf/reloc.rs
@@ -120,8 +120,6 @@ macro_rules! elf_rela_std_impl { ($size:ident, $isize:ty) => {
             use std::io::{Read, Seek};
             use std::io::SeekFrom::Start;
 
-            use plain::Methods;
-
             impl From<Rela> for Reloc {
                 fn from(rela: Rela) -> Self {
                     Reloc {
@@ -191,7 +189,9 @@ macro_rules! elf_rela_std_impl { ($size:ident, $isize:ty) => {
                 let count = size / SIZEOF_RELA;
                 let mut relocs = vec![Rela::default(); count];
                 fd.seek(Start(offset as u64))?;
-                fd.read(relocs.as_mut_bytes())?;
+                unsafe {
+                    fd.read(plain::as_mut_bytes(&mut *relocs))?;
+                }
                 Ok(relocs)
             }
         }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ categories = ["parsing"]
 [lib]
 
 [dependencies]
-plain = "0.0.2"
+plain = "0.2.1"
 
 [dependencies.scroll]
 git = "https://github.com/m4b/scroll"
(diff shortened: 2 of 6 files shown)


src/elf/reloc.rs
```diff
@@ -191,7 +189,9 @@ macro_rules! elf_rela_std_impl { ($size:ident, $isize:ty) => {
                 let count = size / SIZEOF_RELA;
                 let mut relocs = vec![Rela::default(); count];
                 fd.seek(Start(offset as u64))?;
-                fd.read(relocs.as_mut_bytes())?;
+                unsafe {
+                    fd.read(plain::as_mut_bytes(&mut *relocs))?;
```
m4b: The unsafety is technically in the as_mut_bytes so this is slightly over estimated, but I don't care. 

But why is as_mut_bytes unsafe now ?

m4b:
Yea I mean this part of the API is really old, before scroll and etc. I'd be ok with reading the stuff out using scroll but I'm also not too concerned about switching to it. Also there would probably be a perf hit using scroll instead of byte casts

src/elf/reloc.rs
le-jzr: A bit overestimated, yeah. That's pure aesthetic choice on my part.

The deal with `as_mut_bytes()` is that reading structure padding bytes may be undefined behavior in some cases. It's safe if the structure has no implicit padding, or if you only write without reading, but trying to encode that into the type system is too much trouble.

le-jzr:
I figure I could do a library that uses casting in the case data endianness = host endianness, and `byteorder` crate otherwise. That might be an interesting project. :3

m4b:
Yea that would be ideal, but it's not immediately clear to me how to do runtime switching. Maybe you could hack it by check if a global const is in big endianness or little endianness and if different use scroll (not byteorder :laughing:), and cast otherwise ?

le-jzr:
I figure it would be something like
```
if file.little_endian() {
    file.parse::<Endian::Little>()
} else {
    file.parse::<Endian::Big>()
}
```
but under the hood, the library would instead be specialized for `Endian::Native` vs. `Endian::NonNative`, with `Little` and `Big` being conditional aliases. Or something.

m4b:
I'm not sure how that would work. I think checking a constant's repr similar to macho's FEEDFACE is the easiest way to do it (at least on stable). 

Also the more I think about this the more I want it, but I don't know how it would work. Either you copy (in which case endianness conversions aren't that big a deal (ok for most use cases)) or you get a reference. But a reference to "to be endian converted" struct doesn't make sense. 

Specifically I think it would be really awesome to pread out a reference to a struct. In particular, when I mmap something and I need maximum speed and memory constraints I want to only have references to the backing same endianness struct, and not to copy. 

But I don't see how you could combine the two. 

I would however be interested in putting pread_ref or something into scroll which does do an explicit (but safe) bit cast. 

Maybe call it reinterpret a la c++ but safe?

Issue #27, opened 2017-07-03
Tracking issue for fuzzing


m4b:
Many overflow issues should disappear once upgraded to latest scroll

* [ ] With seed 4, I get another overflow at src/strtab.rs:37. (To reproduce, be sure to delete fuzz/corpus.)
* [ ] Seed 10 gives index out of bounds at src/elf/mod.rs:205. (Corrupt sh_link.)

/cc @sanxiyn

sanxiyn:
Seed 41 gives index out of bounds at src/elf/dyn.rs:402. (DT_NEEDED.)

I systematically ran 1 million runs each for seed from 1 to 100.

* 56 successes
* 34 failures
  * 23 in strtab
  * 7 in sh_link
  * 4 in DT_NEEDED
* 9 out-of-memory
* 1 AddressSanitizer failure

At this point, it is probably necessary to fix above 3 issues before finding others.

m4b:
This is really great, thanks for doing this ! :) :+1: 

Unfortunately, these fixes are blocked until latest version of scroll lands. In meantime, good to note them here; eager to see what else is in store once the scroll update lands :)

m4b:
Ok, the scroll port/branch is now on master, and so can fuzz/report bugs from there!

sanxiyn:
I re-ran fuzzing on 9f405aa, and the result is:

* 65 successes
* 11 out-of-memory
* 19 strtab (seed 2)
* 5 sh_link (seed 64)

DT_NEEDED seems fixed.

To reproduce with seed X, run `rm -rf fuzz/{artifacts,corpus}; RUST_BACKTRACE=1 cargo fuzz run parse -- -seed=X`.

sanxiyn:
With #34 merged, 71 successes, 20 out-of-memory, 9 sh_link (seed 1). Hm, I hoped to discover some new issues, but no dice. Maybe I should try something new.

sanxiyn:
With #35 merged, I am happy to report that goblin withstood 100 million fuzzing runs, 1 million runs each for for seed 1~100.

Please close this issue. I will open the new one when I discover more issues by fuzzing differently. First thing to try is adding corpus, instead of starting from zero.

m4b:
:+1: thanks for all your investigations, this was epic! :D

PR #35, opened 2017-08-02
Check sh_link is in bounds

sanxiyn:
cc #27.

### DIFF
--- a/src/elf/mod.rs
+++ b/src/elf/mod.rs
@@ -187,13 +187,18 @@ mod impure {
 
             let section_headers = SectionHeader::parse(bytes, header.e_shoff as usize, header.e_shnum as usize, ctx)?;
 
-            let strtab_idx = header.e_shstrndx as usize;
-            let shdr_strtab = if strtab_idx >= section_headers.len() {
-                Strtab::default()
-            } else {
-                let shdr = &section_headers[strtab_idx];
-                try!(Strtab::parse(bytes, shdr.sh_offset as usize, shdr.sh_size as usize, 0x0))
+            let get_strtab = |section_headers: &[SectionHeader], section_idx: usize| {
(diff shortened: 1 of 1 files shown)


src/elf/mod.rs
```diff
@@ -187,23 +187,26 @@ mod impure {
...
 
-            let strtab_idx = header.e_shstrndx as usize;
-            let shdr_strtab = if strtab_idx >= section_headers.len() {
-                Strtab::default()
-            } else {
-                let shdr = &section_headers[strtab_idx];
-                try!(Strtab::parse(bytes, shdr.sh_offset as usize, shdr.sh_size as usize, 0x0))
+            let get_strtab = |section_headers: &[SectionHeader], section_idx: usize| {
+                if section_idx >= section_headers.len() {
+                    Ok(Strtab::default())
```
m4b: Can you add a `FIXME: warn` comment so when logging goes in this, this will get added?

sanxiyn: Done.

m4b: Hmm I've been thinking about this and it might be better to return an error here, or log, I'm leaning towards latter but:

My reasoning is that it's a pretty serious error if the section index is oob, and returning a default strtab will just return empty strings for any references, further exacerbating debuggging. 

On the other hand the program is still technically executable without any errors whatsoever, so it might be worth parsing and warning.

What do you think ?

m4b:
Meh I'll just warn. I think we'll make the general rule we'll try to parse broken binaries if they're still technically executable. Or something like that ;)

Anyway, thanks again for another great patch ! Soon won't be anything left to fix :laughing: and it'll be all done!

sanxiyn:
Yeah, "is it executable?" is pretty much what I considered as the line in the sand.

PR #41, opened 2017-08-08
Fix mach::MultiArch::get() return value lifetime

willglynn:
`mach::MultiArch::get()`'s return value was missing a lifetime bound.


m4b:
Awesome thanks !

Issue #26, opened 2017-07-03
archive: needs to parse osx/bsd random archives


m4b:
It can parse them without dying but the filename delimiters are different (uses `#`), so it doesn't really parse them at all.

See https://en.wikipedia.org/wiki/Ar_(Unix)#BSD_variant

willglynn:
TIL: `ar` is alive and well on MacOS X. Posting for posterity:

I had thought all recent `.a`s would be Mach-O since the need to support universal binaries over a decade ago. There was a push to replace `ar` with `ranlib` way back in the day, and another to replace `ranlib` with `libtool -static` more recently. Everything besides `ar` produces something that looks like any other Mach-O object file, so there shouldn't be too many `ar` libraries in the wild these days.

Nope, I was wrong. `/usr/lib/*.a` on a current system has plenty of `ar` libraries. It also has something I didn't expect to see at all -- `/usr/lib/libkmod.a` is a Mach-O fat header containing `ar`s for three architectures.

I wanted to believe that `ar` is dead on OSX, but alas 😞

m4b:
Yea, so current archive parser will actually parse the archive "correctly", but doesn't understand the invented indexing "semantics" that the various archive formats ad hoc-ly bolted onto archives.

So easy way to experiment with this is `cargo build --target=x86_64-apple-darwin`, and you can example the resulting `libgoblin.rlib` which LLVM emits, which is the BSD style ar :/

PR #42, opened 2017-08-13
Fix Mach binds

willglynn:
This PR fixes a bunch of issues I had with `mach::BindInterpreter`, including:

* `BIND_OPCODE_DO_BIND*` opcodes all bound after mutating `BindInformation` instead of binding and then mutating,
* `BIND_OPCODE_DO_BIND_ULEB_TIMES_SKIPPING_ULEB` bound only once instead of binding `uleb` times,
* `BindInterpreter::run()` inverted its `is_lazy` condition, processing normal binds when `is_lazy = true` and lazy binds when `is_lazy = false`, and
* Bind addend information was parsed but not exposed through `Import`

This PR also adds `examples/dyldinfo.rs`. It supports only `dyldinfo -bind <mach object>`, but with these changes, it produces output that's identical to Apple's `dyldinfo -bind <mach object>` for the handful of executables I've tried.


willglynn:
I added a `tests/compare_dyldinfos.rs` integration test that compares the output of `/Library/Developer/CommandLineTools/usr/bin/dyldinfo` against `cargo run --example dyldinfo` for various inputs. This test needs MacOS X to function, so I added it to the Travis build matrix.

I also skipped a failing-on-MacOS test on `tests/archive.rs` (#26) so Travis can go green.

willglynn:
I added `dyldinfo -lazy_bind`.

This required a little bit of support code in the library to expose a value which the lazy bind table calls `index`, which I called `Import::start_of_sequence_offset`. I don't see that this value is particularly useful, but this lets the integration test verify that lazy binds are being parsed identically too, and the extra test coverage is probably worth it.

I think I'm done tinkering with `dyldinfo` for now :-)

src/mach/imports.rs
```diff
@@ -195,9 +216,9 @@ impl<'a> BindInterpreter<'a> {
 	            // throwBadBindingAddress(address, segmentEndAddress, segmentIndex, start, end, p);
 	            // (this->*handler)(context, address, type, symbolName, symboFlags, addend, libraryOrdinal, "", &last);
 	            // address += sizeof(intptr_t);
+                    imports.push(Import::new(&bind_info, libs, segments, start_of_sequence));
                     let seg_offset = bind_info.seg_offset.wrapping_add(ctx.size() as u64);
                     bind_info.seg_offset = seg_offset;
-                    imports.push(Import::new(&bind_info, libs, segments));
```
m4b: I don't understand, how could this change be valid?

They're completely semantically different if you push before or after; if this is correct, how was import parser working at all?

willglynn: I don't think it was working. The comment says:

```rust
// we bind the old record, then increment bind info address for the next guy, plus the ptr offset *)
```

The comment matches [the behavior in `dyld`](https://github.com/opensource-apple/dyld/blob/master/src/ImageLoaderMachOCompressed.cpp#L999-L1004):

```c
case BIND_OPCODE_DO_BIND:
	if ( address >= segmentEndAddress ) 
		throwBadBindingAddress(address, segmentEndAddress, segmentIndex, start, end, p);
	(this->*handler)(context, address, type, symbolName, symboFlags, addend, libraryOrdinal, "", &last, false);
	address += sizeof(intptr_t);
	break;
```

Bind then mutate. The code had been doing the opposite.

willglynn: Oh! As for how this could go unnoticed, consider a typical opcode sequence:

```0x0178 BIND_OPCODE_DO_BIND()
0x0179 BIND_OPCODE_SET_SYMBOL_TRAILING_FLAGS_IMM(0x00, __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info)
0x01B9 BIND_OPCODE_ADD_ADDR_ULEB(0xFFDB7F50)
0x01C4 BIND_OPCODE_DO_BIND_ADD_ADDR_ULEB(0x00168778)
0x01C8 BIND_OPCODE_DO_BIND_ADD_ADDR_ULEB(0x00000090)
0x01CB BIND_OPCODE_DO_BIND_ADD_ADDR_ULEB(0x0009A908)
0x01CF BIND_OPCODE_DO_BIND_ADD_ADDR_ULEB(0x000000E0)
0x01D2 BIND_OPCODE_DO_BIND_ADD_ADDR_ULEB(0x000000B8)
0x01D5 BIND_OPCODE_DO_BIND_ADD_ADDR_IMM_SCALED(0x00000060)
0x01D6 BIND_OPCODE_DO_BIND_ADD_ADDR_ULEB(0x000299E8)
0x01DA BIND_OPCODE_DO_BIND_ADD_ADDR_ULEB(0x000277A8)
0x01DE BIND_OPCODE_DO_BIND_ADD_ADDR_ULEB(0x00000BC0)
0x01E1 BIND_OPCODE_DO_BIND_ADD_ADDR_ULEB(0x00000E00)
0x01E4 BIND_OPCODE_DO_BIND_ULEB_TIMES_SKIPPING_ULEB(3, 0x00000030)
```

Bind sequences typically look like "set up stuff, bind, bind, bind, bind, bind", where each bind operation binds and moves the cursor. Binding at the start versus binding at the end is basically an off-by-one error -- the first bind gets missed and an errant bind gets added at the end, but most of the binds in the middle get covered correctly.

m4b: Yea, but wouldn't this mean every single binding has substantially incorrect values; pointing to different libraries, different addresses, etc. I just feel like this would have shown up as very very broken, if it parsed at all.  Dunno.  I guess testing this against the canonical dyldinfo util on osx is just the correct way to check regressions, thanks for adding that!

willglynn: Completing the example, `goblin` as of eab5060f39e88cc4e181afec3f71743c2fdb384f agrees with `dyldinfo -bind`, both showing:

```__DATA  __const          0x102EE7168    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __const          0x10304F8E8    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __const          0x10304F980    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __const          0x1030EA290    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __const          0x1030EA378    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __const          0x1030EA438    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __const          0x1030EA498    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __const          0x103113E88    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __data           0x10313B638    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __data           0x10313C200    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __data           0x10313D008    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __data           0x10313D040    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __data           0x10313D078    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __data           0x10313D0B0    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
```

`goblin` as of 150f92e590d697a0305fac3f39d2a2db79a00f56 shows:

```__DATA  __const          0x10304F8E8    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __const          0x10304F980    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __const          0x1030EA290    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __const          0x1030EA378    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __const          0x1030EA438    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __const          0x1030EA498    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __const          0x103113E88    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __data           0x10313B638    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __data           0x10313C200    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __data           0x10313D008    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __data           0x10313D0B0    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
__DATA  __data           0x10313D0B8    pointer      0 libc++           __ZNKSt3__119__shared_weak_count13__get_deleterERKSt9type_info
```

…which is a combination of the off-by-one I described and the missing repeated binds.

m4b: Yea wow this was horrible... thanks again for fixing this!

I've been testing on `libssl.dylib` (64 bit version sucked out), and, for example, old version gives wrong address for `_time`, so when implementing this line https://github.com/das-labor/panopticon/pull/321/commits/597b1d581f3272338231950887c1b03041c9ff1f#diff-4c438f60247f494e2aee67fcce9e11e0R112

since i was receiving lies from goblin and consequently it says that nothing calls `_time`; but newer fixed version I believe outputs correct values (can't check with a debugger though, since not on an osx system)

willglynn: I fell into this rabbit hole while building a tool that uses `goblin` to explore [C++ RTTI structs](http://www.swag.uwaterloo.ca/acd/docs/ItaniumC++ABI.htm#rtti-layout). I got it working on statically linked binaries, but it was failing on Mac dynamically linked ones.

The RTTI structs have different data members depending on their type, and their type can be distinguished based on their vtable pointers. The failing types should have been importing the vtable of e.g. `abi::__class_type_info` from `libstdc++` -- the imports just weren't there in `goblin`, but I was seeing them at runtime and in a disassembler, which pointed to a `goblin` issue.

That finding prompted me to build a tool to dump Mach-O imports from `goblin`. This tool let me bug-hunt this issue, which ultimately turned into what landed in this PR :-)

Incidentally, `goblin::mach` vtable imports were slightly broken for a separate reason: C++ vtable pointers need to point [after the typeinfo pointer](http://www.swag.uwaterloo.ca/acd/docs/ItaniumC++ABI.htm#vtable), while on Mac the vtable symbols point at the start of the table. The `addend` field accounts for that offset. This differs from e.g. Windows where the vtable symbols point directly to where the vtable pointers need to point, even though those addresses may literally be past the end of the vtable.

Issue #29, opened 2017-07-29
Mach-O: improve entrypoint handling


willglynn:
Comparing `goblin`'s implementation and my current working memory of `dyld`, I have two observations:

1. `LC_MAIN` and `LC_UNIXTHREAD` both provide entrypoint locations, but the environments provided by each are not interchangeable. (See [`dyld.cpp`](https://github.com/opensource-apple/dyld/blob/master/src/dyld.cpp#L4955-L4968), [`dyldStartup.s`](https://github.com/opensource-apple/dyld/blob/master/src/dyldStartup.s#L172-L191).) Last time I needed to know the entrypoint of a Mach-O executable, I had to know which kind it was; `struct MachO` should probably retain this distinction.
2. All those registers in the `LC_UNIXTHREAD` thread state? You know, the arch-specific thread states which are not currently handled by `goblin`? Good news: they don't matter in the slightest. [`dyld` uses only the instruction pointer](https://github.com/opensource-apple/dyld/blob/3f928f32597888c5eac6003b9199d972d49857b5/src/ImageLoaderMachO.cpp#L1061-L1095), and the rest are entirely discarded. This makes them unusable in practice, and thus they're always zero.

m4b:
I like to pretend that `LC_UNIXTHREAD` doesn't exist :hearnoevil:

Second point is good to know, iiuc, this is the current behavior and nothing needs to change ?

I'm not sure I understand the first point completely though? The address in either case is a vm address, iirc, and it shouldn't matter how I arrived at the value, no?

Especially since I think almost all OS X binaries are all PIC and all dynamically linked, unless you happen to have a very old version, but even then I'm not sure it's  too important ?

Would have to see an older binary, likely 10.1 to see if this is an issue

willglynn:
The second point means `goblin` doesn't need to have a ton of different thread state structs. Given that only one register matters, all the arch-specific thread state structs can be boiled down to a single `u64`. `goblin` still needs to know how to find that value for each architecture since the offset and size vary, but the `LC_UNIXTHREAD`-encapsulating struct doesn't need to remember AVX/MMX registers on x86 separately from ARM and PPC since `dyld` ignores them anyway.

[...]

PR #44, opened 2017-08-14
Re-work Mach entrypoints

willglynn:
This is a stab at #29. Highlights:

* `ThreadCommand` now holds a `pub thread_state: &'a [u8]`. This required plumbing `<'a>` down through `CommandVariant` and `LoadCommand`, but should permit `ThreadCommand` to be round-tripped in a way that's faithful to the original bits.
* `ThreadCommand::instruction_pointer(cputype: CpuType)` can fish out a `u64` from `thread_state`. Unfortunately, the `flavor` in the load command doesn't actually specify anything useful here -- we really do need the `cputype` from the Mach header to fish out the right register. I've verified that it works for i386 and x86-64 binaries, and I wrote handlers for ARM though I don't have any binaries with which to test.
* The earlier issue discussed making the `LC_MAIN`/`LC_UNIXTHREAD` distinction a flag, but while implementing this, I realized that the commands actually describe two different kinds of values. I made `pub entry: u64` into `pub entry: Entrypoint`, which is one of `Entrypoint::None`, `Entrypoint::Offset(u64)`, or `Entrypoint::VirtualAddress(u64)`.

Also included: `Debug` output for Mac was printing `&[u8]`s containing the data of each section as well as the data of the entire `MachO`. This made `cargo run --example rdr` difficult to use.

### DIFF
--- a/src/mach/load_command.rs
+++ b/src/mach/load_command.rs
@@ -371,43 +371,136 @@ pub const SIZEOF_DYLINKER_COMMAND: usize = 12;
 // struct XXX_thread_state state   thread state for this flavor
 // ...
 #[repr(C)]
-#[derive(Debug, Clone, Copy, Pread, Pwrite, SizeWith)]
-pub struct ThreadCommand {
+#[derive(Debug, Copy, Clone)]
+pub struct ThreadCommand<'a> {
     /// LC_THREAD or  LC_UNIXTHREAD
     pub cmd:     u32,
     /// total size of this command
     pub cmdsize: u32,
     pub flavor: u32,
-    pub count: u32,
-    /// NOTE: this is actually, in classic mach-o style, a semi-tagged union-esque struct, and is _not_ properly handled here for arches other than i386... e.g., this is incorrect for powerpc, armv7, etc.
-    /// TODO: We need to implement a simple getter for the thread state; or even better, a method that simply returns the entry point (which is what we usually want)
-    pub thread_state: I386ThreadState,
+
+    /// The raw thread state, details of which varies by CPU
+    pub thread_state: &'a [u8],
 }
 
-/// Main thread state consists of
-/// general registers, segment registers,
-/// eip and eflags.
-///
-#[repr(packed)]
-#[derive(Debug, Clone, Copy, Pread, Pwrite, SizeWith)]
-pub struct I386ThreadState {
-    pub eax: u32,
-    pub ebx: u32,
-    pub ecx: u32,
-    pub edx: u32,
-    pub edi: u32,
-    pub esi: u32,
-    pub ebp: u32,
-    pub esp: u32,
-    pub ss: u32,
-    pub eflags: u32,
-    pub eip: u32,
-    pub cs: u32,
-    pub ds: u32,
-    pub es: u32,
-    pub fs: u32,
-    pub gs: u32,
+impl<'a> ThreadCommand<'a> {
+    pub fn instruction_pointer(&self, cputype: super::cputype::CpuType) -> error::Result<u64> {
+        // The thread command includes a `flavor` value which distinguishes between related thread
+        // states. However, `dyld` ignores this entirely, blindly interpreting the thread state
+        // values as a machine-specific set of registers matching the build configuration of the
+        // active `dyld` binary.
+        //
+        // Really the only thing that `dyld` cares is that the Mach header's `cputype`, so that's
+        // what we use here.
+        use scroll::{Pread,LE};
+        match cputype {
+            super::cputype::CPU_TYPE_X86 => {
+                // struct i386_thread_state_t {
+                //   uint32_t eax;
+                //   uint32_t ebx;
+                //   uint32_t ecx;
+                //   uint32_t edx;
+                //   uint32_t edi;
+                //   uint32_t esi;
+                //   uint32_t ebp;
+                //   uint32_t esp;
+                //   uint32_t ss;
+                //   uint32_t eflags;
+                //   uint32_t eip;
+                //   uint32_t cs;
+                //   uint32_t ds;
+                //   uint32_t es;
+                //   uint32_t fs;
+                //   uint32_t gs;
+                // }
+                let eip: u32 = self.thread_state.pread_with(10 * 4, LE)?;
+                Ok(eip as u64)
+            },
+            super::cputype::CPU_TYPE_X86_64 => {
+                // struct x86_thread_state64_t {
+                //   uint64_t rax;
+                //   uint64_t rbx;
+                //   uint64_t rcx;
+                //   uint64_t rdx;
+                //   uint64_t rdi;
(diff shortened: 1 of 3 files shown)


src/mach/load_command.rs
```diff
@@ -371,44 +371,137 @@ pub const SIZEOF_DYLINKER_COMMAND: usize = 12;
 // struct XXX_thread_state state   thread state for this flavor
 // ...
 #[repr(C)]
-#[derive(Debug, Clone, Copy, Pread, Pwrite, SizeWith)]
-pub struct ThreadCommand {
+#[derive(Debug, Copy, Clone)]
+pub struct ThreadCommand<'a> {
```
m4b: I don't like the lifetime here at all.

So I just got done removing it from Section, which was a mistake in the first place, and I'm going to try to remove it from Segment (probably not tho).

The reason is that it makes construction of the load command more difficult (the bytes can't be written as a stream).

Also, the implementation for pwrite will be complicated.

It means that someone has to write the bytes for you, and insert them into the threadcommand (e.g., someone pre-marshalled the bytes in the appropriate endianness, and type (this last part not so relevant for threadcommand, but was for segment/section)).

Then you have to decide who writes the bytes in the struct, either the top reference of the bytes, or the threadcommand pwrite impl.

Nevertheless, if you think it's worthwhile to include the bytes, and don't see a better way (or if its not worth the pain), i'm fine with it.  I don't think I'll be actually ever be writing this load command since it would use the more modern one.

src/mach/mod.rs
```diff
@@ -24,21 +24,43 @@ pub fn peek(bytes: &[u8], offset: usize) -> error::Result<u32> {
     Ok(bytes.pread_with::<u32>(offset, scroll::BE)?)
 }
 
-#[derive(Debug)]
+#[derive(Copy,Clone,Eq,PartialEq,Hash)]
+pub enum Entrypoint {
+    None,
+    /// The entrypoint was from an `LC_MAIN` load command (added in 10.8) and specifies an offset in
+    /// the `__TEXT` segment.
+    Offset(u64),
```
m4b: Heh you keep finding bugs :P

I realized I've never loaded an executable (and for dylibs its always 0)

So I'm wondering if we should relocate the offset based on the vmaddress in the text.

I think the formula is:

`(text_vmaddr - text_offset) + entry_offset`

Actually I just realized after typing this the whole point of your PR was you wanted to distinguish the entrypoint case.

Is this actually super important; e.g., could we just relocate it to be a vm, so that it matches more closely both the PE and ELF entry field?

src/mach/mod.rs
```diff
@@ -51,6 +73,25 @@ pub struct MachO<'a> {
...
+            .field("header",        &self.header)
+            .field("load_commands", &self.load_commands)
+            .field("segments",      &self.segments)
+            .field("entry",         &self.entry)
+            .field("libs",          &self.libs)
+            .field("name",          &self.name)
+            .field("little_endian", &self.little_endian)
+            .field("is_64",         &self.is_64)
+            .field("symbols",       &self.symbols().collect::<Vec<_>>())
+            .field("exports",       &self.exports().unwrap())
```
m4b: I'm trying to reduce profliferation of unwraps (and i know i'm not great with this); in this case, let's do `unwrap_or()` and try to return the error so users can see

src/mach/segment.rs
```diff
@@ -328,7 +328,8 @@ impl<'a> fmt::Debug for Segment<'a> {
             .field("nsects", &self.nsects)
             .field("flags", &self.flags)
             .field("data", &self.data.len())
-            .field("sections", &self.sections().unwrap())
+            .field("sections", &self.sections().unwrap()
```
m4b: Ditto with `unwrap()`

src/mach/load_command.rs
willglynn: I saw these constraints:

1. `impl TryFromCtx for ThreadCommand` doesn't know `cputype`
2. `ThreadCommand` needs to be `Copy`
3. A `pwrite` implementation for `ThreadCommand` should be able to reproduce arbitrary thread states

A `&[u8]` fit the bill, but I figured it would be controversial. If we relax the last constraint from "arbitrary thread states" to "thread states corresponding to one of these four structs", there's a lot of other options.

* We could do some type level stuff to make `ThreadCommand` generic over `cputype`. That type parameter would have to bubble up to `LoadCommand`, which complicates things, but it would mean that `ThreadCommand` could directly contain a different machine-specific struct describing the register state.
* We could treat machine-specific thread states as an untagged union. That avoids the type parameter, but it bumps the minimum to Rust 1.19. It also wouldn't work for big-endian host machines, so nevermind.
* We could treat machine-specific thread states as a series of separate structs, and just… read into all of them. That grows `ThreadCommand` to ~800 bytes and complicates writing (which struct do we write?) but it would solve the reading problem.
* We could treat thread state as an opaque array, say a `[u32; 70]`, and worry about extracting meaningful data ourselves.

Having written all that out, the `[u32; 70]` approach seems least invasive and has the fewest downsides, so I'll go with that.

src/mach/mod.rs
willglynn: > So I'm wondering if we should relocate the offset based on the vmaddress in the text.

Sure, that makes sense. If `pub entry: u64` is a virtual memory address everywhere else, it can be here too.

> I think the formula is:
>
> `(text_vmaddr - text_offset) + entry_offset`

Yep. Things get more complicated if you're in a context where the image has actually been loaded -- you have to calculate and care about the `slide` of the image -- but this formula will at least be consistent with all the other `vmaddr`s that `goblin` exposes.

At that point we can drop the `enum` and go to a boolean flag :-)

src/mach/mod.rs
willglynn: Hmm… these don't need to be `unwrap()` at all. It's `Debug` output; how would you feel about printing `Ok(…)` and `Err(…)` here instead of returning an error?

willglynn:
Updates:

* `ThreadCommand` now holds `pub thread_state: [u32; 70]`, which is big enough to hold all the thread states that `goblin` actually implements. `<'a>` is no longer plumbed all the way down. There's some silly boilerplate because `[T; 70]` is missing some trait implementations that other array sizes get by default.
* `LC_MAIN` is now mapped to a virtual address, so `pub entry: u64` is back. I added `pub old_style_entry: bool` to distinguish `LC_MAIN` from `LC_UNIXTHREAD`. I've manually validated a matrix of {`LC_MAIN`, `LC_UNIXTHREAD`} on { i386, x86_64 } executables against other tools but it would be great to have proper test coverage here.
* `Debug` output no longer panics if e.g. `exports()` fails. It now unconditional prints the `Result`, as in `exports(): Ok(…)` or `exports(): Err(…)`.

m4b:
This is awesome, great work once again!

PR #46, opened 2017-08-17
Remove the last Strings from Archive

### LINKED ISSUE #16: zero-copy archive
see #14 

This might be more tedious.

Also should revert the archive header name back to original byte array, to keep `repr(C)`, which should allow for implementation to be `derive(Pwrite, Pread)`

willglynn:
I think this closes #16. This commit makes `archive/mod.rs` no longer contain any `String`s.

### DIFF
--- a/src/strtab.rs
+++ b/src/strtab.rs
@@ -93,6 +93,9 @@ impl<'a> Index<usize> for Strtab<'a> {
     /// **NB**: this will panic if the underlying bytes are not valid utf8, or the offset is invalid
     #[inline(always)]
     fn index(&self, offset: usize) -> &Self::Output {
+        // This can't delegate to get() because get() requires #[cfg(features = "std")]
+        // It's also slightly less useful than get() because the lifetime -- specified by the Index
+        // trait -- matches &self, even though we could return &'a instead
         get_str(offset, &self.bytes, self.delim).unwrap()
     }
 }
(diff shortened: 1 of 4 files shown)


src/strtab.rs
```diff
@@ -93,7 +93,7 @@ impl<'a> Index<usize> for Strtab<'a> {
     /// **NB**: this will panic if the underlying bytes are not valid utf8, or the offset is invalid
     #[inline(always)]
     fn index(&self, offset: usize) -> &Self::Output {
-        get_str(offset, &self.bytes, self.delim).unwrap()
+        self.get(offset).unwrap().unwrap()
```
m4b: Just curious why did this change ?

willglynn: I thought it was weird to have two code paths to retrieve records from the `Strtab`. `impl Index` can't be the primary path because it can't return a lifetime, so I made it delegate to `get()`.

willglynn: Ahh, [Travis is monkeying with their Mac infrastructure](https://www.traviscistatus.com/incidents/kdvnlc4w3n67) as part of an upgrade. Today's build delays are not normal.

m4b: Ah yea that's why I did that haha; I can't remember if the makefile is present but if so you can take make api and it will compile with different variations

PR #48, opened 2017-08-19
Treat "__.SYMDEF SORTED" as an alias of "__.SYMDEF"

willglynn:
I guess `__.SYMDEF SORTED` was invented to make symbol lookups faster? At any rate, some Mac archives have these instead of `__.SYMDEF`, and parsing them as if they were `__.SYMDEF` works fine, so this change makes that happen.


m4b:
F. Yea. I will just trust you this is correct :D

PR #55, opened 2017-10-11
Minor codestyle fixes

mre:
These are just some minor adjustments that don't change the logic of the library.
Mostly guided by clippy.


m4b:
Heh I dunno wth is going on with GitHub today. Says I can’t merge cause of a server error and it’s out of sync with base branch (sure but so what there aren’t any conflicts afaics)

PR #57, opened 2017-10-26
Remove the use of `goblin::parse` in `lib.rs` example

lion328:
I think you forgot this one. :)


m4b:
Ohhhhh it was in example, got it, good catch thanks again !

PR #60, opened 2017-11-24
Bump version of the `plain` dependency.

le-jzr:
As embarassing as it is to admit, there was a potential integer overflow in `plain`. I don't know for certain that goblin is affected, but you should probably rebuild the package just to be sure.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ include = ["src/**/*", "Cargo.toml", "CHANGELOG.md", "LICENSE", "README.md", "et
 env_logger = "0.4.3"
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


le-jzr:
Ow, wait, I forgot about zero-sized types. One second...

le-jzr:
Now it should be fine.

le-jzr:
I just remembered that cargo doesn't distribute binary packages. So you don't actually have to do anything. It's been so long since I had time to work on anything Rust-related, sorry for the noise. :)

m4b:
Hey @le-jzr thanks for the quick fix and updating this, really appreciate it :)

Issue #21, opened 2017-04-06
add a zero-copy, safe, symtable for use in no std


m4b:
It would pass out references to `Sym`s, which are byte-casted from a backing `&[u8]`; and have a `new -> Result` api (which validates the bounds).  We'd need two for elf32 and elf64 probably; or we can initialize as 32 or 64 bit (or pass a container context, which has less boolean blindness).  This will be tricky and annoying I think due to type name punning stuff, so probably easiest to just add two typed versions and re-export it

Similar to strtab, I would also want it to implement `Index`, so it can be literally drop-in replaced in code that previously used a `&[Sym]`.

Another approach could just provide a newtype wrapper on `&[Sym]` that validates the backing bytes and a count provided, and then `Deref`s to a `&[Sym]` so we get indexing for free.

Lots of options.

If we want to get fancy-pantsy, we might be able generify it to both ELF and Mach symbols, and have it return references via something like:

`get::<Symbol>(index) -> Result<&Symbol>`

but that might not be worth the effort.

m4b:
I think parts of this are implemented; not as interesting, so closing

PR #61, opened 2017-11-26
elf: add Symtab and SymIterator

philipc:
I want this to avoid potentially large and unneeded allocations for the symbol table. So with this, parsing the ELF header doesn't also parse all the symbols, which is useful if you don't care about the symbols. And the iterator lets me parse the symbols and put them in my own container, instead of requiring them all to be put in a Vec first.

This is all wrapped in `if_std!` because it needs `Sym`, so it's probably not what you want for #21. Let me know if you have ideas on how to adapt it for that.

### DIFF
--- a/src/elf/mod.rs
+++ b/src/elf/mod.rs
@@ -69,6 +69,7 @@ if_sylvan! {
     pub type Header = header::Header;
     pub type ProgramHeader = program_header::ProgramHeader;
     pub type SectionHeader = section_header::SectionHeader;
+    pub type Symtab<'a> = sym::Symtab<'a>;
     pub type Sym = sym::Sym;
     pub type Dyn = dyn::Dyn;
     pub type Dynamic = dyn::Dynamic;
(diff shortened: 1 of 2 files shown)


philipc:
> One last thing: this should have no effects on anyone iterating over symbols right (Unless they did explicitly vec things, but if just iterating should be no changes required, yes ?)

No, because the iterator items are now `Result<Sym>` instead of `Sym`. I'm not familiar with scroll, but it seems to me that the `gread_with` will never fail in this case, so maybe we can get away with a `.unwrap()` or `.ok()` to avoid the `Result`?

m4b:
Ya just realized that. We could unwrap the results if we ensure initial construction has valid bounds, which I believe is already done.  The parse will fail according to the `TryFromCtx` impl, which can generically be an out of bounds, or whatever it ends up implementing.  In this case, its just matches the container size and parses the 32 or 64 bit symbol, which is a generic struct which cannot fail if there is enough space.

So, I think it is safe to unwrap actually, as long as the bounds are enforced during symtab construction time.

m4b:
Going to merge this; will push patch making it resultless, thanks for this!

Next step is making relocations lazy, which will usually be much larger than the symbol table

PR #62, opened 2017-11-27
elf: change SymIterator::Item to Sym

philipc:


philipc:
Just missed your merge.. feel free to ignore this if you want.

m4b:
@philipc Thanks for this, but just pushed d10813e which implements this (also tweaks `get` to return option)

m4b:
Ya sorry for miscommunication :)

PR #69, opened 2018-01-01
lazy transducer PR

m4b:
WIP

Demonstrates ELF example update


m4b:
I benchmarked this a while back, and parallel parsing of the relocs did not drop parsing time very much; since we now have a lazy reloc parser thank to https://github.com/m4b/goblin/pull/102, closing this for now, as one of the usecases for introducing a generic lazy iterator is less paramount

BUT I still think lazy transducer is an awesome crate and idea ;P

PR #71, opened 2018-01-15
Allow iteration over note headers and sections

jan-auer:
This PR attempts to fix two conditions:

 - Some linkers, like for instance `lld`, might generate two `PT_NOTE` program header entries (i.e. segments). In that case, only iterating over one might miss essential notes.
 - Unsupported linkers might not perform the `PT_NOTE` magic and just leave `SHT_NOTE` sections in the binary. In that case, it is useful to iterate over those sections. Luckily, those sections are guaranteed to contain the same data structure as a `PT_NOTE` segment. In fact, the sections were first proposed before linkers started doing `PT_NOTE` segments.

It is not possible to create a custom `NoteIterator` from outside of `goblin`, as the required `ctx` is a private field on `Elf`. Therefore, this PR introduces two new methods on `elf::Elf`:
 - `iter_note_headers`: To iterate over all notes defined in `PT_NOTE` program entry headers. This should be the default case and only iterate over one single entry for most of the cases.
 - `iter_note_sections`: To iterate over all sections of type `SHT_NOTE`. This is for legacy linkers and compilers.

These methods are separate, as iterating over both will likely yield duplicates. To keep the breaking change to a minimum, the public facing iterator is still called `NoteIterator`. However, it is a new type that iterates over notes in all headers and sections, respectively. Internally, the old iterator is still present as `NoteDataIterator`. 

I'm not sure if this a way to go, but at least it's a starting point.


jan-auer:
As I started using this in getsentry/symbolic#47, I noticed that it might be convenient to declare the second method rather like this:

```rust
fn iter_note_sections(
    &self, 
    data: &'a [u8], 
    section_name: Option<&str>
) -> Option<note::NoteIterator<'a>>;
```

This trades the overhead of loading and iterating uninteresting node sections with one lookup in the `strtab` for the section name. Whatever's faster is clearly workload dependent (e.g. most of the time `SHT_NOTE` sections try to fit within the first page), so this signature leaves it up to the caller. In my case, this would for instance be:

```rust
elf.iter_note_sections(data, Some(".note.gnu.build-id"))
```

What do you think?

src/elf/note.rs
```diff
@@ -85,6 +85,27 @@ if_std! {
...
+
+    impl<'a> Iterator for NoteIterator<'a> {
+        type Item = error::Result<Note<'a>>;
+        fn next(&mut self) -> Option<Self::Item> {
+            while self.index < self.iters.len() {
+                if let Some(note_result) = self.iters[self.index].next() {
+                    return Some(note_result);
+                }
+
+                self.index += 1;
```
m4b: Won't this only return the first element of every iterator (e.g., shouldn't we also exhaust the iterator at `self.index` until it returns `None`, and then update the `self.index` to point to next iterator?)

src/elf/mod.rs
```diff
@@ -131,23 +131,56 @@ if_sylvan! {
...
-                    return Some(
-                        note::NoteIterator {
-                            data,
-                            offset,
-                            size: offset + phdr.p_filesz as usize,
-                            ctx: (alignment, self.ctx)
-                        }
-                    )
+
+                    iters.push(note::NoteDataIterator {
```
m4b: We can make this not allocate, but it would require another iterator struct which basically implements this logic, but I'm not sure its worth it at all. Alternatively, when `impl Trait` comes will be easy fix

jan-auer: I started that way but then the solution seemed way to complex for what it was trying to do. `impl Trait` will really make iterators much easier. Until then, I'd propose to leave it that way. Since all section and program headers are already parsed at that point, the only optimization would be to save a single (in **very** rare cases more) allocation.

src/elf/note.rs
jan-auer: Let me add a test for this like you suggested, in any case. However, I think this should be correct since `NoteIterator::next` will be called multiple times. Only if a `NoteDataIterator` is exhausted, it will hit the "else" case and increment the index. Otherwise, it just returns the found element and tries again in the next invocation. And since a `NoteDataIterator` could be entirely empty, the while loop tries until it finds one that still contains an item or there are none left.

m4b: Ah derp, you’re totally right! The early return in the loop was mixing me up for some reason

jan-auer: No hard feelings if you'd prefer a different coding style :) Just let me know.

jan-auer:
Updated with tests and the proposed `iter_note_sections` signature.

src/elf/note.rs
m4b: No I just needed coffee ;P

src/elf/note.rs
```diff
@@ -171,4 +192,79 @@ if_std! {
...
+                                     0x06, 0x00, 0x00, 0x00, 0x20, 0x00, 0x00, 0x00,
+                                     0x04, 0x00, 0x00, 0x00, 0x14, 0x00, 0x00, 0x00,
+                                     0x03, 0x00, 0x00, 0x00, 0x47, 0x4e, 0x55, 0x00,
+                                     0xbc, 0xfc, 0x66, 0xcd, 0xc7, 0xd5, 0x14, 0x7b,
+                                     0x53, 0xb1, 0x10, 0x11, 0x94, 0x86, 0x8e, 0xf9,
+                                     0x4f, 0xe8, 0xdd, 0xdb];
+
+        const CONTEXT: (usize, container::Ctx) = (4, container::Ctx {
+            container: container::Container::Big,
+            le: ::scroll::Endian::Little,
```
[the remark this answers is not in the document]
jan-auer: Lemme update that quickly (although you might wanna change that in the definition of `Ctx` then, too):
```rs
    pub struct Ctx {
        pub container: Container,
        pub le: scroll::Endian,
    }
```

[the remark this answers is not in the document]
jan-auer: Oh wait, never mind, misread the comment. Thanks for the tip :)

jan-auer:
Alright, just refactored the weird check for the section name (this block really doesn't get better, no matter which way you try to code it). 

Thanks a lot for your feedback and the great work you do on `goblin` in general! Please let me know once you know when you will release the next version. We depend on this PR for a bugfix we'd like to push out for [`symbolic`](https://github.com/getsentry/symbolic/pull/47) and would appreciate a soon release, if possible.

m4b:
I can release tonight. The parser on big endian machines was totally broken so ok to release ASAP. 

Also I removed all the ctx and endianness methods on mach header, which will break any code using them (which it should because they were totally invalid, as they only worked on LE machines)

Anyway, thanks for the PR, much appreciated !

philipc:
> And it’s hard to catch these regressions since Travis doesn’t expose BE machine tests afaik.

You can use [cross](https://github.com/japaric/cross) to do testing with QEMU on an x86_64 linux host in travis. I recently added a BE test using this in gimli.

m4b:
I have no idea why the CI is still going... says it hasn't even been queued. Whatever

m4b:
@jan-auer mokie dokie, 0.0.14 is published, good luck and have fun and thanks for all the fish! ;)

m4b:
:/ should have tried this with a project before publishing, didn't realize that `iter_notes` -> `iter_note_headers`, which was technically a breaking change and didn't add to release notes.

Also I think I would have preferred `iter_notes` to not be renamed, since notes are almost always in headers, so being default seems reasonable, and also wouldn't be a breaking update. i guess its fine.  Also its my fault for not paying more attention :P

jan-auer:
Sorry, should have made that more clear in the description. 

I was thinking about that rename for quite a while, actually. Another possibility would have been to even leave the old method in place, as this PR is essentially only adding compatibility with old or broken tools.

m4b:
Ya on hindsight would have preferred it; but its fine though; most likely the header removals in mach caused people breakage anyway, but perhaps would have been nicer to add a deprecation note to the `iter_notes` method.

Next time!

jan-auer:
Next time :) Alright, thanks a lot for the quick response and cheers

PR #77, opened 2018-02-01
Add alloc feature

### LINKED ISSUE #75: Add 'alloc' feature
For https://github.com/gimli-rs/object/issues/45, we need to be able to parse all the formats when using `#[no_std]`. Currently, a lot of the parsing requires allocation. Completely avoiding allocations is too much work for now, and probably not a good use of time. And anyway, we can still use the `alloc` crate with `no_std` (but this does require building with nightly currently).

So I propose we add an `alloc` feature that is midway between `no_std` and `std`. This feature will cover everything that uses allocations, and so mostly what will be left in `std` will be things that use `std::fs` or `std::io`.

I've done enough to verify that this approach works, so assuming this is acceptable, now I just need to clean it up and submit a PR.

One question I have is about the `endian_fd` feature. What exactly is this meant to cover? The readme says it 'parses according to the endianness in the binary', but it doesn't cover code such as
https://github.com/m4b/goblin/blob/master/src/elf/section_header.rs#L490. Currently it gates the entire mach/pe/archive formats, and it also requires 'std'. I want to relax this restriction as part of adding the `alloc` feature, but I'm not sure which parts of mach/pe/archive should still require it.

philipc:
This allows building the `endian_fd` feature without `std`, although it does still require the `alloc` crate and a nightly rustc.

Fixes #75

### DIFF
--- a/src/archive/mod.rs
+++ b/src/archive/mod.rs
@@ -11,8 +11,9 @@ use scroll::{self, Pread};
 use strtab;
 use error::{Result, Error};
 
-use std::usize;
-use std::collections::HashMap;
+use core::usize;
+use alloc::btree_map::BTreeMap;
+use alloc::vec::Vec;
 
 pub const SIZEOF_MAGIC: usize = 8;
 /// The magic number of a Unix Archive
@@ -334,9 +335,9 @@ pub struct Archive<'a> {
     sysv_name_index: NameIndex<'a>,
     // the array of members, which are indexed by the members hash and symbol index
     member_array: Vec<Member<'a>>,
-    members: HashMap<&'a str, usize>,
+    members: BTreeMap<&'a str, usize>,
     // symbol -> member
-    symbol_index: HashMap<&'a str, usize>
+    symbol_index: BTreeMap<&'a str, usize>
 }
 
 
@@ -383,8 +384,8 @@ impl<'a> Archive<'a> {
         }
 
         // preprocess member names
-        let mut members = HashMap::new();
-        let mut member_index_by_offset: HashMap<u32, usize> = HashMap::with_capacity(member_array.len());
+        let mut members = BTreeMap::new();
+        let mut member_index_by_offset: BTreeMap<u32, usize> = BTreeMap::new();
         for (i, member) in member_array.iter_mut().enumerate() {
             // copy in any SysV extended names
             if let Ok(sysv_name) = sysv_name_index.get(member.raw_name()) {
@@ -400,7 +401,7 @@ impl<'a> Archive<'a> {
         }
 
         // build the symbol index, translating symbol names into member indexes
-        let mut symbol_index: HashMap<&str, usize> = HashMap::new();
+        let mut symbol_index: BTreeMap<&str, usize> = BTreeMap::new();
         for (member_offset, name) in index.symbol_indexes.iter().zip(index.strtab.iter()) {
             let name = name.clone();
             let member_index = member_index_by_offset[member_offset];
--- a/src/elf/section_header.rs
+++ b/src/elf/section_header.rs
@@ -2,7 +2,7 @@ macro_rules! elf_section_header {
     ($size:ident) => {
         #[repr(C)]
         #[derive(Copy, Clone, Eq, PartialEq, Default)]
-        #[cfg_attr(feature = "std", derive(Pread, Pwrite, SizeWith))]
+        #[cfg_attr(feature = "alloc", derive(Pread, Pwrite, SizeWith))]
         /// Section Headers are typically used by humans and static linkers for additional information or how to relocate the object
         ///
         /// **NOTE** section headers are strippable from a binary without any loss of portability/executability; _do not_ rely on them being there!
@@ -267,15 +267,19 @@ macro_rules! elf_section_header_std_impl { ($size:ty) => {
         }
     }
 
-    if_std! {
+    if_alloc! {
         use elf::section_header::SectionHeader as ElfSectionHeader;
-        use error::Result;
-
-        use std::fs::File;
-        use std::io::{Read, Seek};
-        use std::io::SeekFrom::Start;
 
         use plain::Plain;
+        use alloc::vec::Vec;
+
+        if_std! {
+            use error::Result;
+
+            use std::fs::File;
+            use std::io::{Read, Seek};
+            use std::io::SeekFrom::Start;
+        }
 
         impl From<SectionHeader> for ElfSectionHeader {
             fn from(sh: SectionHeader) -> Self {
@@ -317,6 +321,7 @@ macro_rules! elf_section_header_std_impl { ($size:ty) => {
                 shdrs
             }
 
+            #[cfg(feature = "std")]
             pub fn from_fd(fd: &mut File, offset: u64, shnum: usize) -> Result<Vec<SectionHeader>> {
                 let mut shdrs = vec![SectionHeader::default(); shnum];
                 try!(fd.seek(Start(offset)));
@@ -326,7 +331,7 @@ macro_rules! elf_section_header_std_impl { ($size:ty) => {
                 Ok(shdrs)
             }
         }
-    } // end if_std
+    } // end if_alloc
 };}
 
 
@@ -356,13 +361,16 @@ pub mod section_header64 {
 // Std/analysis/Unified Structs
 ///////////////////////////////
 
-if_std! {
+if_alloc! {
     use error;
     use core::fmt;
     use core::result;
     use core::ops::Range;
     use scroll::ctx;
     use container::{Container, Ctx};
+
+    #[cfg(feature = "endian_fd")]
+    use alloc::vec::Vec;
 
     #[derive(Default, PartialEq, Clone)]
     /// A unified SectionHeader - convertable to and from 32-bit and 64-bit variants
@@ -536,4 +544,4 @@ if_std! {
             }
         }
     }
-} // end if_std
+} // end if_alloc
--- a/src/elf/sym.rs
+++ b/src/elf/sym.rs
@@ -116,16 +116,19 @@ macro_rules! elf_sym_std_impl {
             }
         }
 
-        if_std! {
+        if_alloc! {
             use elf::sym::Sym as ElfSym;
-            use error::Result;
 
             use core::fmt;
             use core::slice;
 
-            use std::fs::File;
-            use std::io::{Read, Seek};
-            use std::io::SeekFrom::Start;
+            if_std! {
+                use error::Result;
+
+                use std::fs::File;
+                use std::io::{Read, Seek};
+                use std::io::SeekFrom::Start;
+            }
 
             impl Sym {
                 /// Checks whether this `Sym` has `STB_GLOBAL`/`STB_WEAK` bind and a `st_value` of 0
@@ -185,6 +188,7 @@ macro_rules! elf_sym_std_impl {
                 slice::from_raw_parts(symp, count)
             }
 
+            #[cfg(feature = "std")]
             pub fn from_fd(fd: &mut File, offset: usize, count: usize) -> Result<Vec<Sym>> {
                 // TODO: AFAIK this shouldn't work, since i pass in a byte size...
                 let mut syms = vec![Sym::default(); count];
@@ -195,7 +199,7 @@ macro_rules! elf_sym_std_impl {
                 syms.dedup();
                 Ok(syms)
             }
-        } // end if_std
+        } // end if_alloc
     };
 }
 
@@ -204,7 +208,7 @@ pub mod sym32 {
 
     #[repr(C)]
     #[derive(Clone, Copy, PartialEq, Default)]
-    #[cfg_attr(feature = "std", derive(Pread, Pwrite, SizeWith))]
+    #[cfg_attr(feature = "alloc", derive(Pread, Pwrite, SizeWith))]
     /// 32-bit Sym - used for both static and dynamic symbol information in a binary
     pub struct Sym {
         /// Symbol name (string tbl index)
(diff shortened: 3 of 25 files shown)


src/elf/section_header.rs
```diff
@@ -356,14 +361,17 @@ pub mod section_header64 {
...
-if_std! {
+if_alloc! {
     use error;
     use core::fmt;
     use core::result;
     use core::ops::Range;
     use scroll::ctx;
     use container::{Container, Ctx};
 
+    #[cfg(feature = "endian_fd")]
```
m4b: Should this be endian fd or alloc?

src/elf/sym.rs
```diff
@@ -116,16 +116,19 @@ macro_rules! elf_sym_std_impl {
...
-            use error::Result;
 
             use core::fmt;
             use core::slice;
 
-            use std::fs::File;
-            use std::io::{Read, Seek};
-            use std::io::SeekFrom::Start;
+            if_std! {
+                use error::Result;
```
m4b: Why is error in std but not alloc?

src/archive/mod.rs
```diff
@@ -334,9 +335,9 @@ pub struct Archive<'a> {
     sysv_name_index: NameIndex<'a>,
     // the array of members, which are indexed by the members hash and symbol index
     member_array: Vec<Member<'a>>,
-    members: HashMap<&'a str, usize>,
+    members: BTreeMap<&'a str, usize>,
```
m4b: Why did this change? Because of random requirements?

src/lib.rs
```diff
@@ -79,33 +79,65 @@
...
+#![cfg_attr(all(feature = "alloc", not(feature = "std")), feature(alloc))]
 
 extern crate plain;
-#[cfg_attr(feature = "std", macro_use)]
+#[cfg_attr(feature = "alloc", macro_use)]
 extern crate scroll;
 
-#[cfg_attr(feature = "std", macro_use)]
-#[cfg(feature = "std")] extern crate log;
+#[cfg(feature = "log")]
```
m4b: Log isn’t a feature, will this work?

src/lib.rs
```diff
@@ -79,33 +79,65 @@
...
 
-#[cfg(feature = "std")]
-pub mod error;
+#[cfg(all(feature = "alloc", not(feature = "std")))]
+#[macro_use]
+extern crate alloc;
 
-pub mod strtab;
+#[cfg(feature = "std")]
+mod alloc {
```
m4b: Why is this a mod?

src/mach/fat.rs
```diff
@@ -58,8 +60,7 @@ impl FatHeader {
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, Default)]
-#[cfg_attr(feature = "std", derive(Pread, Pwrite, SizeWith))]
+#[derive(Clone, Copy, Default, Pread, Pwrite, SizeWith)]
```
m4b: If my understanding is correct these are unconditionally derived because in order for the module to be compiled it must have mach and alloc feature, yes? In which case we are safe to derive them. 

If so I almost prefer the explicit cfg feature since although it’s redundant it gives reader a clue what’s going on. Or in case of refactors it’s a second fail safe of sorts.

src/lib.rs
```diff
@@ -261,38 +292,38 @@ if_std! {
         fd.seek(SeekFrom::Start(0))?;
         peek_bytes(&bytes)
     }
-}
 
-#[cfg(all(feature = "endian_fd", feature = "elf64", feature = "elf32", feature = "pe64", feature = "pe32", feature = "mach64", feature = "mach32", feature = "archive"))]
```
m4b: Why is it safe to remove the intersection of all these features here and in the impl ?

Couldn’t the user compile with elf, mach, and std and the get a missing symbol error in the impl parser since PE wasn’t compiled. 

Or does endian_fd mean everything is enabled now ?

src/lib.rs
m4b: Nevremind, it’s getting reexportrd and used elsewhere

src/lib.rs
philipc: In https://github.com/m4b/goblin/issues/75#issuecomment-361093100, you said that the parser flag (aka endian_fd) should turn on every other flag. So that is now enforced in Cargo.toml. Since this is all wrapped in `if_endian_fd`, specifying the features here is redundant. As I said in that issue, I'm not sure if this is correct, but I actually kind of like not needing to list out all the features here, so I went with that, but I'm happy to change it.

src/mach/fat.rs
philipc: Yes, `mach` depends on `alloc`, and `alloc` depends on `scroll/derive`. I kind of agree with you, except that there's a ton of these in `load_command.rs` and I was lazy.  Additionally, for the ones in `load_command.rs` we would also need `cfg_attr(feature = "std", derive(IOread, IOwrite))` (I don't actually understand why they don't give errors currently...).

philipc:
> Can use on stable unless you explicitly request the alloc feature ? Then required to use nightly, yes ?

Yes, you can use on stable as long as either you request `std`, or you don't request `alloc` (or something that depends on `alloc`).  One change due to this PR though is that previously mach/pe/arcive required `std`, but now they only require `alloc`. So if you previously depended on only mach/pe/archive, then you'll need to add `std` to your features, so that's a breaking change, but you can still get it working on stable.

philipc:
> However the feature flag switching compiler requirements does worries me a little since features are supposed to be additive otherwise it can break cargo and have weird effects at a distance, or so I’m told.

The features are still additive, so I don't think it will break cargo. I think it's fine to have a feature that depends on nightly, plenty of crates do that, and this case isn't even as bad as some other uses of nightly because you can always request std in order to avoid the nightly requirement.

src/mach/fat.rs
m4b: Yea be lazy, I forgot there’s like 40 of them ;)

m4b:
So I should have asked this before but all things were done in haste, but what is the general motivating usecase for having the alloc flag?

So originally I thought that object wanted a no_std parser because it was being considered/suggested for use in the rustc compiler in its backtrace mechanism. https://github.com/rust-lang/rust/pull/44251

The first thing is I don’t understand precisely why the rustc compiler would require no std there, and furthermore why we couldn’t allocate. 

Perhaps that explains why the alloc flag is just for allocation parts?

But I still don’t understand who the alloc flag is for, specifically its nightly requirement, etc. 

So essentially I’m a little worried if the changes here are a solution in need of a problem.

philipc:
> The first thing is I don’t understand precisely why the rustc compiler would require no std there

I'm not clear on that either, since https://github.com/rust-lang/rust/pull/44251 is modifying libstd and using std::io. @whitequark can you elaborate?

> and furthermore why we couldn’t allocate.

But we can allocate. If we couldn't allocate then there would be no point adding an alloc feature.

> Perhaps that explains why the alloc flag is just for allocation parts?
>
> But I still don’t understand who the alloc flag is for, specifically its nightly requirement, etc.

The purpose of the alloc flag for the parts that require allocation, but don't require libstd.

The nightly requirement is because the [alloc](https://doc.rust-lang.org/alloc/) crate is not stable.

src/archive/mod.rs
m4b: Meh it’s fine. I don’t think O(n) lookup important for members, since there’s usually only like 10

src/lib.rs
m4b: Well that’s awesome, did not know this!

m4b:
I’ll address couple comments in a bit but also why are the tests failing ?

philipc:
Tests are failing on rust 1.18 because of cyclic features, which is only supported in relatively recent rust. The fix for that depends on what we decide about the endian_fd feature changes that you [commented](https://github.com/m4b/goblin/pull/77#discussion_r165276513) on.

whitequark:
@m4b For example, I would like to use goblin in the [ARTIQ](https://m-labs.hk/artiq) firmware so that it would symbolize backtraces on its own instead of requiring the user to manually run addr2line.

m4b:
@philipc 

> But we can allocate. If we couldn't allocate then there would be no point adding an alloc feature.

I just meant in the context of the compiler, it seemed like no_std was required, unless I missed something, and I was trying to figure out why that would be required.

Is it correct to say that in order to print backtraces in rustc, `no_std` is required? This still seems really strange to me; rustc afaics uses `std` as expected.

And i'm sure the C libbacktrace lib allocates, and is a regular C/C++ library without any funny business.

> The purpose of the alloc flag is for the parts that require allocation, but don't require libstd.

So my point in asking about who wants it was to try to get a feel for whether it's actually a real pressing usecase.

Specifically, I understand at this point how it works, but I'm trying to understand some usecases for it.

Which brings me to @whitequark :

> For example, I would like to use goblin in the ARTIQ firmware so that it would symbolize backtraces on its own instead of requiring the user to manually run addr2line.

Ok, so this is start for a good motivating example.  If I'm understanding right, on or1k firmware you're in a `no_std` environment + you use the `alloc` crate, is this correct?  To be 100% clear: you still can allocate, which you provide using the `alloc` crate nightly stuff, but in terms of rust, you need the parser crate to be `no_std`, yes?

I'm still curious also how goblin/object would help you in terms of "not requiring the user to manually run addr2line" - are you saying that object would be embedded in the firmware itself, and when an exception is raised of some kind, it prints the backtrace automatically before termination? (if so, that's awesome)

Bringing it back to the original rustc pr for backtrace printing, @whitequark, are the changes here, which allow a `no_std` mode with the `alloc` crate enough to have the `object` crate be substituted for backtrace generation, or is there another piece I'm missing?

Sorry for all the questions, just trying to understand the precise problem that's being solved by adding `no_std` + nightly `alloc` crate; I was worried that these changes were for a very, very niche usecase that likely wouldn't even be seen, maybe that's still the case, but flexibility is also good.

m4b:
> Tests are failing on rust 1.18 because of cyclic features, which is only supported in relatively recent rust. The fix for that depends on what we decide about the endian_fd feature changes that you commented on.

@philipc 

I'm not familiar with cyclic features; do you know offhand which rust? I would really like to avoid a version bump if possible, and am ok with a little momentary pain/redundancy if it means not bumping, but it depends on what exactly needs to change in order for that to happen.

philipc:
The cyclic features PR is https://github.com/rust-lang/cargo/pull/4473, I don't which rust that is in. But my intention is to fix the cycle, so that shouldn't require a version bump. However, I don't want to spend effort on that until we decide which way the endian_fd dependency should be. Either way can work, what's in this PR is just my best effort based on my understanding of your comments in #75, but maybe I misunderstood.

philipc:
> And i'm sure the C libbacktrace lib allocates, and is a regular C/C++ library without any funny business.

It uses anonymous mmap to allocate memory instead of malloc/free.

m4b:
> It uses anonymous mmap to allocate memory instead of malloc/free.

So it has a posix libc... with an operating system that allocates memory for you? So not no_std at all

m4b:
> However, I don't want to spend effort on that until we decide which way the endian_fd dependency should be

So I'm not sure what we're deciding here, but it sounds like its just whether endian_fd means the union of every feature, i.e., `endian_fd -> mach, elf, archive, std` more or less?

I don't think its unreasonable to want just an Archive or just ELF or just Mach-o parser (and the crate was built this way), so I think its better to leave this.

If you want the Object parser, then you need all of the features enabled (the default); like it is now.

It sounds like the changes as they are now would force you to have the Object parser (and hence the archive, Elf, and PE parsers), even if you just wanted the Mach parser, yes?

I think keeping the way it is now, with composable parsers, and a `cfg(all` flag for everything in order to get the `goblin::Object` the better, and if i'm understanding, the diff will actually be smaller?

whitequark:
> To be 100% clear: you still can allocate, which you provide using the alloc crate nightly stuff, but in terms of rust, you need the parser crate to be no_std, yes?

That is correct. Of course, goblin would be better off as a piece of software *in general* if it didn't allocate, and goblin not allocating would let me print backtraces in case of out-of-memory errors (we have a 4MB RAM region currently dedicated to that CPU core), but goblin+alloc would still be a huge improvement.

> I'm still curious also how goblin/object would help you in terms of "not requiring the user to manually run addr2line" - are you saying that object would be embedded in the firmware itself, and when an exception is raised of some kind, it prints the backtrace automatically before termination? (if so, that's awesome)

Correct. I already have an ELF loader for other reasons, so I would simply flash a statically linked ELF with debug symbols.

philipc:
> I think keeping the way it is now, with composable parsers, and a cfg(all flag for everything in order to get the goblin::Object the better, and if i'm understanding, the diff will actually be smaller?

Agreed. I've reverted the endian_fd dependency changes, and tests pass now.

m4b:
Ok this is looking good to me. I will test this later, will likely merge over weekend. 

If anyone has any comments speak now or forever hold your peace. This is pretty significant change. 

Thanks for your hard work and guidance on this @philipc!

/cc @willglynn if you have any comments (and again anyone else)

m4b:
This is looking, good, doesn't appear to be a breaking change at all!

Just tried this with dryad, which somehow miraculously had no other breaking changes, and works fine (besides segfaulting as expected)

So, one question I have is: what happens when user specifies alloc + std in feature flags, std is used instead right?

philipc:
Yes, std already depends on alloc in Cargo.toml.

m4b:
Alright let's merge this fella

m4b:
Minor nit, I'll fix it later, but shouldn't:

```
mach32 = ["alloc", "endian_fd"]
```

just be:

```
mach32 = ["endian_fd"]
```

since `endian_fd` depends on `alloc`?

philipc:
Doesn't matter to me, but I think it documents the dependency better: there's non-endian_fd code in mach that depends on alloc. If we later changed mach so that it didn't depend on endian_fd, there would still be code in mach that depended on alloc.

m4b:
ya, agreed actually.

PR #82, opened 2018-03-21
Fix parsing of PE32+ import tables

kjempelodott:
I noticed that only the first imported symbol from each library is being parsed in 64-bit PEs. This is one way to fix it.

### DIFF
--- a/src/pe/import.rs
+++ b/src/pe/import.rs
@@ -19,25 +19,28 @@ pub const IMPORT_RVA_MASK_64: u64 = 0x0000_0000_8fff_ffff;
 pub trait Bitfield<'a>: PartialEq + Eq + LowerHex + Debug + TryFromCtx<'a, scroll::Endian, Error=scroll::Error, Size=usize> {
     fn is_ordinal(&self) -> bool;
     fn to_ordinal(&self) -> u16;
-    fn to_rva(&self) -> usize;
+    fn to_rva(&self) -> u32;
     fn size_of() -> usize;
     fn is_zero(&self) -> bool;
+    fn vec_to_import_address_table(Vec<Self>) -> ImportAddressTable;
 }
 
 impl<'a> Bitfield<'a> for u64 {
     fn is_ordinal(&self) -> bool { self & IMPORT_BY_ORDINAL_64 == IMPORT_BY_ORDINAL_64 }
     fn to_ordinal(&self) -> u16 { (0xffff & self) as u16 }
-    fn to_rva(&self) -> usize { (self & IMPORT_RVA_MASK_64) as usize }
+    fn to_rva(&self) -> u32 { (self & IMPORT_RVA_MASK_64) as u32 }
     fn size_of() -> usize { 8 }
-    fn is_zero(&self) -> bool { *self == 0}
+    fn is_zero(&self) -> bool { *self == 0 }
+    fn vec_to_import_address_table(table: Vec<u64>) -> ImportAddressTable { ImportAddressTable::PE64(table) }
 }
 
 impl<'a> Bitfield<'a> for u32 {
     fn is_ordinal(&self) -> bool { self & IMPORT_BY_ORDINAL_32 == IMPORT_BY_ORDINAL_32 }
     fn to_ordinal(&self) -> u16 { (0xffff & self) as u16 }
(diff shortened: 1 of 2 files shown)


src/pe/import.rs
```diff
@@ -26,41 +60,37 @@ impl<'a> HintNameTableEntry<'a> {
 #[derive(Debug, Clone)]
 pub enum SyntheticImportLookupTableEntry<'a> {
     OrdinalNumber(u16),
-    HintNameTableRVA ((u32, HintNameTableEntry<'a>)), // [u8; 31] bitfield :/
+    HintNameTableRVA ((usize, HintNameTableEntry<'a>)), // [u8; 31] bitfield :/
```
le-jzr: Why is this changed to `usize`?

le-jzr: Never mind, I see it now. :)

philipc: As far as I can tell, it didn't need to change, and `u32` seems more correct to me, since it's never more than 32 bits even for P32+.

src/pe/mod.rs
```diff
@@ -90,9 +90,16 @@ impl<'a> PE<'a> {
             }
             debug!("exports: {:#?}", exports);
             if let &Some(import_table) = optional_header.data_directories.get_import_table() {
-                let id = import::ImportData::parse(bytes, &import_table, &sections)?;
+                let id = match is_64 {
```
philipc: An `if` statement seems more natural here?

src/pe/import.rs
```diff
@@ -83,9 +113,7 @@ impl<'a> ImportLookupTableEntry<'a> {
 }
 
 // get until entry is 0
-pub type ImportAddressTable = Vec<u32>;
-
-pub const SIZEOF_IMPORT_ADDRESS_TABLE_ENTRY: usize = 4;
+pub type ImportAddressTable = Vec<usize>;
```
philipc: This should be `u64`, and the import address table parsing also needs fixing to read either `u32` or `u64` based on whether it is PE32+.

src/pe/import.rs
```diff
@@ -1,13 +1,47 @@
...
 use pe::utils;
 use pe::data_directories;
 
+
+pub const IMPORT_BY_ORDINAL_32: u32 = 0x8000_0000;
+pub const IMPORT_BY_ORDINAL_64: u64 = 0x8000_0000_0000_0000;
+pub const IMPORT_RVA_MASK_32: u32 = 0x8fff_ffff;
+pub const IMPORT_RVA_MASK_64: u64 = 0x0000_0000_8fff_ffff;
+pub const SIZEOF_IMPORT_ADDRESS_TABLE_ENTRY_32: usize = 4;
+pub const SIZEOF_IMPORT_ADDRESS_TABLE_ENTRY_64: usize = 8;
```
philipc: Since this value is only used for the import lookup table (not the address table), I think it would be better to use 'lookup' in the name, even though both tables have the same entry size (and sometimes the same entry format).

src/pe/import.rs
```diff
@@ -26,41 +60,37 @@ impl<'a> HintNameTableEntry<'a> {
 #[derive(Debug, Clone)]
 pub enum SyntheticImportLookupTableEntry<'a> {
     OrdinalNumber(u16),
-    HintNameTableRVA ((u32, HintNameTableEntry<'a>)), // [u8; 31] bitfield :/
+    HintNameTableRVA ((usize, HintNameTableEntry<'a>)), // [u8; 31] bitfield :/
 }
 
 #[derive(Debug)]
 pub struct ImportLookupTableEntry<'a> {
-    pub bitfield: u32,
```
philipc: If we're going to  delete `bitfield` (which I think is fine to do), then there doesn't seem much point having this struct at all.

src/pe/import.rs
kjempelodott: Thank you for the comments. I changed ImportAddressTable to an enum that holds a Vec<u32> or Vec<u64>, and added a method in the Bitfield trait that converts Vec<T> to an ImportAddressTable. Not sure if this is the optimal solution though.

philipc: For unified 32-bit/64-bit types, I think the rest of the crate tends to convert the fields to `u64`, so I think a plain `Vec<u64>` is better. You can add a method in the Bitfield trait to convert a value to `u64`.

PR #85, opened 2018-04-13
Add constants for n_types when N_STAB field is being used

jrmuizel:


jrmuizel:
The reference that I was using didn't include N_PC or N_AST. I've now added them.

willglynn:
Merged. Thanks!

PR #87, opened 2018-04-26
Inline symbol-related methods

Amanieu:
Together with https://github.com/m4b/scroll_derive/pull/6, this significantly improves the performance of `SymIterator`.

When iterating through an ELF file with 2.5M symbols:
Before: 219ms
After: 120ms

### DIFF
--- a/src/elf/sym.rs
+++ b/src/elf/sym.rs
@@ -70,6 +70,7 @@ pub fn is_import(info: u8, value: u64) -> bool {
 }
 
 /// Convenience function to get the &'static str type from the symbols `st_info`.
+#[inline]
 pub fn get_type(info: u8) -> &'static str {
     type_to_str(st_type(info))
 }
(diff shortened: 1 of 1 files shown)


philipc:
How did you decide which functions to inline? Inlining functions is fine, just trying to gauge the completeness of this, or whether maybe some of these functions are big enough that maybe they shouldn't be inlined.

m4b:
Yea I don't think this many functions need to be inlined to speed up symbol iteration (probably just the try_from is my guess, taken care of in the scroll PR), e.g., `is_import`, `is_function`, `size_with`, `from_raw`, etc., shouldn't be need to be inlined

Amanieu:
I just marked functions which had a very simple implementation (e.g. `is_function`) or that can significantly benefit from constant propagation or loop unswitching (e.g. `try_from_ctx` where the `match` on `Container` can be moved out of the hot loop by the compiler).

src/elf/sym.rs
```diff
@@ -468,6 +483,7 @@ if_alloc! {
         type Item = <SymIterator<'a> as Iterator>::Item;
         type IntoIter = SymIterator<'a>;
 
+        #[inline]
```
philipc: This is the only one that I don't see any benefit for, since it won't be called often and there's little room for optimization.

Issue #76, opened 2018-02-01
Parsing PE with empty export data directory


tathanhdinh:
Some well-formed PE may have its export data directory with an empty table of `NumberOfFunctions` (as well as `NumberOfNames`,  `AddressOfFunctions`, etc.), for example the `Export Data Directory` of `apisetschema.dll` (a dll for [API redirection](https://blog.quarkslab.com/runtime-dll-name-resolution-apisetschema-part-i.html) from Windows 7) on my machine `Windows7.SP1`:

![apisetschema](https://user-images.githubusercontent.com/79418/35662192-18e7d004-0717-11e8-8fff-368edf4e963f.PNG)

But the PE parser of goblin will refuse such a PE, that is because of some checks, for example when parsing the [`AddressOfNames`](https://github.com/m4b/goblin/blob/master/src/pe/export.rs#L76):

`let name_pointer_table_offset = &mut utils::find_offset_or(export_directory_table.name_pointer_rva as usize, sections, &format!("Cannot map export_directory_table.name_pointer_rva ({:#x}) into offset", export_directory_table.name_pointer_rva))?;`

because `export_directory_table.name_pointer_rva` will be zero then the including function `parse` returns immediately with an `Err(_)`.

The same with `AddressOfNameOrdinals` and `AddressOfFunctions` tables, a quick-and-dirty fix might be checking the value of `export_directory_table.number_of_name_pointers` and of `export_directory_table.address_table_entries`  with `0` before parsing offsets of these tables:

```
let mut export_name_pointer_table: ExportNamePointerTable = Vec::with_capacity(number_of_name_pointers);
let mut export_ordinal_table: ExportOrdinalTable = Vec::with_capacity(number_of_name_pointers);

if number_of_name_pointers > 0 {
    let name_pointer_table_offset = &mut utils::find_offset_or(export_directory_table.name_pointer_rva as usize, sections, &format!("Cannot map export_directory_table.name_pointer_rva ({:#x}) into offset", export_directory_table.name_pointer_rva))?;
    for _ in 0..number_of_name_pointers {
        export_name_pointer_table.push(bytes.gread_with(name_pointer_table_offset, scroll::LE)?);
    }

    let export_ordinal_table_offset = &mut utils::find_offset_or(export_directory_table.ordinal_table_rva as usize, sections, &format!("Cannot map export_directory_table.ordinal_table_rva ({:#x}) into offset", export_directory_table.ordinal_table_rva))?;

[...]

PR #88, opened 2018-04-29
Parsing PE export

tathanhdinh:
IMHO, goblin parser might be "too strict" in parsing export directory table and export entries of "abnormal" PEs. 

First, goblin will [refuse](https://github.com/m4b/goblin/blob/master/src/pe/export.rs#L77) parsing a PE if it cannot get the RVA of the name pointer table. In fact, an executable PE can have this RVA invalid, for example [this sample](https://github.com/corkami/pocs/blob/master/PE/bin/exportsdata.exe) from Corkami is a totally valid PE (it can even execute), but goblin will reject this PE.

Second, goblin parses export entries using [number of names](https://github.com/m4b/goblin/blob/master/src/pe/export.rs#L235) it found in the `export name pointer table` (and not from `number of names`), it will [stop parsing and reject](https://github.com/m4b/goblin/blob/master/src/pe/export.rs#L80) if there is a name with invalid RVA. But a `dll` can have invalid RVAs in the `export name pointer table`, in the following example of `wmi.dll` in Windows:

wmi_dll

I have modified the first entry (photo on the right), which contains RVA to the name of the first exported function `ControlTrace`, to zero; `Adlice PE viewer` still be able to list other entries (other tools like `FileAlyzer` and `PE Insider` are be able too). Moreover, the modified file can be loaded happily by `LoadLibraryEx` API (with `DONT_RESOLVE_DLL_REFERENCES` flag), and other entries can be found by `GetProcAddress` using the returned handle (of `LoadLibraryEx`)

This pull request try to fix these problems (and the problem I have reported in [issue 76](https://github.com/m4b/goblin/issues/76)). Of course YMMV, but please give me some ideas.

### DIFF
--- a/src/pe/export.rs
+++ b/src/pe/export.rs
@@ -58,9 +58,9 @@ pub type ExportOrdinalTable = Vec<u16>;
 pub struct ExportData<'a> {
     pub name: Option<&'a str>,
     pub export_directory_table: ExportDirectoryTable,
-    pub export_name_pointer_table: Option<ExportNamePointerTable>,
-    pub export_ordinal_table: Option<ExportOrdinalTable>,
-    pub export_address_table: Option<ExportAddressTable>,
+    pub export_name_pointer_table: ExportNamePointerTable,
+    pub export_ordinal_table: ExportOrdinalTable,
+    pub export_address_table: ExportAddressTable,
 }
 
 impl<'a> ExportData<'a> {
@@ -74,15 +74,14 @@ impl<'a> ExportData<'a> {
         let number_of_name_pointers = export_directory_table.number_of_name_pointers as usize;
         let address_table_entries = export_directory_table.address_table_entries as usize;
 
-        let export_name_pointer_table = utils::find_offset(export_directory_table.name_pointer_rva as usize, sections).map(|table_offset| {
+        let export_name_pointer_table = utils::find_offset(export_directory_table.name_pointer_rva as usize, sections).map_or(vec![], |table_offset| {
             let mut offset = table_offset;
             let mut table: ExportNamePointerTable = Vec::with_capacity(number_of_name_pointers);
 
             for _ in 0..number_of_name_pointers {
                 if let Ok(name_rva) = bytes.gread_with(&mut offset, scroll::LE) {
                     table.push(name_rva);
-                }
-                else {
+                } else {
                     break;
                 }
             }
@@ -90,15 +89,14 @@ impl<'a> ExportData<'a> {
             table
         });
 
-        let export_ordinal_table = utils::find_offset(export_directory_table.ordinal_table_rva as usize, sections).map(|table_offset| {
+        let export_ordinal_table = utils::find_offset(export_directory_table.ordinal_table_rva as usize, sections).map_or(vec![], |table_offset| {
             let mut offset = table_offset;
             let mut table: ExportOrdinalTable = Vec::with_capacity(number_of_name_pointers);
 
             for _ in 0..number_of_name_pointers {
                 if let Ok(name_ordinal) = bytes.gread_with(&mut offset, scroll::LE) {
                     table.push(name_ordinal);
-                }
-                else {
+                } else {
                     break;
                 }
             }
@@ -106,7 +104,7 @@ impl<'a> ExportData<'a> {
             table
         });
 
-        let export_address_table = utils::find_offset(export_directory_table.export_address_table_rva as usize, sections).map(|table_offset| {
+        let export_address_table = utils::find_offset(export_directory_table.export_address_table_rva as usize, sections).map_or(vec![], |table_offset| {
             let mut offset = table_offset;
             let mut table: ExportAddressTable = Vec::with_capacity(address_table_entries);
             let export_end = export_rva + size;
@@ -118,8 +116,7 @@ impl<'a> ExportData<'a> {
                     } else {
                         table.push(ExportAddressTableEntry::ExportRVA(func_rva));
                     }
-                }
-                else {
+                } else {
                     break;
                 }
             }
@@ -203,13 +200,13 @@ pub struct Export<'a> {
     pub name: Option<&'a str>,
     pub offset: usize,
     pub rva: usize,
-    // pub size: usize,
+    pub size: usize,
     pub reexport: Option<Reexport<'a>>,
 }
 
 #[derive(Debug, Copy, Clone)]
 struct ExportCtx<'a> {
-    pub ptr: Option<u32>,
+    pub ptr: u32,
     pub idx: usize,
     pub sections: &'a [section_table::SectionTable],
     pub addresses: &'a ExportAddressTable,
@@ -223,10 +220,7 @@ impl<'a, 'b> scroll::ctx::TryFromCtx<'a, ExportCtx<'b>> for Export<'a> {
     fn try_from_ctx(bytes: &'a [u8], ExportCtx { ptr, idx, sections, addresses, ordinals }: ExportCtx<'b>) -> Result<(Self, Self::Size), Self::Error> {
         use self::ExportAddressTableEntry::*;
 
-        let name = ptr.map_or(None, |ptr| {
-            let name_offset = utils::find_offset(ptr as usize, sections);
-            name_offset.and_then(|offset| bytes.pread::<&str>(offset).ok())
-        });
+        let name = utils::find_offset(ptr as usize, sections).map_or(None, |offset| bytes.pread::<&str>(offset).ok());
 
         if let Some(ordinal) = ordinals.get(idx) {
             if let Some(rva) = addresses.get(*ordinal as usize) {
@@ -234,22 +228,20 @@ impl<'a, 'b> scroll::ctx::TryFromCtx<'a, ExportCtx<'b>> for Export<'a> {
                     ExportRVA(rva) => {
                         let rva = rva as usize;
                         let offset = utils::find_offset_or(rva, sections, &format!("cannot map RVA ({:#x}) of export ordinal {} into offset", rva, ordinal))?;
-                        Ok((Export { name, offset, rva, reexport: None }, 0))
+                        Ok((Export { name, offset, rva, reexport: None, size: 0 }, 0))
                     },
 
                     ForwarderRVA(rva) => {
                         let rva = rva as usize;
                         let offset = utils::find_offset_or(rva, sections, &format!("cannot map RVA ({:#x}) of export ordinal {} into offset", rva, ordinal))?;
                         let reexport = Reexport::parse(bytes, offset)?;
-                        Ok((Export { name, offset, rva, reexport: Some(reexport) }, 0))
+                        Ok((Export { name, offset, rva, reexport: Some(reexport), size: 0 }, 0))
                     }
                 }
-            }
-            else {
+            } else {
                 Err(error::Error::Malformed(format!("cannot get RVA of export ordinal {}", ordinal)))
             }
-        }
-        else {
+        } else {
             Err(error::Error::Malformed(format!("cannot get ordinal of export name entry {}", idx)))
         }
     }
@@ -257,27 +249,15 @@ impl<'a, 'b> scroll::ctx::TryFromCtx<'a, ExportCtx<'b>> for Export<'a> {
 
 impl<'a> Export<'a> {
     pub fn parse(bytes: &'a [u8], export_data: &ExportData, sections: &[section_table::SectionTable]) -> error::Result<Vec<Export<'a>>> {
-        let number_of_name_pointers = export_data.export_directory_table.number_of_name_pointers;
+        let pointers = &export_data.export_name_pointer_table;
+        let addresses = &export_data.export_address_table;
+        let ordinals = &export_data.export_ordinal_table;
 
-        let mut addresses = &Vec::new();
-        if let Some(ref table) = export_data.export_address_table {
-            addresses = table;
-        }
-
-        let mut ordinals = &Vec::new();
-        if let Some(ref table) = export_data.export_ordinal_table {
-            ordinals = table;
-        }
-
-        let mut exports = Vec::with_capacity(number_of_name_pointers as usize);
-        for idx in 0..number_of_name_pointers {
-            let ptr = export_data.export_name_pointer_table.as_ref().map_or(None, |pointers| {
-                pointers.get(idx as usize).map(|v| *v)
-            });
-
-            if let Ok(export) = bytes.pread_with(0, ExportCtx { ptr: ptr, idx: idx as usize, sections: sections, addresses: addresses, ordinals: ordinals }) {
+        let mut exports = Vec::with_capacity(pointers.len());
+        for (idx, &ptr) in pointers.iter().enumerate() {
+            if let Ok(export) = bytes.pread_with(0, ExportCtx { ptr, idx, sections, addresses, ordinals }) {
                 exports.push(export);
-            }
+            }    
         }
 
         // TODO: sort + compute size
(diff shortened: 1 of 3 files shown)


m4b:
I haven't looked at changes yet, but thanks for your work here!

I briefly tested on some PE binaries which gave goblin parser issues before, and they seem resolved, which is really awesome.

Since the PR is so large, it will be good to have a couple of people check this out. @willglynn and  anyone else are welcome to review.

Also, from a git perspective, I'd like to see the commit history cleaned up a bit; there's a lot of merges, and not very informative git messages.

I know that there isn't a policy on this repo for commit messages but I don't think its useful to have a separate commit with, to pick on a couple, "code cleanup", or "minor" as content.

So I think we can do this two ways:

1. If you want to preserve several commits assuming this gets ok'd and merged, which I could understand, since its non-trivial amount of work and more commits shows up better in repo stats, etc., then I'd like to ask you to clean up the patch into 2-4 commits.

2. If that sounds boring and you don't care, then I'd prefer you to either rebase and squash all your commits into one with a good summary of the changes, or whoever merges this should do a squash and merge.

If you don't know how to do the latter or can't be bothered, that's ok, I or others can help out here :)

And thanks for your hard work!

tathanhdinh:
Many thanks for this encouraging response, and sorry for this messy PR. I would prefer the second approach (i.e. rebase + squash). I still do not know how to do this yet, but never mind, I would handle this myself.
Again, thanks for your feedback (I have been hesitated a lot before making the PR, I know that it is messy and though that it would be rejected immediately).

tathanhdinh:
Hello all
I've rebased my fork and merged my commit into a single commit. I am still not sure that is what you are suggested. In brief, what I have done is:

```
git rebase -i HEAD~26
(move external commits to the top, leave them as pick)
pick my first commit
fixup others
```

Please do not hesitate to correct me if you find any problem.

m4b:
hmmm, assuming you just pushed the rebased version, seems you've still got some redundant commits.

I think the easiest way to fix this would be to get the sum of the changes, which now appear to be in 0f8704630bedc91425e9b8a03159db32439acaab, and then do something like:

```
git show > MY_PATCH.diff # assuming all your changes are in HEAD, which looks like they are
git checkout -b branch_backup # save branch just in case
git checkout origin/master && git pull # get latest master, assuming origin is this repo, i.e.: github.com/m4b/goblin
git checkout <your branch name> # i think your branch is actually on master?
git reset --hard origin/master
git apply MY_PATCH.diff # this should only complain about whitespace errors 
```

I've verified this works for me.

Once the patch is on top of master, you should be able to force push to your branch:

```
git push --force
```

Also, I noticed there are some whitespace "errors" (git show should highlight whitespace with red); if you could remove those, i think the patch is ready for review!

tathanhdinh:
Hello all,
Sorry for the delayed update. I have just reapplied the patch following the guide of @m4b (many thanks for your help); the PR now contains a single monolithic commit.

src/pe/export.rs
```diff
@@ -173,16 +200,16 @@ impl<'a> Reexport<'a> {
 #[derive(Debug, Default)]
 /// An exported symbol in this binary, contains synthetic data (name offset, etc., are computed)
 pub struct Export<'a> {
```
m4b: So the export has every field optional? This doesn't seem right somehow, and i'm not sure what it means exactly for every field to be optional now.

I don't like this, and don't really think this is an appropriate API to expose to consumers; if its somehow:

```
Export {
  name: None,
  offset: None,
  rva: None
  reexport: None
}
```

Then the export is really just None, and the api/construction of exports should drop it, not push optionalness into the fields.

src/pe/export.rs
```diff
@@ -195,47 +222,69 @@ impl<'a, 'b> scroll::ctx::TryFromCtx<'a, ExportCtx<'b>> for Export<'a> {
...
-                    debug!("{}: reexport {:?}", name, reexport);
-                    Ok((Export { name: name, offset: rva, rva: rva, reexport: Some(reexport), size: 0 }, 0))
-                },
+        
+        let name = ptr.map_or(None, |ptr| {
+            let name_offset = utils::find_offset(ptr as usize, sections);
+            name_offset.and_then(|offset| bytes.pread::<&str>(offset).ok())
+        });
+
+        let ordinal = ordinals.get(idx);
```
m4b: This check should be in the loop below which constructs the export vector; not here, this way you don't have to make every field in export None; I believe your switching name to None is valid, but the other fields all cannot be, for the reasons I gave above.

src/pe/export.rs
```diff
@@ -195,47 +222,69 @@ impl<'a, 'b> scroll::ctx::TryFromCtx<'a, ExportCtx<'b>> for Export<'a> {
...
+            ordinals = table;
+        }
+
+        let mut exports = Vec::with_capacity(number_of_name_pointers as usize);
+        for idx in 0..number_of_name_pointers {
+            let ptr = export_data.export_name_pointer_table.as_ref().map_or(None, |pointers| {
+                pointers.get(idx as usize).map(|v| *v)
+            });
+
+            let export = bytes.pread_with(0, ExportCtx { ptr: ptr, idx: idx as usize, sections: sections, addresses: addresses, ordinals: ordinals })?;
```
m4b: I.e., here check if idx is valid, if not, you can skip the export

src/pe/mod.rs
```diff
@@ -42,7 +42,7 @@ pub struct PE<'a> {
     /// Data for any imported symbols, and from which `dll`, etc., in this binary
     pub import_data: Option<import::ImportData<'a>>,
     /// The list of exported symbols in this binary, contains synthetic information for easier analysis
-    pub exports: Vec<export::Export<'a>>,
+    pub exports: Option<Vec<export::Export<'a>>>,
```
m4b: I typically prefer an empty vector for zero exports, as its less heavy-weight on the user, and in principle, they still need to check whether the vector is empty anyway.

In other words, in this usecase, making it optional doesn't really give us any extra semantic help here, so I'd just leave it as a regular vector (and empty if no exports), unless there's some other compelling reason I'm not seeing

m4b:
Oh, and I'll note goblin cannot currently parse these (65) binaries, but with these changes it can:

```
psmsrv.dll
shunimpl.dll
winload.exe
nlsfunc.exe
mscdexnt.exe
DeviceUxRes.dll
ntvdmd.dll
vdmredir.dll
ntvdm.exe
fhuxgraphics.dll
winresume.exe
ntasn1.dll
ctl3dv2.dll
msacm.dll
HalExtIntcLpioDMA.dll
mfc100u.dll
compobj.dll
debug.exe
mswdat10.dll
prncache.dll
sysedit.exe
miguiresource.dll
apisetschema.dll
mem.exe
dosx.exe
append.exe
exe2bin.exe
win87em.dll
wscript.exe
uDWM.dll
mswstr10.dll
profapi.dll
olecli.dll
setver.exe
SetProxyCredential.dll
fastopen.exe
dnsext.dll
pmspl.dll
WinSyncMetastore.dll
typelib.dll
mfc100.dll
odbc16gt.dll
ole2nls.dll
msidle.dll
ext-ms-win-gdi-private-l1-1-0.dll
redir.exe
avifile.dll
WinSync.dll
oleacchooks.dll
wfapigp.dll
ole2.dll
edlin.exe
lzexpand.dll
cscript.exe
krnl386.exe
ole2disp.dll
share.exe
XpsGdiConverter.dll
storage.dll
netapi.dll
bootux.dll
HalExtIntcUartDMA.dll
ver.dll
tapi.dll
msrepl40.dll
```

m4b:
Also, feel free to push new commits with the changes; no need to rebase and squash, etc.

src/pe/export.rs
```diff
@@ -56,52 +56,79 @@ pub type ExportOrdinalTable = Vec<u16>;
...
-        let mut export_ordinal_table: ExportOrdinalTable = Vec::with_capacity(number_of_name_pointers);
-        for _ in 0..number_of_name_pointers {
-            export_ordinal_table.push(bytes.gread_with(export_ordinal_table_offset, scroll::LE)?);
-        }
+            for _ in 0..number_of_name_pointers {
+                if let Ok(name_rva) = bytes.gread_with(&mut offset, scroll::LE) {
+                    table.push(name_rva);
+                }
+                else {
+                    break;
```
m4b: if i understand, we stop parsing all `name_pointers` here (and other `break` below for `export_address_table`) if a single one is bad, yes? is it possible to bump the offset by sizeof(name_rva), and continue?

This might not be a good idea for various reasons, and feel free not to implement this if its too much work, etc., since your PR already successfully parses my PE binary suite, so it would only be super-malformed binaries which isn't really a high priority.

src/pe/export.rs
```diff
@@ -56,52 +56,79 @@ pub type ExportOrdinalTable = Vec<u16>;
...
+
+            for _ in 0..address_table_entries {
+                if let Ok(func_rva) = bytes.gread_with::<u32>(&mut offset, scroll::LE) {
+                    if utils::is_in_range(func_rva as usize, export_rva, export_end) {
+                        table.push(ExportAddressTableEntry::ForwarderRVA(func_rva));
+                    } else {
+                        table.push(ExportAddressTableEntry::ExportRVA(func_rva));
+                    }
+                }
+                else {
```
m4b: Nit: else on newline (doesn't match style above)

src/pe/export.rs
```diff
@@ -173,16 +200,16 @@ impl<'a> Reexport<'a> {
 #[derive(Debug, Default)]
 /// An exported symbol in this binary, contains synthetic data (name offset, etc., are computed)
 pub struct Export<'a> {
-    pub name: &'a str,
+    pub name: Option<&'a str>,
     pub offset: usize,
     pub rva: usize,
-    pub size: usize,
+    // pub size: usize,
```
m4b: Why was this field removed?

src/pe/export.rs
```diff
@@ -195,47 +222,64 @@ impl<'a, 'b> scroll::ctx::TryFromCtx<'a, ExportCtx<'b>> for Export<'a> {
...
+
+                    ForwarderRVA(rva) => {
+                        let rva = rva as usize;
+                        let offset = utils::find_offset_or(rva, sections, &format!("cannot map RVA ({:#x}) of export ordinal {} into offset", rva, ordinal))?;
+                        let reexport = Reexport::parse(bytes, offset)?;
+                        Ok((Export { name, offset, rva, reexport: Some(reexport) }, 0))
+                    }
+                }
             }
+            else {
```
m4b: Nit: else on newline

src/pe/export.rs
```diff
@@ -195,47 +222,64 @@ impl<'a, 'b> scroll::ctx::TryFromCtx<'a, ExportCtx<'b>> for Export<'a> {
...
+            ordinals = table;
         }
+
+        let mut exports = Vec::with_capacity(number_of_name_pointers as usize);
+        for idx in 0..number_of_name_pointers {
+            let ptr = export_data.export_name_pointer_table.as_ref().map_or(None, |pointers| {
+                pointers.get(idx as usize).map(|v| *v)
+            });
+
+            if let Ok(export) = bytes.pread_with(0, ExportCtx { ptr: ptr, idx: idx as usize, sections: sections, addresses: addresses, ordinals: ordinals }) {
```
m4b: Great usecase for type punning/field initializer syntax, e.g., `ExportCtx { ptr, sections, addresses, ordinals: idx: idx as usize }`

src/pe/export.rs
```diff
@@ -195,47 +222,64 @@ impl<'a, 'b> scroll::ctx::TryFromCtx<'a, ExportCtx<'b>> for Export<'a> {
...
-        //let ordinal_base = export_data.export_directory_table.ordinal_base as usize;
-        let mut exports = Vec::with_capacity(pointers.len());
-        for (idx, ptr) in pointers.iter().enumerate() {
-            let export = bytes.pread_with(0, ExportCtx { ptr: *ptr, idx: idx, sections: sections, addresses: addresses, ordinals: ordinals })?;
-            exports.push(export);
+        let number_of_name_pointers = export_data.export_directory_table.number_of_name_pointers;
+
+        let mut addresses = &Vec::new();
+        if let Some(ref table) = export_data.export_address_table {
+            addresses = table;
```
m4b: That you are doing this here, and below, in order to get an empty vec immediately suggests to me my concern above about the optional values is well-founded, and you should unmark it optional, and the empty vec is the "None" case.

src/pe/export.rs
```diff
@@ -56,52 +56,79 @@ pub type ExportOrdinalTable = Vec<u16>;
...
 /// Export data contains the `dll` name which other libraries can import symbols by (two-level namespace), as well as other important indexing data allowing symbol lookups
 pub struct ExportData<'a> {
-    pub name: &'a str,
+    pub name: Option<&'a str>,
     pub export_directory_table: ExportDirectoryTable,
-    pub export_name_pointer_table: ExportNamePointerTable,
-    pub export_ordinal_table: ExportOrdinalTable,
-    pub export_address_table: ExportAddressTable,
+    pub export_name_pointer_table: Option<ExportNamePointerTable>,
+    pub export_ordinal_table: Option<ExportOrdinalTable>,
```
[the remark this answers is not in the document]
philipc: I would say ditto for `export_name_pointer_table` too. I would only use `Option` for these if you can come up with a reason why you want to distinguish between `None` and `Some(vec![])`, ie between invalid RVA and zero size. Otherwise it's simpler for consumers if it's plain `Vec` that is empty for both invalid RVA and zero size.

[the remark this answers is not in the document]
tathanhdinh: Many thanks to @m4b  and @philipc for the encouraging response.

Yes, my argument for `Option` is to distinguish the case where a [correct offset](https://github.com/m4b/goblin/blob/master/src/pe/export.rs#L77) cannot be obtained (i.e. the table does not exist), from the case where the table's offset can be obtained (i.e. the table exists) but its size is zero.

For executable `PE`(i.e. `.exe` files), I think that the Windows's PE loader will not care much if the export-related tables are valid or not (it is possible to make a runable `.exe` having invalid `export_name_pointer_table`). So these tables "can be skipped".

But I am happy also if we choose to unify these cases :), I don't know if this choice is better or not but, as you have said, it will not break the API and maybe simpler for the consumers.

I have changed the PR to restore these tables to not optional.

src/pe/export.rs
tathanhdinh: Ah, I think I get your point.

But the case where a `name_rva` cannot be read

```
for _ in 0..number_of_name_pointers {
  if let Ok(name_rva) = bytes.gread_with(&mut offset, scroll::LE) {
    table.push(name_rva);
    } else {
      break;
    }
}
```
there would be a serious error in the PE file for which data cannot be read from `bytes` at the given `offset` (e.g. the file is truncated), in such a case continuing to read would not help (personal opinion).

src/pe/export.rs
tathanhdinh: Sorry, I did not think of the style.

src/pe/export.rs
tathanhdinh: Sorry, because I do not understand why this field is needed, I have restored `size` (but I still do understand :))

src/pe/export.rs
tathanhdinh: Sorry again.

src/pe/export.rs
tathanhdinh: Ah, I think I get your point. Letting empty vec as `None` case would minimize the change, if not we should change also `ExportCtx` !?

tathanhdinh:
Many thanks for the encouraging response and the exchange of ideas. I have modified the PR to reflect this.

tathanhdinh:
Many thanks!!!

koutheir:
This issue seems to be fixed in the master, but it is not provided in the version `0.0.15` of goblin, which is declared to be the latest version (See [`https://docs.rs/crate/goblin/`](https://docs.rs/crate/goblin/)). Could you please create a new version with this fix?

m4b:
Done, version 0.0.16 is out :)

PR #89, opened 2018-04-30
add MIPS-specific DT entries

endeav0r:


src/elf/dyn.rs
```diff
@@ -153,6 +153,47 @@ pub const DT_VERNEED: u64 = 0x6ffffffe;
...
+/// Holds the index of the first dynamic symbol table entry that corresponds to
+/// an entry in the global offset table.
+pub const DT_MIPS_GOTSYM: u64 = 0x70000013;
+/// Holds the number of page table entries in the global offset table.
+pub const DT_MIPS_HIPAGENO: u64 = 0x70000014;
+/// Address of a 32-bit word in the .data section used by debugging. Its
+/// contents are unspecified.
+pub const DT_MIPS_RLD_MAP: u64 = 0x70000016;
+
+
```
philipc: LLVM has [more MIPS definitions](https://github.com/llvm-mirror/llvm/blob/f4f861bbdf9be5247430542e8b4b86c28b7eeef5/include/llvm/BinaryFormat/DynamicTags.def#L131-L177). Do these need to be added to?

src/elf/dyn.rs
```diff
@@ -204,6 +245,22 @@ pub fn tag_to_str(tag: u64) -> &'static str {
...
+        DT_MIPS_CONFLICT => "DT_MIPS_CONFLICT",
+        DT_MIPS_LIBLIST => "DT_MIPS_LIBLIST",
+        DT_MIPS_LOCAL_GOTNO => "DT_MIPS_LOCAL_GOTNO",
+        DT_MIPS_CONFLICTNO => "DT_MIPS_CONFLICTNO",
+        DT_MIPS_LIBLISTNO => "DT_MIPS_LIBLISTNO",
+        DT_MIPS_SYMTABNO => "DT_MIPS_SYMTABNO",
+        DT_MIPS_UNREFEXTNO => "DT_MIPS_UNREFEXTNO",
+        DT_MIPS_GOTSYM => "DT_MIPS_GOTSYM",
+        DT_MIPS_HIPAGENO => "DT_MIPS_HIPAGENO",
+        DT_MIPS_RLD_MAP => "DT_MIPS_RLD_MAP",
```
philipc: Since these are processor specific, I'm not sure it's correct to return them without knowing that the processor is MIPS. Is this something that other libraries do?

src/elf/dyn.rs
endeav0r: I don't believe so. These are the definitions in the System V ABI.

src/elf/dyn.rs
endeav0r: Normally they're just `#define`. I don't think it should be an issue, the tags shouldn't overlap.

src/elf/dyn.rs
philipc: Sure, but that's 20 years old isn't it? I think the ABI has evolved since then. I don't know where the current ABI is specified. I found some info [here](https://dmz-portal.mips.com/wiki/MIPS_Dynamic).

src/elf/dyn.rs
philipc: `tag_to_str` is used in `<Dyn as Debug>::fmt`. Since we can't pass a processor argument to  `Debug::fmt`, I don't see any way for that to correctly print these values. So `tag_to_str` can't print these values. We could improve it to print something like `PROCESSOR_SPECIFIC` instead `UNKNOWN_TAG` though.

If there is a need to get these strings for other use cases, then you could define an additional function that does accept a processor argument.

m4b:
@endeav0r Sorry for the delay: I'm ok with merging the DT_MIPS_* values, but I don't think its appropriate to print them in the 2string method.

We can add a specific printer I suppose, but that should be easy enough for any client to do, and I'm not too worried about it. If you remove the 2string cases, I'm ok with merging.

m4b:
The mach-o cpusubtypes have a similar issue, can see one solution here: https://github.com//m4b/goblin/blob/7964a0e8fccd6cdfc43660b9ab0b91ebc7d5141d/src/mach/constants.rs#L341

m4b:
I'm going to close this in a month or so if there isn't any resolution here.

Afaics:

1. The 2string implementation isn't correct, and needs to not print processor specific entries
2. One proper way to do this is given in the example linked, by passing in the processor type so that it can print processor specific ranges correctly.
3. MIPS constants are desired, and in the meantime, a suitable fixup of this PR for least common denominator is just to print the non-processor specific values, and add all the appropriate values from the link that philipc gave.

m4b:
Closing, no activity; please re-open if someone wants to fix thsi!

PR #93, opened 2018-07-16
Fix include paths for alloc crate on nightly

gz:


philipc:
You also need to fix the `alloc` mod in lib.rs when compiling with libstd. It'll need a `collections` submodule. This is a bit ugly, but the idea is we can simply delete the `alloc` mod once the `alloc` crate is stabilized.

Are you using this `alloc` support? I was aware of the breakage, and was waiting to see if there were going to be more changes due to stabilization, but it's fine to fix this now if you're using it.

philipc:
Actually, it doesn't need a `collections` submodule, it can `pub use std collections` instead.

gz:
Yes so I'm using the no_std/alloc feature in gimli which in turn relies on goblin. I can look at why the build breaks and fix it.

gz:
Actually, the crate compiles fine at 8bbbcf5 with `cargo +nightly build --no-default-features --features alloc`. Not sure what happened yesterday night when I compiled it... I'm going to close this, sorry for my confusion.

gz:
Hm ok this seems to be still an issue but it's rather due to some weird interaction with the object crate, I will file an issue.

PR #98, opened 2018-07-22
parsing PE(s) whose headers are remapped

tathanhdinh:
In studying a corner case of PE format: [dual PE header](https://media.blackhat.com/bh-us-11/Vuksan/BH_US_11_VuksanPericin_PECOFF_WP.pdf), I realize that the header can be remapped, then some PE parsers can be fooled in parsing such a PE.

The idea is that the section table can be manipulated so that it maps other part of the PE into the "already mapped" header, so the mapped header is overwritten. A proof-of-concept is [`foldedhdr.exe`](https://github.com/corkami/pocs/blob/master/PE/bin/foldedhdr.exe) (with [source code](https://github.com/corkami/pocs/blob/master/PE/foldedhdr.asm)). The section table will maps the file offset `0x200` into rva `0x1000`, but the header is beyond `0x1000`: the offset `0x1000` is the location of the `import data directory`; then when the PE is loaded, the data at `0x1000` is overwritten by one at `0x200`.

In reading the PE loaders of [Wine](https://github.com/wine-mirror/wine/blob/master/dlls/ntdll/virtual.c#L1360) and of [ReactOS](https://doxygen.reactos.org/d8/dd5/peloader_8c.html#a6ec50c32ab426768ca06bddbf9d58cb8), they seem to me that the loaders mainly use the section table to load the PE into memory but do not try to parse more header's information from the file.

Currently, `goblin` does not consider that the PE header can be remapped. When parsing `foldedhdr.exe`, it rejects this PE since the virtual address `0x86600010` of the import directory cannot be found in any section (actually, this value becomes `0x10e0` when the PE is loaded into the memory), while the PE is totally valid (it executes on Windows and even on Wine).

Adlce PeViewer cannot parse the IAT of `foldedhdr.exe` (the right image), but only when the loaded PE is dumped (the left one).

foldhdr

I've tested other parsers (Adlice PeViewer, FileAlyzer, PPEE, CFF Explorer, LIEF (a library)), unfortunately none of them can correctly parse this PE.

This PR tries to handle this problem, and a small problem in calculating the [address range](https://github.com/m4b/goblin/blob/master/src/pe/utils.rs#L16) of an section, the size of a mapped section is not `VirtualSize` (neither `SizeOfRawData`), but somehow [more](https://github.com/wine-mirror/wine/blob/master/dlls/ntdll/virtual.c#L1463) [sophisticated](https://reverseengineering.stackexchange.com/questions/4324/reliable-algorithm-to-extract-overlay-of-a-pe). For example, goblin rejects [`quine.exe`](https://github.com/corkami/pocs/blob/master/PE/bin/quine.exe) since the `VirtualSize` of the section is 0, but it will accepts this PE with this PR.

I do not know whether the problem of remapped PE header is worth to be fixed (since none of tested PE parses handles it), but many thanks for any suggestion.


Cargo.toml
```diff
@@ -18,6 +18,7 @@ env_logger = "0.5"
 
 [dependencies]
 plain = "0.2.3"
+memoffset = "0.2.1"
```
m4b: I haven't reviewed the rest of PR, and I'm definitely interested, but I can tell you that I do not want this (or probably any other) external dependency, and it'll have to go before I consider it.

tathanhdinh: @m4b many thanks for your feedback.
I didn't know that you don't want any more dependency, I've changed the PR to fix that (just a rewrite of `offset_of!` though).

tathanhdinh:
Hi @m4b 
Just ping you a little bit to know whether you have time to look at the PR?

philipc:
What's the use case for this? Analyzing malware and handmade executables? Is it possible that there are use cases that would not want this behaviour?

An idea for an alternative way of handling this is to leave goblin unchanged, and instead require the caller to first use goblin to load the PE into memory, and then use goblin a second time to analyze the in-memory copy.

I don't have a strong opinion on which way is best.

m4b:
@tathanhdinh Hi, I'll try to review this weekend; please ping me again then just in case I forget. It's a pretty big PR though; so to save everyone's time, similar to last time, make sure the commits are in good shape, it's rebased on master, CI is passing, and only the minimal required, technically acceptable changes are made to the public facing api.

If its in that shape, I'll give it a look, but i can see that CI is already failing. I'll also note similar to @philipc i'm unsure of the usecase, primarily because I'm not sure exactly what this PR does; but that doesn't mean it doesn't do something good! it just means i don't know, mostly because I don't know a great deal about the more finer points of PE binaries :)

Otherwise maybe paste a little explanation for why we'd want this, why it's good, etc.  And as always, thank you for your time and energy!

tathanhdinh:
Hi @philipc and @m4b 
Many thanks for your comments. I will try to answer some of your questions:

- Q. What's is the use case for this? 
- A. The use cases are PEs whose the memory section containing the header is mapped from some other part (rather than the part from the offset 0). They are _normal PE by definition_ since there is no restriction saying that the PE header cannot not be remapped (the system loaders can indeed execute such a PE, e.g.  the binaries given in my previous post).

The schema is like that:

```
    +-----------+ - - - - - - - - >+-----------+
    |           |                  |           |
    |           |                  |           |
    |           |                  | (mapped)  |
    |           |                  |  header   |
    |  header   |                  |           |
    |           |        +-------->+-----------|
    |           |        |         |           |
    |           |        |         |           |
    +-----------+ - - - - - - - - ->           +
    |           |        |         | section i |
    |           |        |    +--->+-----------|
    |           |        |    |    |           |
    |           |        |    |    |           |
    +-----------+--------+    |    | ........  |
    |           |             |    |           |
    |           |             |    +-----------+
    |           |             |    |           |
    | section i |             |    |           |
    +-----------+-------------+    |           |
    |           |                  |           |
    |           |                  | section j |
    |           |                  +-----------+
    |           |
    | section j |
    +-----------+

      on disk                        on memory

```
where the `section i`, when being mapped into the memory, overwrites the  `header`. The system loader parses PE's information using the mapped header (and not the header on disk).

- Q. ...leave goblin unchanged, ...first use goblin to load the PE into memory, and then use goblin a second time to analyze the in-memory copy
- A. This is a nice approach indeed, as far as I know [libpeconv](https://github.com/hasherezade/libpeconv) follows this approach. Currently, goblin does not allow loading the PE into memory, should we implement a PE loader in goblin?

 - Q. ...CI is already failing
 - A. Sorry, I will check this.

NB. The PR indeed tries to handle two problems: the remapped header and the valid address range of a mapped section. The latter is much simpler to review, do you want me to split it into a separated PR?

PR #100, opened 2018-08-06
Get proper name for over-8-character sections

### LINKED ISSUE #99: Pe SectionTable::name should parse MinGW-style names
If a section has a name longer than 8 (for instance, all the .debug sections), MinGW will give that section the name "/4", where "/" indicates the "real" section name can be found in the string table, and 4 indicates the offset in the string table.

Goblin should find about this, and return the correct name.

Question: this requires whoever will be responsible for this to know about the string table location, which means having access to CoffHeader somehow. Should we just then the CoffHeader to SectionTable::parse?

roblabla:
Fixes #99 

An early PR to get some reviews, as there are probably better ways to do this. I need to implement the base64 kind of section table names, and fix the error handling in some places.

### DIFF
--- a/src/pe/mod.rs
+++ b/src/pe/mod.rs
@@ -59,8 +59,9 @@ impl<'a> PE<'a> {
         let offset = &mut (header.dos_header.pe_pointer as usize + header::SIZEOF_COFF_HEADER + header.coff_header.size_of_optional_header as usize);
         let nsections = header.coff_header.number_of_sections as usize;
         let mut sections = Vec::with_capacity(nsections);
+        let string_table_offset = header.coff_header.pointer_to_symbol_table + header.coff_header.number_of_symbol_table * 18;
         for i in 0..nsections {
-            let section = section_table::SectionTable::parse(bytes, offset)?;
+            let section = section_table::SectionTable::parse(bytes, offset, string_table_offset as usize)?;
             debug!("({}) {:#?}", i, section);
             sections.push(section);
         }
--- a/src/pe/mod.rs
+++ b/src/pe/mod.rs
@@ -18,6 +18,9 @@ mod utils;
 use error;
 use container;
 
+/// Size of a single symbol in the COFF Symbol Table.
+const COFF_SYMBOL_SIZE: u32 = 18;
+
 #[derive(Debug)]
 /// An analyzed PE32/PE32+ binary
 pub struct PE<'a> {
@@ -59,7 +62,8 @@ impl<'a> PE<'a> {
         let offset = &mut (header.dos_header.pe_pointer as usize + header::SIZEOF_COFF_HEADER + header.coff_header.size_of_optional_header as usize);
         let nsections = header.coff_header.number_of_sections as usize;
         let mut sections = Vec::with_capacity(nsections);
-        let string_table_offset = header.coff_header.pointer_to_symbol_table + header.coff_header.number_of_symbol_table * 18;
+        // Note that if we are handling a BigCoff, the size of the symbol will be different!
+        let string_table_offset = header.coff_header.pointer_to_symbol_table + header.coff_header.number_of_symbol_table * COFF_SYMBOL_SIZE;
         for i in 0..nsections {
             let section = section_table::SectionTable::parse(bytes, offset, string_table_offset as usize)?;
             debug!("({}) {:#?}", i, section);
--- a/src/pe/section_table.rs
+++ b/src/pe/section_table.rs
@@ -19,6 +19,27 @@ pub struct SectionTable {
 
 pub const SIZEOF_SECTION_TABLE: usize = 8 * 5;
 
+// Based on https://github.com/llvm-mirror/llvm/blob/af7b1832a03ab6486c42a40d21695b2c03b2d8a3/lib/Object/COFFObjectFile.cpp#L70
+// Decodes a string table entry in base 64 (//AAAAAA). Expects string without
+// prefixed slashes.
+fn base64_decode_string_entry(s: &str) -> Result<usize, ()> {
+    assert!(s.len() <= 6, "String too long, possible overflow.");
+
+    let mut val = 0;
+    for c in s.bytes() {
+        let v = match c {
+            b'A'..=b'Z' => c - b'A' + 00, // 00..=25
+            b'a'..=b'z' => c - b'a' + 26, // 26..=51
+            b'0'..=b'9' => c - b'0' + 52, // 52..=61
+            b'+'        => 62,            // 62
+            b'/'        => 63,            // 63
+            _           => return Err(())
+        };
+        val = val * 64 + v as usize;
+    }
+    Ok(val)
+}
+
 impl SectionTable {
     pub fn parse(bytes: &[u8], offset: &mut usize, string_table_offset: usize) -> error::Result<Self> {
         let mut table = SectionTable::default();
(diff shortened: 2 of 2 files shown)


src/pe/section_table.rs
```diff
@@ -35,10 +37,25 @@ impl SectionTable {
...
         table.number_of_linenumbers = bytes.gread_with(offset, scroll::LE)?;
         table.characteristics = bytes.gread_with(offset, scroll::LE)?;
+
+        // Based on https://github.com/llvm-mirror/llvm/blob/af7b1832a03ab6486c42a40d21695b2c03b2d8a3/lib/Object/COFFObjectFile.cpp#L1054
+        if name[0] == b'/' {
+            let idx: usize = if name[1] == b'/' {
+                // TODO: Base-64 encoding
+                panic!("At the disco")
+            } else {
+                name[1..].pread::<&str>(0)?.parse().unwrap()
```
m4b: `pread::<&str>(1)` is preferred; also no unwraps, please.

src/pe/section_table.rs
```diff
@@ -35,10 +37,25 @@ impl SectionTable {
         table.number_of_relocations = bytes.gread_with(offset, scroll::LE)?;
         table.number_of_linenumbers = bytes.gread_with(offset, scroll::LE)?;
         table.characteristics = bytes.gread_with(offset, scroll::LE)?;
+
+        // Based on https://github.com/llvm-mirror/llvm/blob/af7b1832a03ab6486c42a40d21695b2c03b2d8a3/lib/Object/COFFObjectFile.cpp#L1054
+        if name[0] == b'/' {
+            let idx: usize = if name[1] == b'/' {
+                // TODO: Base-64 encoding
+                panic!("At the disco")
```
m4b: I assume this panic (at the disco) will disappear in the final PR?

src/pe/section_table.rs
```diff
@@ -35,10 +37,25 @@ impl SectionTable {
...
+            } else {
+                name[1..].pread::<&str>(0)?.parse().unwrap()
+            };
+            table.real_name = Some(bytes.pread::<&str>(string_table_offset + idx)?.to_string());
```
m4b: since you're writing directly into a field whose type is statically known, i believe you can remove the `&str` annotation and be super cool.

It might also be a good time to make this struct zero copy w.r.t. strings, but maybe who cares? since this probably isn't super common and copying some section names isn't a big deal.

src/pe/section_table.rs
```diff
@@ -35,10 +37,25 @@ impl SectionTable {
...
+        }
         Ok(table)
     }
+
     pub fn name(&self) -> error::Result<&str> {
-        Ok(self.name.pread(0)?)
+        match self.real_name.as_ref() {
```
m4b: Do you need this as_ref() here?

src/pe/mod.rs
```diff
@@ -59,8 +59,9 @@ impl<'a> PE<'a> {
         let offset = &mut (header.dos_header.pe_pointer as usize + header::SIZEOF_COFF_HEADER + header.coff_header.size_of_optional_header as usize);
         let nsections = header.coff_header.number_of_sections as usize;
         let mut sections = Vec::with_capacity(nsections);
+        let string_table_offset = header.coff_header.pointer_to_symbol_table + header.coff_header.number_of_symbol_table * 18;
```
m4b: where does this magic 18 come from? I believe it should be a `const` with a (good) name, or at least a comment explaining it

roblabla: This is the size_of a single symbol in the table. Do we have a struct representing a Symbol in goblin ? I hardcoded 18, which is the common case scenario. But if we're handling a BigObj COFF (https://github.com/llvm-mirror/llvm/blob/af7b1832a03ab6486c42a40d21695b2c03b2d8a3/lib/Object/COFFObjectFile.cpp#L704), then that'll be wrong. 

I believe goblin doesn't support bigobj coffs, though. I guess I'll put a comment about this, and put it in a const.

src/pe/section_table.rs
roblabla: Yes, I run into borrow errors otherwise. The second as_ref in the Some case isn't necessary though.

src/pe/section_table.rs
roblabla: The type annotation is necessary, rust can't infer the type otherwise.

WRT zero-copy, I could do this pretty easy, just have to tuck a lifetime on the struct. Would you prefer this approach?

For what it's worth, this pattern is pretty common on MinGW binaries. Most of their sections are setup this way.

roblabla:
CI is still failing, but I think it's unrelated to me? It's complaining about lazy_static using unstable features on 1.19. lazy_static is a dependency of env_logger -> regex -> thread_local.

<details><summary>graph</summary>
graph
</details>

m4b:
Thanks for the graph! That’s it, I’m done with env logger. Second time transitive deps have broken us; I’m tired of this shit.

m4b:
heh; i assumed the update branch button would rebase your commits on top; but it does not... what a pain in the ass

m4b:
So are we good to go for this? @philipc anyone else?

philipc:
LGTM. We're not validating that the offsets are within the string table (the string table size is stored at the start of the table), not sure if that matters much.

src/pe/section_table.rs
```diff
@@ -18,13 +19,40 @@ pub struct SectionTable {
 
 pub const SIZEOF_SECTION_TABLE: usize = 8 * 5;
 
+// Based on https://github.com/llvm-mirror/llvm/blob/af7b1832a03ab6486c42a40d21695b2c03b2d8a3/lib/Object/COFFObjectFile.cpp#L70
+// Decodes a string table entry in base 64 (//AAAAAA). Expects string without
+// prefixed slashes.
+fn base64_decode_string_entry(s: &str) -> Result<usize, ()> {
+    assert!(s.len() <= 6, "String too long, possible overflow.");
```
m4b: I don't think this should assert unless it's a fundamental programmer error somehow. Return an Err is better

philipc:
The offsets aren't a blocker, but maybe put a todo comment?

src/pe/section_table.rs
philipc: I thought the same, but looking into it I don't think it's possible for this to assert because the section name is at most 8 bytes, and 2 of those are slashes.

roblabla: This should not return an error. If this assert is triggered, then something went terribly wrong.

roblabla:
Maybe parse should pass a string table slice instead of a string table offset ? That would take care of ensuring we stay within the table. The first 4 bytes of the string table should be its size (though I'm not sure if it's enforced by the PE loader or anything).

philipc:
Yeah I have no idea if the PE loader enforces it, and it's not like we're going to panic, so let's leave it.

philipc:
@m4b Is the out-of-date branch check necessary? Or should I be clicking that `Update branch` button and adding more merge commits...?

m4b:
The out of date button doesn’t rebase as I would expect so I don’t want to use it anymore. Probably should have PR authors rebase on master and force push latest before merging. In this case can just merge I think and it’ll be fine

m4b:
Hah i just got “server error” when trying to merge; dunno

philipc:
My "Merge pull request" button is disabled. It seems like you have disabled "Allow merge commits" in the settings. Is that correct? Requiring authors to rebase their PR just because another PR was merged first seems cumbersome to me.

m4b:
Yea agreed, ok i removed that from settings (had no idea it was even enabled)

PR #101, opened 2018-08-09
Calculate number of read bytes for sections using file alignment

tathanhdinh:
Hello all,

The  `pointer_to_raw_data` of each section should be [rounded down](https://reverseengineering.stackexchange.com/questions/4324/reliable-algorithm-to-extract-overlay-of-a-pe) to a multiple of 512. Moreover, the [valid address range](https://github.com/m4b/goblin/blob/master/src/pe/utils.rs#L15) of each section should be [calculated](https://reverseengineering.stackexchange.com/questions/4324/reliable-algorithm-to-extract-overlay-of-a-pe) using also `size_of_raw_data` and `file_alignment`, e.g. [quine.exe](https://github.com/corkami/pocs/blob/master/PE/bin/quine.exe) has `virtual_size` (of its unique section) is zero, but it is a valid PE.

To make the review becomes easier, I split the [previous PR](https://github.com/m4b/goblin/pull/98) (which is closed), into parts.

Many thanks for any comments.

### DIFF
--- a/src/pe/utils.rs
+++ b/src/pe/utils.rs
@@ -4,22 +4,54 @@ use error;
 
 use super::section_table;
 
+use std::cmp;
+
 pub fn is_in_range (rva: usize, r1: usize, r2: usize) -> bool {
     r1 <= rva && rva < r2
 }
 
+// reference: Peter Ferrie. Reliable algorithm to extract overlay of a PE. https://bit.ly/2vBX2bR
+#[inline]
+fn aligned_pointer_to_raw_data(pointer_to_raw_data: usize) -> usize {
+    const PHYSICAL_ALIGN: usize = 0x1ff;
+    pointer_to_raw_data & !PHYSICAL_ALIGN
+}
+
+#[inline]
+fn section_read_size(section: &section_table::SectionTable, file_alignment: u32) -> usize {
+    fn round_size(size: usize) -> usize {
+        const PAGE_MASK: usize = 0xfff;
+        (size + PAGE_MASK) & !PAGE_MASK
+    }
+
+    let file_alignment = file_alignment as usize;
+    let size_of_raw_data = section.size_of_raw_data as usize;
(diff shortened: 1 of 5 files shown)


m4b:
heh; i assumed the update branch button would rebase your commits on top; but it does not... what a pain in the ass

m4b:
I think this looks good to me, but I don't really have the bandwidth at the moment to fully review this. @philipc @willglynn (or anyone else) do you have any comments here? I'm inclined to just merge since it seems more or less fine.

src/pe/utils.rs
```diff
@@ -8,18 +8,58 @@ pub fn is_in_range (rva: usize, r1: usize, r2: usize) -> bool {
...
+    let size_of_raw_data = section.size_of_raw_data as usize;
+    let virtual_size = section.virtual_size as usize;
+    let read_size = {
+        let read_size = (section.pointer_to_raw_data as usize + size_of_raw_data + file_alignment - 1) & !(file_alignment - 1);
+        let raw_data_round_size = round_size(size_of_raw_data);
+        if read_size < raw_data_round_size {
+            read_size
+        } else {
+            raw_data_round_size
+        }
```
philipc: It is clearer to do: `cmp::min(read_size, round_size(size_of_raw_data))` (and same for virtual size).

philipc:
Looks fine, assuming the linked stackoverflow answer is correct (which is not something I can judge, this is all undocumented, in fact it's disallowed by the documentation...). From what I've read, this only applies to PE executables, not COFF objects, so this will need changing if we reuse this code for COFF.

src/pe/utils.rs
tathanhdinh: @philipc  thanks for your comments. It's is indeed clearer with `cmp::min`

philipc:
@tathanhdinh Can you rebase this to remove the extra 3 commits?

Issue #67, opened 2017-12-24
ELF: add relocation iterators


m4b:
similar to symbol, add relocation iterators, for maximum laziness. This will give immediate perf results for clients reading large binaries, but not needing the relocations

PR #102, opened 2018-09-28
Implemented a solution for #67

ibabushkin:
I think the provided iterator implementation and boilerplate does what is needed to lazily parse relocations.

### DIFF
--- a/src/elf/reloc.rs
+++ b/src/elf/reloc.rs
@@ -253,9 +253,9 @@ pub mod reloc64 {
 // Generic Reloc
 /////////////////////////////
 if_alloc! {
+    use scroll::{ctx, Pread};
     use core::fmt;
     use core::result;
-    use scroll::ctx;
     use container::{Ctx, Container};
     #[cfg(feature = "endian_fd")]
     use alloc::vec::Vec;
@@ -389,6 +389,35 @@ if_alloc! {
                     self.r_sym,
                 )
             }
+        }
+    }
+
+    pub struct RelocIterator<'a> {
+        bytes: &'a [u8],
+        offset: usize,
+        index: usize,
+        count: usize,
+        ctx: RelocCtx,
+    }
+
+    impl<'a> Iterator for RelocIterator<'a> {
+        type Item = Reloc;
+
+        #[inline]
+        fn next(&mut self) -> Option<Self::Item> {
+            if self.index >= self.count {
+                None
+            } else {
+                self.index += 1;
+                Some(self.bytes.gread_with(&mut self.offset, self.ctx).unwrap())
+            }
+        }
+    }
+
+    impl<'a> ExactSizeIterator for RelocIterator<'a> {
+        #[inline]
+        fn len(&self) -> usize {
+            self.count - self.index
         }
     }
 } // end if_alloc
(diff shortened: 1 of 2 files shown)


src/elf/reloc.rs
```diff
@@ -391,4 +380,111 @@ if_alloc! {
             }
         }
     }
+
+    #[derive(Default)]
+    pub struct RelocSection<'a> {
```
m4b: Doc comments please :)

src/elf/reloc.rs
```diff
@@ -391,4 +380,111 @@ if_alloc! {
...
+                .field("range", &format!("{:#x}..{:#x}", self.start, self.end))
+                .field("count", &self.count)
+                .field("Relocations", &self.to_vec())
+                .finish()
+        }
+    }
+
+    impl<'a> RelocSection<'a> {
+        #[cfg(feature = "endian_fd")]
+        pub fn parse(bytes: &'a [u8], offset: usize, filesz: usize, is_rela: bool, ctx: Ctx) -> ::error::Result<RelocSection<'a>> {
```
m4b: Doc comment

src/elf/reloc.rs
```diff
@@ -391,4 +380,111 @@ if_alloc! {
...
+            })
+        }
+
+        /// Try to parse a single relocation from the binary, at `index`.
+        #[inline]
+        pub fn get(&self, index: usize) -> Option<Reloc> {
+            if index >= self.count {
+                None
+            } else {
+                Some(self.bytes.pread_with(index * Reloc::size_with(&self.ctx), self.ctx).unwrap())
```
m4b: Because constructor checks the size, we're safe to unwrap here and in the iterator, correct?

ibabushkin: Yep, pretty much.

ibabushkin:
Alright, I rebased on the new master and added the doc comments you asked for :)

m4b:
@ibabushkin Thanks so much! So the PR is techncially backwards incompatible; i used as_slice a couple of times, but I don't think its that big of a deal.

However, this one is strange:

```
error[E0631]: type mismatch in closure arguments                                                                                                      
   --> src/format_elf.rs:336:59                                                                                                                       
    |                                                                                                                                                 
336 |         let num_shdr_relocs = self.elf.shdr_relocs.iter().fold(0, &|acc, &(_, ref v): &(usize, Vec<_>)| acc + v.len());                         
    |                                                           ^^^^     -------------------------------------------------- found signature of `for<'r> fn(_, &'r (usize, std::vec::Vec<_>)) -> _`
    |                                                           |                                                                                     
    |                                                           expected signature of `fn({integer}, &(usize, goblin::elf::reloc::RelocSection<'_>)) -> _`
    |                                                                                                                                                 
    = note: required because of the requirements on the impl of `std::ops::FnMut<({integer}, &(usize, goblin::elf::reloc::RelocSection<'_>))>` for `&[closure@src/format_elf.rs:336:68: 336:118]`
```

I probably just didn't read it carefully; @philipc are you ok with these changes? Is it going to be a pain for object?

m4b:
To be clear I don't care much about moving from Vec to iterator, as we need lazy iteration, and this is great!

Just not missing anything from the api it is presenting :)

m4b:
@ibabushkin Thanks for the PR, decided to just merge :)  We need lazy relocation iterators

ibabushkin:
Great! Glad to help.

PR #107, opened 2018-10-06
Handle multiple LOAD program headers with different biases

### LINKED ISSUE #106: Can't open some ELF files (numpy binaries for examples)
```
cd /tmp
mkdir numpy
wget https://files.pythonhosted.org/packages/40/c5/f1ed15dd931d6667b40f1ab1c2fe1f26805fc2b6c3e25e45664f838de9d0/numpy-1.15.2-cp27-cp27mu-manylinux1_x86_64.whl
unzip numpy-1.15.2-cp27-cp27mu-manylinux1_x86_64.whl
rdr /tmp/numpy/./numpy/core/multiarray.so
Malformed entity: Strtable size (9556) + offset (4280848) is out of bounds for 2070376 #bytes. Overflowed: false
```

philipc:
Partial fix for #106 

Note that this changes `elf::dyn::DynamicInfo` so that it is no longer the same as `elf::dyn::dyn32::DynamicInfo`.

### DIFF
--- a/src/elf/dyn.rs
+++ b/src/elf/dyn.rs
@@ -277,7 +277,6 @@ if_alloc! {
     use core::result;
     use container::{Ctx, Container};
     use strtab::Strtab;
-    use self::dyn32::{DynamicInfo};
     use alloc::vec::Vec;
 
     #[derive(Default, PartialEq, Clone)]
@@ -362,7 +361,7 @@ if_alloc! {
     impl Dynamic {
         #[cfg(feature = "endian_fd")]
         /// Returns a vector of dynamic entries from the underlying byte `bytes`, with `endianness`, using the provided `phdrs`
-        pub fn parse(bytes: &[u8], phdrs: &[::elf::program_header::ProgramHeader], bias: usize, ctx: Ctx) -> ::error::Result<Option<Self>> {
+        pub fn parse(bytes: &[u8], phdrs: &[::elf::program_header::ProgramHeader], ctx: Ctx) -> ::error::Result<Option<Self>> {
             use scroll::ctx::SizeWith;
             use scroll::Pread;
             use elf::program_header;
@@ -381,8 +380,7 @@ if_alloc! {
                     }
                     let mut info = DynamicInfo::default();
                     for dyn in &dyns {
-                        let dyn: dyn32::Dyn = dyn.clone().into();
-                        info.update(bias, &dyn);
+                        info.update(phdrs, dyn);
                     }
                     let count = dyns.len();
                     return Ok(Some(Dynamic { dyns: dyns, info: info, count: count }));
@@ -463,36 +461,6 @@ macro_rules! elf_dyn_std_impl {
                 }
             }
 
-            impl fmt::Debug for DynamicInfo {
-                fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-                    let gnu_hash = if let Some(addr) = self.gnu_hash { addr } else { 0 };
-                    let hash = if let Some(addr) = self.hash { addr } else { 0 };
-                    let pltgot = if let Some(addr) = self.pltgot { addr } else { 0 };
-                    write!(f, "rela: 0x{:x} relasz: {} relaent: {} relacount: {} gnu_hash: 0x{:x} hash: 0x{:x} strtab: 0x{:x} strsz: {} symtab: 0x{:x} syment: {} pltgot: 0x{:x} pltrelsz: {} pltrel: {} jmprel: 0x{:x} verneed: 0x{:x} verneednum: {} versym: 0x{:x} init: 0x{:x} fini: 0x{:x} needed_count: {}",
-                           self.rela,
-                           self.relasz,
-                           self.relaent,
-                           self.relacount,
-                           gnu_hash,
-                           hash,
-                           self.strtab,
-                           self.strsz,
-                           self.symtab,
-                           self.syment,
-                           pltgot,
-                           self.pltrelsz,
-                           self.pltrel,
-                           self.jmprel,
-                           self.verneed,
-                           self.verneednum,
-                           self.versym,
-                           self.init,
-                           self.fini,
-                           self.needed_count,
-                    )
-                }
-            }
-
             /// Returns a vector of dynamic entries from the given fd and program headers
             #[cfg(feature = "std")]
             pub fn from_fd(mut fd: &File, phdrs: &[$phdr]) -> Result<Option<Vec<Dyn>>> {
@@ -547,6 +515,23 @@ macro_rules! elf_dyn_std_impl {
                 needed
             }
         }
+    };
+}
+
+macro_rules! elf_dynamic_info_std_impl {
+    ($size:ident, $phdr:ty) => {
+        /// Convert a virtual memory address to a file offset
+        fn vm_to_offset(phdrs: &[$phdr], address: $size) -> Option<$size> {
+            for ph in phdrs {
+                if address >= ph.p_vaddr {
+                    let offset = address - ph.p_vaddr;
+                    if offset < ph.p_memsz {
+                        return ph.p_offset.checked_add(offset );
+                    }
+                }
+            }
+            None
+        }
 
         /// Important dynamic linking info generated via a single pass through the `_DYNAMIC` array
         #[derive(Default)]
@@ -587,34 +572,34 @@ macro_rules! elf_dyn_std_impl {
 
         impl DynamicInfo {
             #[inline]
-            pub fn update(&mut self, bias: usize, dyn: &Dyn) {
+            pub fn update(&mut self, phdrs: &[$phdr], dyn: &Dyn) {
                 match dyn.d_tag as u64 {
-                    DT_RELA => self.rela = dyn.d_val.wrapping_add(bias as _) as usize, // .rela.dyn
+                    DT_RELA => self.rela = vm_to_offset(phdrs, dyn.d_val).unwrap_or(0) as usize, // .rela.dyn
                     DT_RELASZ => self.relasz = dyn.d_val as usize,
                     DT_RELAENT => self.relaent = dyn.d_val as _,
                     DT_RELACOUNT => self.relacount = dyn.d_val as usize,
-                    DT_REL => self.rel = dyn.d_val.wrapping_add(bias as _) as usize, // .rel.dyn
+                    DT_REL => self.rel = vm_to_offset(phdrs, dyn.d_val).unwrap_or(0) as usize, // .rel.dyn
                     DT_RELSZ => self.relsz = dyn.d_val as usize,
                     DT_RELENT => self.relent = dyn.d_val as _,
                     DT_RELCOUNT => self.relcount = dyn.d_val as usize,
-                    DT_GNU_HASH => self.gnu_hash = Some(dyn.d_val.wrapping_add(bias as _)),
-                    DT_HASH => self.hash = Some(dyn.d_val.wrapping_add(bias as _)) as _,
-                    DT_STRTAB => self.strtab = dyn.d_val.wrapping_add(bias as _) as usize,
+                    DT_GNU_HASH => self.gnu_hash = vm_to_offset(phdrs, dyn.d_val),
+                    DT_HASH => self.hash = vm_to_offset(phdrs, dyn.d_val),
+                    DT_STRTAB => self.strtab = vm_to_offset(phdrs, dyn.d_val).unwrap_or(0) as usize,
                     DT_STRSZ => self.strsz = dyn.d_val as usize,
-                    DT_SYMTAB => self.symtab = dyn.d_val.wrapping_add(bias as _) as usize,
+                    DT_SYMTAB => self.symtab = vm_to_offset(phdrs, dyn.d_val).unwrap_or(0) as usize,
                     DT_SYMENT => self.syment = dyn.d_val as usize,
-                    DT_PLTGOT => self.pltgot = Some(dyn.d_val.wrapping_add(bias as _)) as _,
+                    DT_PLTGOT => self.pltgot = vm_to_offset(phdrs, dyn.d_val),
                     DT_PLTRELSZ => self.pltrelsz = dyn.d_val as usize,
                     DT_PLTREL => self.pltrel = dyn.d_val as _,
-                    DT_JMPREL => self.jmprel = dyn.d_val.wrapping_add(bias as _) as usize, // .rela.plt
-                    DT_VERNEED => self.verneed = dyn.d_val.wrapping_add(bias as _) as _,
+                    DT_JMPREL => self.jmprel = vm_to_offset(phdrs, dyn.d_val).unwrap_or(0) as usize, // .rela.plt
+                    DT_VERNEED => self.verneed = vm_to_offset(phdrs, dyn.d_val).unwrap_or(0),
                     DT_VERNEEDNUM => self.verneednum = dyn.d_val as _,
-                    DT_VERSYM => self.versym = dyn.d_val.wrapping_add(bias as _) as _,
-                    DT_INIT => self.init = dyn.d_val.wrapping_add(bias as _) as _,
-                    DT_FINI => self.fini = dyn.d_val.wrapping_add(bias as _) as _,
-                    DT_INIT_ARRAY => self.init_array = dyn.d_val.wrapping_add(bias as _) as _,
+                    DT_VERSYM => self.versym = vm_to_offset(phdrs, dyn.d_val).unwrap_or(0),
+                    DT_INIT => self.init = vm_to_offset(phdrs, dyn.d_val).unwrap_or(0),
+                    DT_FINI => self.fini = vm_to_offset(phdrs, dyn.d_val).unwrap_or(0),
+                    DT_INIT_ARRAY => self.init_array = vm_to_offset(phdrs, dyn.d_val).unwrap_or(0),
                     DT_INIT_ARRAYSZ => self.init_arraysz = dyn.d_val as _,
-                    DT_FINI_ARRAY => self.fini_array = dyn.d_val.wrapping_add(bias as _) as _,
+                    DT_FINI_ARRAY => self.fini_array = vm_to_offset(phdrs, dyn.d_val).unwrap_or(0),
                     DT_FINI_ARRAYSZ => self.fini_arraysz = dyn.d_val as _,
                     DT_NEEDED => self.needed_count += 1,
                     DT_FLAGS => self.flags = dyn.d_val as _,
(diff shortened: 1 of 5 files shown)


src/elf/dyn.rs
```diff
@@ -624,17 +609,52 @@ macro_rules! elf_dyn_std_impl {
...
                 for dyn in dynamic {
-                    info.update(bias, &dyn);
+                    info.update(phdrs, &dyn);
                 }
                 info
             }
-        } // end if_std
+        }
+
+        if_alloc! {
```
m4b: Just curious: how come this got moved to alloc?

src/elf/mod.rs
```diff
@@ -347,7 +324,6 @@ if_sylvan! {
                 is_64: is_64,
                 is_lib: is_lib,
                 entry: entry as u64,
-                bias: bias as u64,
```
m4b: hmmm; I don't like to see this field go, but given now we're supporting multiple biases we have to. Or maybe I'm wrong and its useless? Still might be nice to expose the iteration macro as a function hanging off the Elf type for users to pass an unbiased value in and get biased value out?

m4b:
Running this code on patch, I'm seeing a lot of what could be garbage (indicative of bad bias calculations?):

```
                st_name: 256160347 UNKNOWN_STB OBJECT st_other: 40 st_shndx: 10544 st_value: a3416c12eaa1bae1 st_size: 4343740557238714799,
                st_name: 1231268623 GLOBAL FUNC st_other: 13 st_shndx: 2453 st_value: f5267cdcea69ca7 st_size: 17000724336414452430,
                st_name: 322171258 UNKNOWN_STB OBJECT st_other: 52 st_shndx: 39099 st_value: 5d8533ab0f5267d3 st_size: 938747773043678002,
                st_name: 3467145249 UNKNOWN_STB UNKNOWN_STT st_other: 202 st_shndx: 27083 st_value: 93ac3a015d805dba st_size: 7622048091411052839,
                st_name: 3793167682 UNKNOWN_STB TLS st_other: 105 st_shndx: 41898 st_value: 397f9993f4c22f23 st_size: 14891172709048592475,
                st_name: 625681951 UNKNOWN_STB UNKNOWN_STT st_other: 130 st_shndx: 46492 st_value: 304b1a28162944b7 st_size: 14891172735670931007,
                st_name: 810239313 UNKNOWN_STB COMMON st_other: 161 st_shndx: 21404 st_value: ceaef39821260a17 st_size: 578240182440425589,
                st_name: 2746643623 UNKNOWN_STB UNKNOWN_STT st_other: 81 st_shndx: 3629 st_value: 7614838dc57e59b6 st_size: 8196851217518267431,
                st_name: 3976518154 UNKNOWN_STB UNKNOWN_STT st_other: 248 st_shndx: 56631 st_value: a423f7dfa3b674ad st_size: 870023370323442660,
                st_name: 1870234271 UNKNOWN_STB GNU_IFUNC st_other: 142 st_shndx: 29772 st_value: 3c481138ceabc1e3 st_size: 1381626612315706365,
                st_name: 3573891993 UNKNOWN_STB TLS st_other: 4 st_shndx: 9547 st_value: e8e78c35563cdbc6 st_size: 11014612366144553449,
                st_name: 1011355967 UNKNOWN_STB UNKNOWN_STT st_other: 137 st_shndx: 3622 st_value: a09c27a1f87554dc st_size: 14893380077069025244,
                st_name: 257056982 UNKNOWN_STB UNKNOWN_STT st_other: 161 st_shndx: 23163 st_value: 27dbfbbfceafea46 st_size: 1019653916751153770,
                st_name: 2753820660 UNKNOWN_STB OBJECT st_other: 206 st_shndx: 24942 st_value: 5d7df065ceafea4a st_size: 85899345924,
                st_name: 3 UNKNOWN_STB NUM st_other: 78 st_shndx: 85 st_value: 6165aa2043794870 st_size: 17404001873853089997,
                st_name: 3401280212 LOCAL NOTYPE st_other: 0 st_shndx: 0 st_value: 0 st_size: 0,
                st_name: 0 LOCAL NOTYPE st_other: 0 st_shndx: 0 st_value: 0 st_size: 0,
                st_name: 0 LOCAL NOTYPE st_other: 0 st_shndx: 0 st_value: 0 st_size: 0,
                st_name: 0 LOCAL NOTYPE st_other: 0 st_shndx: 0 st_value: 0 st_size: 0,
                st_name: 0 LOCAL NOTYPE st_other: 0 st_shndx: 0 st_value: 0 st_size: 0,
                st_name: 0 LOCAL NOTYPE st_other: 0 st_shndx: 0 st_value: 0 st_size: 0,
                st_name: 0 LOCAL NOTYPE st_other: 0 st_shndx: 0 st_value: 0 st_size: 0,
                st_name: 0 LOCAL NOTYPE st_other: 0 st_shndx: 0 st_value: 0 st_size: 0,
```

Similarly, when i run bingrep using the new patch (it's much better than before, and gets past the program headers), but fails printing the remainder, in particular the size of a note is bad (and I'm sure if it got to the symbols, it would fail there as well.

Before:
```
Malformed entity: Strtable size (9556) + offset (4280848) is out of bounds for 2070376 #bytes. Overflowed: false
```

```
m4b@efrit ::  [ ~/projects/bingrep ] RUST_LOG=goblin=debug cargo run -- ../goblin/numpy/core/multiarray.so
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s                                                                                         
     Running `target/debug/bingrep ../goblin/numpy/core/multiarray.so`
ELF DYN X86_64-little-endian @ 0x1eee0:

e_phoff: 0x40 e_shoff: 0x1f2040 e_flags: 0x0 e_ehsize: 64 e_phentsize: 56 e_phnum: 9 e_shentsize: 64 e_shnum: 28 e_shstrndx: 23

ProgramHeaders(9):
  Idx   Type              Flags   Offset      Vaddr       Paddr       Filesz      Memsz       Align     
  0     PT_LOAD           R+X     0x0         0x0         0x0         0x1d905c    0x1d905c    0x200000  
  1     PT_GNU_STACK      RW      0x0         0x0         0x0         0x0         0x0         0x10      
  2     PT_NOTE           R       0x1c8       0x1c8       0x1c8       0x24        0x24        0x4       
  3     PT_GNU_EH_FRAME   R       0x1a0fa4    0x1a0fa4    0x1a0fa4    0x8a94      0x8a94      0x4       
  4     PT_LOAD           RW      0x1d9060    0x3d9060    0x3d9060    0x18e60     0x18e60     0x200000  
  5     PT_TLS            R       0x1d9060    0x3d9060    0x3d9060    0x0         0x0         0x10      
  6     PT_LOAD           RW      0x1f3000    0x411000    0x411000    0x3010      0x3010      0x1000    
  7     PT_DYNAMIC        RW      0x1f7000    0x415000    0x415000    0x210       0x210       0x8       
  8     PT_LOAD           RW      0x1f7000    0x415000    0x415000    0x2768      0x2768      0x1000    
DEBUG:<unknown>: NoteIterator - 0x1c8
DEBUG:<unknown>: NoteHeader { n_namesz: 2, n_descsz: 6, n_type: 2060288 } - 0xc
DEBUG:<unknown>: note name  - 0x10
DEBUG:<unknown>: desc [0, 80, 65, 0, 0, 0] - 0x18
DEBUG:<unknown>: NoteIterator - 0x1e0
DEBUG:<unknown>: NoteHeader { n_namesz: 4280320, n_descsz: 0, n_type: 528 } - 0xc
type is too big (4280319) for 2069884
```

src/elf/dyn.rs
```diff
@@ -547,6 +515,23 @@ macro_rules! elf_dyn_std_impl {
                 needed
             }
         }
+    };
+}
+
+macro_rules! elf_dynamic_info_std_impl {
+    ($size:ident, $phdr:ty) => {
+        /// Convert a virtual memory address to a file offset
+        fn vm_to_offset(phdrs: &[$phdr], address: $size) -> Option<$size> {
```
m4b: I have some reservations about looping over the phdrs everytime an address lookup is requested; we really need to create an interval tree (and we could also use this in PE, which has very similar lookups).

I know you're trying to avoid allocation, but core dumps, for example, can have hundreds of phdrs, but then that might not ever have any bias calculations? I guess if this only occurs in the single pass over the dynamic info we're ok. I dunno, just some thoughts here, as opposed to allocating an associative list or a proper interval tree (which rust afaics doesn't have a robust one with a nice api unfortunately...)

src/elf/dyn.rs
philipc: It was always in alloc. I just moved it into another macro and kept it in alloc without thinking about whether it should be. I can move it out as part of this PR if you want, but it's probably better as another PR.

src/elf/mod.rs
philipc: Since there's no single bias, this field seems wrong or at least misleading to me. Exposing the iteration might be ok, but we can wait until someone needs it. The current iteration can't be part of the Elf type because we need it during construction of the Elf type, but any exposed version would be best as part of the Elf type.

src/elf/dyn.rs
philipc: I think it would be overkill to do that for the single pass over the dynamic info.

philipc:
> Running this code on patch, I'm seeing a lot of what could be garbage (indicative of bad bias calculations?):

No, this is because the dynamic info only has the start of the symtab, not it's size. The current calculation assumes it finishes at the strtab, and for this file, the symtab is at the start and the strtab is at the end. I guess we could do better by assuming it ends at the end of the program header containing its virtual address, it's still too long though so there will still be invalid symbols.

In my opinion the best fix is to use the `.dynsym` section if it exists.

philipc:
> in particular the size of a note is bad

Looks to me like the file really does have a bad size for the note.. or more likely, that note segment is garbage, because the previous note doesn't look correct either, and there is no corresponding note section to match it. Note that binutils 'readelf -n' doesn't display notes in NOTE segments unless it is a core file. So in my opinion the fix is bingrep should handle bad notes.

m4b:
Wow ok, will check if skipping the note can let me parse rest of binary. Just realized that the debug dump using rdr example can parse it, so might be ok. Let me get back to you; great work though, thanks!

I think we’ll publish a new 0.18 once this goes in, its a good feature :)

m4b:
@philipc are you sure the symbol table is correctly constructed in this new patch? I'm seeing complete garbage using both example=rdr and bingrep to parse the multiarray.so example:

<pre>  <font color="#CC0000">451e78c085fffff6 </font>   <span style="background-color:#34E2E2"><font color="#555753"><b>LOCAL      </b></font></span>   NOTYPE        <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0x4cda8948c931c031 </font>   <font color="#D3D7CF"><i>         BAD_IDX=40424</i></font>   0x0    
  <font color="#CC0000">c489411ff8c1ffff </font>   UNKNOWN_STB   UNKNOWN_STT   <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0xaed6e8fffffbf2e9 </font>   <font color="#D3D7CF"><i>         BAD_IDX=38294</i></font>   0xe8   
  <font color="#CC0000">7b2e058b48fffffd </font>   UNKNOWN_STB   NOTYPE        <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0xaebee8188b480034 </font>   <font color="#D3D7CF"><i>         BAD_IDX=33156</i></font>   0xf    
  <font color="#CC0000"> fc085fff8abb3e8 </font>   UNKNOWN_STB   NUM           <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0xbc86e8fffffd5f84 </font>   <font color="#D3D7CF"><i>         BAD_IDX=56969</i></font>   0x48   
  <font color="#CC0000">7c8b48fffffbb2e9 </font>   <span style="background-color:#EF2929"><font color="#555753"><b>GLOBAL     </b></font></span>   UNKNOWN_STT   <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0xff4824448b482024 </font>                            0x44   
  <font color="#CC0000">894118247c8b48ff </font>   UNKNOWN_STB   COMMON        <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0x8b48fff8a88ae8c4 </font>   <font color="#D3D7CF"><i>         BAD_IDX=65534</i></font>   0x3d   
  <font color="#CC0000">  1f0ffffffb82e9 </font>   UNKNOWN_STB   NOTYPE        <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0xc08548fff8ae63e8 </font>   .fini(9)                 0xe3   
  <font color="#CC0000"> 824448b49fffffd </font>   UNKNOWN_STB   UNKNOWN_STT   <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0x79e93050ffe7894c </font>   <font color="#D3D7CF"><i>          BAD_IDX=2537</i></font>   0xff   
  <font color="#CC0000">e9e93050ffe7894c </font>   UNKNOWN_STB   UNKNOWN_STT   <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0xfff8a9d0e8fffffc </font>   <font color="#D3D7CF"><i>          BAD_IDX=2084</i></font>   0x44   
  <font color="#CC0000">90fffffcd5e90009 </font>   <span style="background-color:#AD7FA8"><font color="#555753"><b>WEAK       </b></font></span>   NOTYPE        <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0xd3894853cd894855 </font>   <font color="#D3D7CF"><i>         BAD_IDX=58166</i></font>   0xe8   
  <font color="#CC0000">        2404c748 </font>   UNKNOWN_STB   UNKNOWN_STT   <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0x102444c748       </font>   <font color="#D3D7CF"><i>          BAD_IDX=1849</i></font>   0x6    
  <font color="#CC0000">ffff8420e8e28948 </font>   UNKNOWN_STB   UNKNOWN_STT   <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0x74c0854824048b48 </font>   <font color="#D3D7CF"><i>          BAD_IDX=4132</i></font>   0x4c   
  <font color="#CC0000">247cf7489948c3af </font>   UNKNOWN_STB   UNKNOWN_STT   <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0x48c0310045894810 </font>   <font color="#D3D7CF"><i>          BAD_IDX=3912</i></font>   0x32   
  <font color="#CC0000">          841f0f </font>   UNKNOWN_STB   UNKNOWN_STT   <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0xbb7504473904468b </font>   <font color="#D3D7CF"><i>         BAD_IDX=11878</i></font>   0xc3   
  <font color="#CC0000">   1ba10244c8b48 </font>   NUM           <font color="#FCE94F"><b>OBJECT     </b></font>   <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0x2948c3af0f480000 </font>   <font color="#D3D7CF"><i>         BAD_IDX=57323</i></font>   0xc0   
  <font color="#CC0000">c4834800458948f9 </font>   UNKNOWN_STB   UNKNOWN_STT   <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0x2e66c35d5bc03128 </font>   <font color="#D3D7CF"><i>         BAD_IDX=63304</i></font>   0x99   
  <font color="#CC0000">66a8ebffffffffb8 </font>   <span style="background-color:#34E2E2"><font color="#555753"><b>LOCAL      </b></font></span>   NOTYPE        <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0x841f0f           </font>                            0x0    
  <font color="#CC0000">4855cc89495441d5 </font>   UNKNOWN_STB   <font color="#FCE94F"><b>OBJECT     </b></font>   <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0x8348f3894853fd89 </font>   <font color="#D3D7CF"><i>         BAD_IDX=35137</i></font>   0x55   
  <font color="#CC0000">  c6f7000000a8b7 </font>   UNKNOWN_STB   UNKNOWN_STT   <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0xbf840f180000     </font>   <font color="#D3D7CF"><i>         BAD_IDX=35656</i></font>   0x8    
  <font color="#CC0000">      ab87f60000 </font>   <span style="background-color:#AD7FA8"><font color="#555753"><b>WEAK       </b></font></span>   FILE          <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0x102444c74810     </font>                            0x20   
  <font color="#CC0000">48de894901038348 </font>   UNKNOWN_STB   UNKNOWN_STT   <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0x24748d482024548d </font>                            0x1    
  <font color="#CC0000">   3b7880fc085ff </font>   UNKNOWN_STB   UNKNOWN_STT   <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0x85482024448b4800 </font>   <font color="#D3D7CF"><i>         BAD_IDX=63674</i></font>   0x27   
  <font color="#CC0000">   167840f03f883 </font>   <span style="background-color:#34E2E2"><font color="#555753"><b>LOCAL      </b></font></span>   <font color="#FCE94F"><b>OBJECT     </b></font>   <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0x8d4810247c8b4800 </font>   <font color="#D3D7CF"><i>         BAD_IDX=18432</i></font>   0x0    
  <font color="#CC0000">  302444c7480000 </font>   UNKNOWN_STB   UNKNOWN_STT   <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0x45e8ff3145000000 </font>   .rodata(10)              0xba   
  <font color="#CC0000">24442b483024448b </font>   UNKNOWN_STB   UNKNOWN_STT   <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0xf412024443b4810  </font>   <font color="#D3D7CF"><i>         BAD_IDX=18468</i></font>   0x4    
  <font color="#CC0000"> fff854500000384 </font>   <span style="background-color:#AD7FA8"><font color="#555753"><b>WEAK       </b></font></span>   UNKNOWN_STT   <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0xf63145000000a884 </font>   <font color="#D3D7CF"><i>         BAD_IDX=33807</i></font>   0x1    
  <font color="#CC0000"> 445c70000000e00 </font>   UNKNOWN_STB   COMMON        <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0x906623eb00000001 </font>   <font color="#D3D7CF"><i>         BAD_IDX=17863</i></font>   0x33   
  <font color="#CC0000"> 14e840fff007d83 </font>   UNKNOWN_STB   NOTYPE        <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0xf63145df89480000 </font>   <font color="#D3D7CF"><i>         BAD_IDX=12404</i></font>   0x1    
  <font color="#CC0000">f0894448c4834824 </font>   UNKNOWN_STB   UNKNOWN_STT   <font color="#FCE94F"><b>BAD NAME                            </b></font>   <font color="#4E9A06">0x5e415d415c415d5b </font>   <font color="#D3D7CF"><i>          BAD_IDX=1161</i></font>   0x49   </pre>

objdump doesn't appear to have any issues parsing the dynamic and symbol table, so I think we're in error here...

m4b:
Also we have an infinite loop in our note iterator implementation...

philipc:
objdump doesn't have a problem because it uses the `.dynsym` section, not the `DT_SYMTAB` value.

re the infinite loop, https://github.com/gimli-rs/object/pull/74 may be of interest.

m4b:
Well i'm shocked, had no idea objdump (and readelf!) relied on the section headers, which are technically strippable. We don't need to use .dynamic though, the dynamic linker has to know which dynamic symbols are available without section headers, so imho we should only use that (downstream clients can use the .dynamic section if they want manually, or we can implement a specific iterator over it if needed).

There just shouldn't be garbage coming out, it's just a bug in our implementation; i'm just trying to figure out if it was already present, or is a result of this patch? probably the former?

philipc:
The bug was already there. It's at [this line](https://github.com/m4b/goblin/blob/3dfb49944ab4d4a9915e3716cb55bb810c8674d1/src/elf/mod.rs#L302) where it is calculating the number of symbols. The problem is that without a `.dynsym` section, we don't know the size of the the dynamic symbol table, and that code makes a wild guess, which is wrong in this case. The dynamic linker doesn't need to know the size because it uses `DT_HASH` to index into it (at least, that's my understanding, this isn't something I've looked into in detail). So maybe what we could do is iterate over the hash table instead, if that's possible?

philipc:
https://flapenguin.me/2017/04/24/elf-lookup-dt-hash/ and https://flapenguin.me/2017/05/10/elf-lookup-dt-gnu-hash/ have good explanations of the hash tables. For `DT_HASH` we can directly use `nchains` as the number of symbols. For `DT_GNU_HASH`, we would have to walk the last bucket to determine its length, and add `symoffset` (the link has a longer description of this).

m4b:
Ok it sounds like that’s somewhat unrelated to this patch, and this is ready to go ? I’ll let you merge. Thanks for this and the subsequent discussion :)

philipc:
readelf will read the hash table instead if you use `readelf --syms -D`. It skips the undefined symbols at the start though.

PR #109, opened 2018-10-10
Use hash table to determine number of dynamic symbols

philipc:
Partial fix for #106

### DIFF
--- a/src/elf/mod.rs
+++ b/src/elf/mod.rs
@@ -276,7 +276,13 @@ if_sylvan! {
                 if dyn_info.needed_count > 0 {
                     libraries = dynamic.get_libraries(&dynstrtab);
                 }
-                let num_syms = if dyn_info.syment == 0 { 0 } else { if dyn_info.strtab <= dyn_info.symtab { 0 } else { (dyn_info.strtab - dyn_info.symtab) / dyn_info.syment }};
+                let num_syms = if let Some(gnu_hash) = dyn_info.gnu_hash {
+                    gnu_hash_len(bytes, gnu_hash as usize, ctx)?
+                } else if let Some(hash) = dyn_info.hash {
+                    hash_len(bytes, hash as usize, header.e_machine, ctx)?
+                } else {
+                    0
+                };
                 dynsyms = Symtab::parse(bytes, dyn_info.symtab, num_syms, ctx)?;
                 // parse the dynamic relocations
                 dynrelas = RelocSection::parse(bytes, dyn_info.rela, dyn_info.relasz, true, ctx)?;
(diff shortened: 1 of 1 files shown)


src/elf/mod.rs
```diff
@@ -276,7 +276,13 @@ if_sylvan! {
                 if dyn_info.needed_count > 0 {
                     libraries = dynamic.get_libraries(&dynstrtab);
                 }
-                let num_syms = if dyn_info.syment == 0 { 0 } else { if dyn_info.strtab <= dyn_info.symtab { 0 } else { (dyn_info.strtab - dyn_info.symtab) / dyn_info.syment }};
+                let num_syms = if let Some(gnu_hash) = dyn_info.gnu_hash {
+                    gnu_hash_len(bytes, gnu_hash as usize, ctx)?
+                } else if let Some(hash) = dyn_info.hash {
+                    hash_len(bytes, hash as usize, header.e_machine, ctx)?
+                } else {
+                    0
```
m4b: Ok, was going to ask; in case of a non-existent gnu/hash table, we effectively don't parse any symbols. I'm ok with this, but now I'm wondering (from a purely scientific curiosity):

1. Can dynamic libraries _not_ have a hash table
2. How does the dynamic linker compute dynsym size if no? I imagine they must have hash tables if they're dynamic libraries?

romainthomas: 1. No, dynamic libraries must have a hash table (SYSV or GNU_HASH). Nevertheless the table can be empty (c.f. https://github.com/lief-project/LIEF/blob/master/src/ELF/Builder.cpp#L95-L133) 
   Object file (.o) don't have hash table, neither dynamic symbols.
2. Actually it doesn't compute the number of symbols. It look for an index and check that this index is consistent

m4b: @romainthomas awesome, thanks for the clarification! It's super annoying to me the _DYNAMIC doesn't provide a `DT_SYMTABSZ`, like it does for basically everything else:

```
   DT_INIT_ARRAY 0x3d9060
 DT_INIT_ARRAYSZ 0x8
   DT_FINI_ARRAY 0x3d9068
 DT_FINI_ARRAYSZ 0x8
     DT_GNU_HASH 0x413548
       DT_STRTAB 0x415210
       DT_SYMTAB 0xc90
        DT_STRSZ 0x2554
       DT_SYMENT 0x18
```

src/elf/mod.rs
```diff
@@ -338,6 +344,44 @@ if_sylvan! {
             Ok((elf, src.len()))
         }
     }
+
+    fn gnu_hash_len(bytes: &[u8], offset: usize, ctx: Ctx) -> error::Result<usize> {
+        let buckets_num = bytes.pread_with::<u32>(offset, ctx.le)? as usize;
+        let min_chain = bytes.pread_with::<u32>(offset + 4, ctx.le)? as usize;
+        let bloom_size = bytes.pread_with::<u32>(offset + 8, ctx.le)? as usize;
```
m4b: looking again, `gread_with` would  have been better here to automate the offset computations, but will update another time :)

src/elf/mod.rs
m4b: thanks :) if you ever felt like swapping out a backend, I'd definitely be interested in spec-ing out/ designing some C bindings for goblin (though it will probably be a lot of work?)

PR #112, opened 2018-10-22
elf: use relocations to find number of dynamic symbols if hash is empty

### LINKED ISSUE #111: goblin 0.0.18 fails to parse dynsyms
I am parsing a simple binary (I have uploaded the binary here: http://reversing.io/goblin/loop-0 ).

I recently updated to goblin 0.0.18, and noticed some symbols were not being parsed. `Elf.dynsyms.len()` returns a length of zero. I ran the binary through `bingrep`, and dynsyms were parsed fine, but then noticed I was on `bingrep` `0.5.0`. I upgraded `bingrep` to the latest version, and now I get the following:

```
                 0    WEAK       NOTYPE      _ITM_registerTMCloneTable                0x0                             0x0    
                 0    WEAK       FUNC        __cxa_finalize@@GLIBC_2.2.5              0x0                             0x0    
               530    GLOBAL     FUNC        _init                                    0x0     .init(11)               0x0    

Dyn Syms(0):
  
thread 'main' panicked at 'index out of bounds: the len is 0 but the index is 0', /Users/travis/build/rust-lang/rust/src/libcore/slice/mod.rs:2085:10
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

I am troubleshooting through goblin, but thought I would share the issue now.

philipc:
Fixes #111

### DIFF
--- a/src/elf/mod.rs
+++ b/src/elf/mod.rs
@@ -290,14 +290,12 @@ if_sylvan! {
                 } else {
                     0
                 };
-                if num_syms == 0 {
-                    let max_reloc_sym = dynrelas.iter()
-                        .chain(dynrels.iter())
-                        .chain(pltrelocs.iter())
-                        .fold(0, |num, reloc| cmp::max(num, reloc.r_sym));
-                    if max_reloc_sym != 0 {
-                        num_syms = max_reloc_sym + 1;
-                    }
+                let max_reloc_sym = dynrelas.iter()
+                    .chain(dynrels.iter())
+                    .chain(pltrelocs.iter())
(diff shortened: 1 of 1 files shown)


philipc:
`gnu_hash_len` and `hash_len` already include imported symbols too, because imported symbols are always at the start of the table before the exported symbols. The problem is only when there are no exported symbols. If you want, we could do a `max(relocs, gnu_hash_len)`, but it should not be necessary.

philipc:
I may have been wrong about the imported symbols always being at the start... I couldn't find anything that says that. I've added another commit to always use the reloc symbols.

PR #114, opened 2018-11-02
Update for breaking changes in scroll

willglynn:
This PR updates `goblin` for the breaking changes in m4b/scroll#45 and m4b/scroll#47. It's mostly stripping out `Size`/`Units` and adding `&`.

There's a few places where `self` -> `&self` meant that `self.into()` became `&(*self.into())`. This change results in copying different structs – sometimes structs which could have been `#[derive(Copy)]` but weren't, so I added that where necessary.

### DIFF
--- a/examples/scroll.rs
+++ b/examples/scroll.rs
@@ -18,7 +18,7 @@ fn run () -> error::Result<()> {
     println!("header: {:?}", &header);
     // now lets write the header into some bytes
     let mut bytes = [0u8; elf64::header::SIZEOF_EHDR];
-    bytes.pwrite(header, 0)?;
+    bytes.pwrite(&header, 0)?;
     // read it back out
     let header2: elf64::header::Header = bytes.pread(0)?;
     // they're the same
--- a/src/elf/compression_header.rs
+++ b/src/elf/compression_header.rs
@@ -166,7 +166,7 @@ if_alloc! {
     use scroll::ctx;
     use container::{Container, Ctx};
 
-    #[derive(Default, PartialEq, Clone)]
+    #[derive(Default, PartialEq, Clone, Copy)]
     /// A unified CompressionHeader - convertable to and from 32-bit and 64-bit variants
     pub struct CompressionHeader {
         /// Compression format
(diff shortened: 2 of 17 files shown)


raindev:
This would be great to get in. I've hit the incompatibility attempting to build goblin against the head of scroll.

m4b:
Yes I’d like to see this go in too, just a matter of someone doing the work I think. I don’t think there were any outstanding semantic issues with the change (and taking a reference for writers is probably the more intuitive approach).

Iirc, the state of the PR was that it just needed to be finished, and then the PRs in faerie and goblin could be updated to test the functionality.

Eg basically this thread: https://github.com/m4b/scroll/pull/47

Iirc I think someone had a concern about annotated generics in turbofish position requiring a change, but because this is less common when `pwrite`-ing, not a super big deal ?

Otherwise @raindev afaik you should be able to build scroll master against goblin, since pwrite references haven’t landed; what issue are you bumping up against precisely ?

m4b:
Ah size associated type in SizeWith is also removed, I see. Well I think we could fix that separate to the PR and then maybe rebase @willglynn pr but that might have a ton of annoying conflicts

andrewshadura:
Any updates?

joshtriplett:
It would be nice to have goblin using an updated version of scroll, since scroll 0.9 uses older (pre-1.0.0) versions of proc_macro2, quote, and syn. Currently, using goblin alongside crates with newer proc macros will result in compiling multiple copies of those crates, which substantially increases compile time.

m4b:
Yup, will get that updated ASAP, probably this weekend, unless someone else wants to give it a shot ?

m4b:
And @joshtriplett to be clear this PR was for a PR that @willglynn prototyped for some semi major changes to scroll, namely having pwrite take references instead of ownership. 

I’m still holding out hope it’ll be fixed up one day in scroll and we can land some changes, there were some outstanding issues however.

Anyway updating to crates.io scroll 0.10 should be relatively straightforward I believe

andrewshadura:
Thanks @m4b.

joshtriplett:
@m4b Ah, sorry for the mixup.

m4b:
Just heads up, goblin with latest scroll was published as `0.1.0`; see #190

m4b:
closing as this is stale and and scroll has not merged change for pwrite to take by reference instead of owned version; all the scroll version changes should no longer be an issue, as we are on the latest scroll

PR #116, opened 2018-12-09
Public utils + more examples

### LINKED ISSUE #115: Need more examples
Hello guys.

I tried to use your library and it's api looks amazing. However, I don't see how to extract anything but very basic info.

I'm currently stuck getting CLI headers from PE executable file. I'm using [ECMA335](https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-335.pdf), section `II.25.2.3.3 PE header data directories`. I want to read `CLI Header` data so I'm writing following code:

```rust
use failure::{bail, Error, err_msg};
use goblin::pe::PE;
use goblin::pe::data_directories::DataDirectory;
use goblin::pe::section_table::SectionTable;

fn main() -> Result<(), Error> {
    let path = std::env::args()
        .skip(1)
        .next()
        .expect("Path to executable was not specified");
    let file = std::fs::read(path)?;
    let pe = PE::parse(&file)?;
    if pe.header.coff_header.machine != 0x14c {
        bail!("Is not a .Net executable");
    }
    let optional_header = pe.header.optional_header.ok_or_else(|| err_msg("No optional header"))?;
    let cli_header = optional_header.data_directories.data_directories[14].ok_or_else(|| err_msg("No CLI header"))?;
    let section = get_section(&pe, &cli_header)?;
    // what should I do here with section?
    Ok(())
}

fn get_section<'a>(pe: &'a PE, header: &DataDirectory) -> Result<&'a SectionTable, Error> {
    for section in pe.sections.iter() {
        if header.virtual_address >= section.virtual_address && header.virtual_address < section.virtual_address + header.size {
            return Ok(&section);
        }
    }
    bail!("Section for address {} was not found", header.virtual_address)
}
```

Being said, I don't see how it could be done.

OTOH, I have following working C code:

```c
#define T_CALLOC(type) ((type*)calloc(1, sizeof(type)))
#define PTR(p, ofs) ((void*)((char*)(p) + (ofs)))
#define VAL(type, p, ofs) (*(type*)(PTR(p, ofs)))
void* RVA_FindData(tFile *pFile, int addr) {
	tRVA *pRVA = pFile->pRVAs;
	while (pRVA != NULL) {
		if (addr >= pRVA->baseAddress && addr < pRVA->baseAddress + pRVA->size) {
			return PTR(pRVA->pData, addr - pRVA->baseAddress);
		}
		pRVA = pRVA->pNext;
	}
	return NULL;
}
   ...

	for (int i = 0; i < numSections; i++) {
		void *pSection = PTR(pSectionHeaders, i * 40);
		tRVA *pRVA = T_CALLOC(tRVA);
		pRVA->baseAddress = VAL(int, pSection, 12);
		pRVA->size = VAL(int, pSection, 8);
		pRVA->pData = calloc(1, pRVA->size);
		int rvaOfs = VAL(int, pSection, 20);
		memcpy(pRVA->pData, PTR(pData, rvaOfs), min(pRVA->size, VAL(int, pSection, 16)));
		pRVA->pNext = pFile->pRVAs;
		pFile->pRVAs = pRVA;
	}
	printf("Loaded sections :)\n");
	// Load CLI header
	void *pCLIHeader = RVA_FindData(pFile, VAL(int, pPEOptionalHeader, 208));
	printf("Runtime version: %i.%i\n", VAL(short, pCLIHeader, 4), VAL(short, pCLIHeader, 6));
   ...
```

So the question is basically how can I analyze dynamic file structure? Maybe some more examples?

There should be some way to utilize `pointer_to_raw_data`, but it's just an index, but I don't havy any buffer to index on.

Pzixel:
Fixes #115


examples/dotnet_pe_analysis.rs
```diff
@@ -0,0 +1,89 @@
...
+    let sections = &pe.sections;
+
+    let rva = cli_header.virtual_address as usize;
+    let offset = find_offset(rva, sections, file_alignment).ok_or_else(|| err_msg("Cannot map rva into offset"))?;
+    let cli_header_value: CliHeader = file.pread_with(offset, scroll::LE)?;
+
+    println!("{:#?}", cli_header_value);
+    let rva = cli_header_value.metadata.virtual_address as usize;
+    let offset = find_offset(rva, sections, file_alignment).ok_or_else(|| err_msg("Cannot map rva into offset"))?;
+    let root: MetadataRoot = file.pread_with(offset, scroll::LE)?;
```
Pzixel: I wonder if this may be enhanced as built-in function. Unfortunately, I ended up with multiple generics errors with code:

```
pub fn get_data<'a, T>(bytes: &'a [u8], sections: &[section_table::SectionTable], directory: &DataDirectory, file_alignment: u32) -> Result<T, DataDirectoryConversionError>
    where T: scroll::ctx::TryFromCtx<'a, scroll::Endian> {
    let rva = directory.virtual_address as usize;
    let offset = find_offset(rva, sections, file_alignment)
        .ok_or(DataDirectoryConversionError::MissingOffset(directory.virtual_address))?;
    let result: T = bytes.pread_with(offset, scroll::LE).map_err(DataDirectoryConversionError::ReadError)?;
    Ok(result)
}
```

Pzixel:
Also renamed CLR -> CLI to match documentation

image

Cargo.toml
```diff
@@ -11,6 +11,9 @@ documentation = "https://docs.rs/goblin"
 categories = ["parsing", "development-tools::debugging"]
 include = ["src/**/*", "Cargo.toml", "CHANGELOG.md", "LICENSE", "README.md", "etc/*", "examples/*", "tests/*", "fuzz/**/*"]
 
+[dev-dependencies]
+failure = "0.1.3"
```
m4b: I don’t want a dev dep for the example; various upstream deps have already broken our CI a few times and unwrap is sufficient.

kgv:
> Also renamed CLR -> CLI to match documentation
> 
> image

See official [microsoft documentation](https://docs.microsoft.com/ru-ru/windows/desktop/Debug/pe-format) it names CLR

Pzixel:
@kgv hmm, looks like there is some mess with them. However, I was based o ECMA335 (see [link](https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-335.pdf)) which sould be more precise source.

Cargo.toml
Pzixel: I removed it. However, could you be more detailed? This information could save me lots of hair in the future :)

kgv:
@Pzixel
Microsoft documentation has always been a mess.
However, ESMA-335 primarily concerns .NET specification. PE format is not only .NET.

In my opinion, to make changes that are consistent with the documentation and ESMA-335, you should send an issue to https://github.com/MicrosoftDocs, and then, when it is approved, change it in this repo.

Pzixel:
Yep, but ECMA-335 conserns CLI specification (not .Net as you said) and this header is specifically CLI header, so it must be the most competent in this question.

However, you are right that docs should be updated. I'm creating an issue

kgv:
Check it: https://github.com/MicrosoftDocs/feedback

m4b:
So what is the resolution (if any) for CLR/CLI data directory naming ? I would prefer to not make a breaking change and I used the original documentation that @kgv linked. We can always update it in a later PR if need be. And thanks for both your investigations here!

Cargo.toml
m4b: It was just becoming annoying to have deps for the sake of some small examples. Eg the rdr.rs file had env_logger dep, just because it made debugging it easier, but multiple times their downstream deps had upgraded their rustc and we didn’t so it broke our CI. It’s not that big a deal but I got tired of spurious failures from dependencies we don’t even use so I basically ripped them all out in a minor rage :)

Pzixel:
@m4b I created a PR in MS documentation repo, however, it keeps silent. It looks like it takes 2-3 days until someone react on it.

I can revert this change if you wish, OTOH I have strong feeling that CLI specification is more compenent in this question.

Pzixel:
I reverted rename change.

P.S. I wonder why you keep rust 1.20. It doesn't make sense to me since rust is pretty well backward compatible.

Pzixel:
Ping?

src/pe/utils.rs
```diff
@@ -74,3 +76,24 @@ pub fn try_name<'a>(bytes: &'a [u8], rva: usize, sections: &[section_table::Sect
...
+    where T: scroll::ctx::TryFromCtx<'a, scroll::Endian, Size = usize, Error = scroll::Error>  {
+    let rva = directory.virtual_address as usize;
+    let offset = find_offset(rva, sections, file_alignment)
+        .ok_or(DataDirectoryConversionError::MissingOffset(directory.virtual_address))?;
+    let result: T = bytes.pread_with(offset, scroll::LE).map_err(DataDirectoryConversionError::ReadError)?;
+    Ok(result)
+}
+
+#[derive(Debug)]
+pub enum DataDirectoryConversionError {
```
m4b: No more error enums please. It just makes working with code more annoying. It also doesn’t implement Error, etc. please use the goblin error; if you can’t reuse one of the errors without losing semantics, just use the Custom string option.

src/pe/utils.rs
```diff
@@ -74,3 +76,24 @@ pub fn try_name<'a>(bytes: &'a [u8], rva: usize, sections: &[section_table::Sect
         }
     }
 }
+
+pub fn get_data<'a, T>(bytes: &'a [u8], sections: &[section_table::SectionTable], directory: &DataDirectory, file_alignment: u32) -> Result<T, DataDirectoryConversionError>
```
m4b: This seems like it belongs as a method in DataDirectories ?

src/pe/utils.rs
```diff
@@ -74,3 +76,24 @@ pub fn try_name<'a>(bytes: &'a [u8], rva: usize, sections: &[section_table::Sect
...
+pub enum DataDirectoryConversionError {
+    MissingOffset(u32),
+    ReadError(scroll::Error)
+}
+
+impl Display for DataDirectoryConversionError {
+    fn fmt(&self, f: &mut Formatter) -> FmtResult {
+        Debug::fmt(self, f)
+    }
+}
```
m4b: No newline

src/pe/utils.rs
Pzixel: I'm not sure, because it belongs to SectionTable as well.. Utils is the good place for this imo

Pzixel:
Pong?

Issue #120, opened 2019-03-05
Crash on malformed ELF file


Shnatsel:
Attempting to decode any of the attached files with `goblin::elf::Elf::parse` crashes the process. Memory allocator runs out of virtual memory and the process is aborted.

[goblin-elf-oom-crashes.zip](https://github.com/m4b/goblin/files/2933592/goblin-elf-oom-crashes.zip)

Found via AFL.rs. Fuzzing harness: https://github.com/Shnatsel/goblin/blob/master/fuzz-afl/src/main.rs

m4b:
Many parts of the parser rely on the binary giving lengths to various variable length entities; I guess the best we can do here is to clamp the maximum size to the size of the binary?

Shnatsel:
Sounds like a great idea!

Alternatively you could let the API user set the limits, which is what the decompression libraries do. But if you know the size up front, it's probably best to just use that.

m4b:
@shnatsel do you want to make a PR :)

Shnatsel:
It would be interesting, but I'm afraid I'll have to decline. I already have two ongoing Rust projects that have higher priority, and one of them isn't even announced yet.

philipc:
In general, how does goblin want to handle errors: parse as much as it can, or return an error immediately when something is obviously wrong? For example, if the size of a section is larger than the size of the file, should it still try to parse whatever it can in that section? Or should it skip that section but keep trying to parse other sections? Or should it immediately return an error?

I can have a look at some of these. However, not all of them are giving errors for me.
Running:
```
for i in all-oom-crashes/*; do target/debug/examples/rdr $i 2&>1 >/dev/null && echo $i; done
```
gives:
<details>

```
all-oom-crashes/id:000000,sig:06,src:000001,op:flip1,pos:135
all-oom-crashes/id:000000,sig:06,src:000001,op:havoc,rep:32
all-oom-crashes/id:000000,sig:06,src:000001,op:havoc,rep:4
all-oom-crashes/id:000000,sig:06,src:000001,op:havoc,rep:64
all-oom-crashes/id:000001,sig:06,src:000001,op:flip1,pos:135
all-oom-crashes/id:000001,sig:06,src:000001,op:havoc,rep:128
all-oom-crashes/id:000001,sig:06,src:000001,op:havoc,rep:32
all-oom-crashes/id:000001,sig:06,src:000001,op:havoc,rep:64
[...]

PR #121, opened 2019-03-07
elf: handle some invalid sizes

philipc:
This fixes all the errors that I could reproduce from #120.

TODO:
- validate all other `Vec::with_capacity`
- validate all other arithmetic for overflows

### DIFF
--- a/src/elf/dyn.rs
+++ b/src/elf/dyn.rs
@@ -367,11 +367,17 @@ if_alloc! {
             use elf::program_header;
             for phdr in phdrs {
                 if phdr.p_type == program_header::PT_DYNAMIC {
+                    let offset = phdr.p_offset as usize;
                     let filesz = phdr.p_filesz as usize;
+                    // Ensure offset and filesz are valid.
+                    let bytes = bytes
+                        .pread_with::<&[u8]>(offset, filesz)
+                        .map_err(|_| ::error::Error::Malformed(format!("Invalid PT_DYNAMIC size (offset {:#x}, filesz {:#x})",
+                                                               offset, filesz)))?;
                     let size = Dyn::size_with(&ctx);
                     let count = filesz / size;
                     let mut dyns = Vec::with_capacity(count);
-                    let mut offset = phdr.p_offset as usize;
+                    let mut offset = 0;
                     for _ in 0..count {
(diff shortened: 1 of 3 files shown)


src/elf/dyn.rs
```diff
@@ -466,6 +472,7 @@ macro_rules! elf_dyn_std_impl {
             pub fn from_fd(mut fd: &File, phdrs: &[$phdr]) -> Result<Option<Vec<Dyn>>> {
                 for phdr in phdrs {
                     if phdr.p_type == PT_DYNAMIC {
+                        // FIXME: validate filesz before allocating
```
philipc: I don't know how to validate this. Probably we could do `fd.take(filesz).read_to_end()` (and then check the correct length was read?), but then need to convert that `Vec[u8]` into a `Vec[Dyn]`.

src/elf/sym.rs
```diff
@@ -500,7 +500,10 @@ if_alloc! {
     impl<'a> Symtab<'a> {
         /// Parse a table of `count` ELF symbols from `offset`.
         pub fn parse(bytes: &'a [u8], offset: usize, count: usize, ctx: Ctx) -> Result<Symtab<'a>> {
-            let size = count * Sym::size_with(&ctx);
+            let size = count
+                .checked_mul(Sym::size_with(&ctx))
+                .ok_or(::error::Error::Malformed(format!("Too many ELF symbols (offset {:#x}, count {})",
```
m4b: Are we 2018 in this crate ? We should update soon if not

src/elf/dyn.rs
m4b: I think this method is only used by things that load off disk like a dynamic linker so it shouldn’t be on the parser code path; however yes we should do something here but I wouldn’t worry about it for now

m4b:
I’ll let you push the merge button since it feels good :)

src/elf/sym.rs
philipc: Yep, so rust 2018 + rustfmt?

m4b: the only thing is we’ve had a really great story for supporting older compilers so I’d like to see what the Linux distros are shipping at the moment. If we can target a minimum rustc/ cargo combination which does 2018 that would be ideal. Need to investigate.

PR #124, opened 2019-03-12
Adds the clone to strtab

Skarlett:
This is useful when borrowing the `elf::Elf` in a function that returns a non-borrowed value from it and, trying to return moved memory.

```rs
fn pair_section_names(bin: &goblin::elf::Elf ) -> Vec<(goblin::elf::SectionHeader, String)>{
    bin.section_headers.clone()
        .into_iter()
        .zip(
            bin.shdr_strtab.clone()
                .to_vec()
                .unwrap()
                .into_iter()
        )
        .map(|(s, tabname)| (s, tabname.to_string()))
        .collect::<Vec<(goblin::elf::SectionHeader, String)>>()
}
```

### DIFF
--- a/src/strtab.rs
+++ b/src/strtab.rs
@@ -14,6 +14,7 @@ if_alloc! {
 /// A common string table format which is indexed by byte offsets (and not
 /// member index). Constructed using [`parse`](#method.parse)
 /// with your choice of delimiter. Please be careful.
+#[derive(Clone)]
 pub struct Strtab<'a> {
     bytes: &'a[u8],
     delim: ctx::StrCtx,


philipc:
You should be using the `sh_name` field instead of zipping. Something like:
```rust
    bin.section_headers
        .iter()
        .map(|s| {
            let name = if let Some(Ok(name)) = bin.shdr_strtab.get(s.sh_name) {
                name.to_string()
            } else {
                String::new()
            };
            (s.clone(), name)
        })
        .collect::<Vec<(goblin::elf::SectionHeader, String)>>()
```

m4b:
Yea I’m not sure why you’d want to clone the strtab in above case. However I’m not opposed to adding it necessarily. I do remember specifically not implementing it intentionally though but perhaps that reason is no longer valid

PR #125, opened 2019-03-22
PE: check IAT for import entries if lookup table is null

iximeow:
hi! In an interesting choice in the PE format, it's possible for an import thunk to have a null import lookup table, but still list imports. In that case, the loader can/will check the referenced IAT for RVAs of hints just as the lookup table would reference.

goblin doesn't know that, so PEs that have a null lookup table RVA in an import thunk just appear to have no imports. I don't have a particularly good test case to share, but RollerCoaster Tycoon Deluxe does this with its import thunks

from what I can tell, this change is correct, but I don't see any tests around PEs, so I just know RCT.EXE has imports now where it didn't before :D

edit: forgot the relevant bits of MSDN that I was going to include: the [Import Address Table](https://docs.microsoft.com/en-us/windows/desktop/Debug/pe-format#import-lookup-table) section mentions that lookup table and IAT have identical content, but I don't know anywhere linkable that mentions the loader-side detail that the lookup table may be null

### DIFF
--- a/src/pe/import.rs
+++ b/src/pe/import.rs
@@ -147,11 +147,17 @@ impl<'a> SyntheticImportDirectoryEntry<'a> {
         let name = utils::try_name(bytes, name_rva as usize, sections, file_alignment)?;
         let import_lookup_table = {
             let import_lookup_table_rva = import_directory_entry.import_lookup_table_rva;
-            debug!("Synthesizing lookup table imports for {} lib, with import lookup table rva: {:#x}", name, import_lookup_table_rva);
+            let import_address_table_rva = import_directory_entry.import_address_table_rva;
             if let Some(import_lookup_table_offset) = utils::find_offset(import_lookup_table_rva as usize, sections, file_alignment) {
+                debug!("Synthesizing lookup table imports for {} lib, with import lookup table rva: {:#x}", name, import_lookup_table_rva);
                 let import_lookup_table = SyntheticImportLookupTableEntry::parse::<T>(bytes, import_lookup_table_offset, sections, file_alignment)?;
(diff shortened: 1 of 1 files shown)


leo-lb:
I was about to make this change to the library too and found that PR.
It indeed should be merged, I've written PE parsers in C before and had to make the change.
AFAIK it is due to the fact that Borland compilers used to use a nil ILT with a filled in IAT and it happened to work with Windows's executable loader so as Windows does not want to break backwards compatibility they kept it that way even though it was not defined in the PE format this way in the first place.

There's a lot of hacks involved in parsing PE files because it has a _very_ long history of backwards compatibility.

leo-lb:
@m4b As a test, you know about UPX right? (https://github.com/upx/upx)
Pack a PE with it then unpack it and run it through your bingrep, then run it again through your bingrep with goblin dependency set to the updated version.
You will see the difference.
UPX when it unpacks compressed binaries only sets IAT values and not ITL values. Just like the Borland compiler (it's proprietary and I don't have it here), Delphi is also Borland IIRC so that too.

m4b:
Well if you all think it’s good, someone push the merge button!

leo-lb:
I am confused, are you not the repository owner? Can't you merge? o.O

m4b:
Yes but I don’t necessarily like to steal the endorphin rush when one gets to merge ones own PR :)

philipc:
@m4b I thought only collaborators with write access could merge.

m4b:
Hmmm, I don't know, I always thought if the PR was approved, anyone could merge. Ok I'll just merge them whenever I see them from now on :D

PR #129, opened 2019-03-24
Impl Debug for Rel

### LINKED ISSUE #24: needs `debug_struct`, etc., in Debug impls
E.g.:

```rust
        impl fmt::Debug for Rel {
                let sym = r_sym(self.r_info);
                write!(f,
                       self.r_offset,
                       typ,
                       sym
                )
            }
        }
```

lzutao:
cc #24

### DIFF
--- a/src/elf/reloc.rs
+++ b/src/elf/reloc.rs
@@ -82,7 +82,7 @@ macro_rules! elf_reloc {
                 f.debug_struct("Rela")
                     .field("r_offset", &format_args!("{:x}", self.r_offset))
                     .field("r_info", &self.r_info)
-                    .field("r_addend", &self.r_addend)
+                    .field("r_addend", &format_args!("{:x}", self.r_addend))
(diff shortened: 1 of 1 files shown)


src/elf/reloc.rs
```diff
@@ -79,24 +79,25 @@ macro_rules! elf_reloc {
...
                 let typ = r_type(self.r_info);
-                write!(f,
-                       "r_offset: {:x} r_typ: {} r_sym: {} r_addend: {:x}",
-                       self.r_offset,
-                       typ,
-                       sym,
-                       self.r_addend)
+                f.debug_struct("Rela")
+                    .field("r_offset", &format_args!("{:x}", self.r_offset))
+                    .field("r_info", &self.r_info)
```
philipc: I think `r_info` is better as hex too, since it is a bit field.

lzutao: Fixed now.

PR #128, opened 2019-03-23
impl Debug for RelocIterator

lzutao:
Fix #127, cc #24

PR #131, opened 2019-03-24
Beautify debugging by using debug_struct

### LINKED ISSUE #24: needs `debug_struct`, etc., in Debug impls

lzutao:
With #128 and #129, it should be sufficient to fix #24 .

### DIFF
--- a/src/elf/sym.rs
+++ b/src/elf/sym.rs
@@ -224,15 +224,14 @@ macro_rules! elf_sym_std_impl {
                     let bind = st_bind(self.st_info);
                     let typ = st_type(self.st_info);
                     let vis = st_visibility(self.st_other);
-                    write!(f,
-                           "st_name: {} {} {} {} st_shndx: {} st_value: {:x} st_size: {}",
-                           self.st_name,
-                           bind_to_str(bind),
-                           type_to_str(typ),
-                           visibility_to_str(vis),
-                           self.st_shndx,
-                           self.st_value,
-                           self.st_size)
+                    f.debug_struct("Sym")
+                        .field("st_name", &format_args!("{} {} {} {}", self.st_name, bind_to_str(bind), type_to_str(typ), visibility_to_str(vis)))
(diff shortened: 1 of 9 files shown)


lzutao:
r @philipc

src/elf/sym.rs
```diff
@@ -224,15 +224,14 @@ macro_rules! elf_sym_std_impl {
...
-                           "st_name: {} {} {} {} st_shndx: {} st_value: {:x} st_size: {}",
-                           self.st_name,
-                           bind_to_str(bind),
-                           type_to_str(typ),
-                           visibility_to_str(vis),
-                           self.st_shndx,
-                           self.st_value,
-                           self.st_size)
+                    f.debug_struct("Sym")
+                        .field("st_name", &format_args!("{} {} {} {}", self.st_name, bind_to_str(bind), type_to_str(typ), visibility_to_str(vis)))
```
philipc: These extra values aren't part of the name. Since you're displaying `st_info` and `st_other` now too, it might be better to move these to the field they are derived from. Same for the other `Sym` below.

src/elf/sym.rs
```diff
@@ -224,15 +224,14 @@ macro_rules! elf_sym_std_impl {
...
+                        .field("st_value", &format_args!("{:x}", self.st_value))
+                        .field("st_size", &self.st_size)
+                        .field("st_info", &self.st_info)
```
philipc: As in the other PR, this might be better as hex.

m4b: yes agreed

m4b:
Hey just wanted to say thanks for taking this up; it's not the most glorious work, but its definitely highly appreciated :)

PR #132, opened 2019-03-25
Beautify debugging by using debug_struct for mach

### LINKED ISSUE #24: needs `debug_struct`, etc., in Debug impls

lzutao:
Hopefully this one is enough to to fix #24 .

### DIFF
--- a/src/mach/symbols.rs
+++ b/src/mach/symbols.rs
@@ -112,13 +112,13 @@ pub const SIZEOF_NLIST_32: usize = 12;
 
 impl Debug for Nlist32 {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
-        write!(fmt, "strx: {:04} type: {:#02x} sect: {:#x} desc: {:#03x} value: {:#x}",
-               self.n_strx,
-               self.n_type,
-               self.n_sect,
-               self.n_desc,
-               self.n_value,
-        )
+        fmt.debug_struct("Nlist32")
+           .field("n_strx", &format_args!("{:04}", self.n_strx))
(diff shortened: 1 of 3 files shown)


src/mach/symbols.rs
```diff
@@ -405,18 +405,11 @@ impl<'a> Symbols<'a> {
...
-            match res {
-                Ok((name, nlist)) => {
-                    writeln!(fmt, "{: >10x} {} sect: {:#x} type: {:#02x} desc: {:#03x}", nlist.n_value, name, nlist.n_sect, nlist.n_type, nlist.n_desc)?;
-                },
-                Err(error) => {
-                    writeln!(fmt, "  Bad symbol, index: {}, sym: {:?}", i, error)?;
-                }
-            }
-        }
-        writeln!(fmt, "}}")
```
lzutao: Should we keep this? And how?

philipc: The elf symtab [prints them as a vec](https://github.com/m4b/goblin/blob/4c24bc5a615960b99a6eb80b36c34497f99ece96/src/elf/sym.rs#L493). We use this as a lazy way of implementing the `rdr` example, and it's handy for it to print out the symbols too, so I think we should keep it.

lzutao: Fixed.

philipc: Looking again, we don't actually use this for the `rdr` example (`MachO` creates a vec for its symbols instead), but I think this is still fine.

PR #134, opened 2019-03-26
Update to Rust 2018 edition

lzutao:
By running `cargo +nightly fix --edition --edition-idioms`.


lzutao:
The next PR would be running `cargo fmt`.

src/elf/dyn.rs
```diff
@@ -317,35 +317,35 @@ if_alloc! {
...
         type Size = usize;
         fn try_into_ctx(self, bytes: &mut [u8], Ctx { container, le}: Ctx) -> result::Result<Self::Size, Self::Error> {
             use scroll::Pwrite;
             match container {
                 Container::Little => {
-                    let dyn: dyn32::Dyn = self.into();
-                    Ok(bytes.pwrite_with(dyn, 0, le)?)
+                    let r#dyn: dyn32::Dyn = self.into();
+                    Ok(bytes.pwrite_with(r#dyn, 0, le)?)
                 },
```
lzutao: This file and these variables need to be renamed. Could you suggest a new name for it?

m4b: Wow the file needs to be renamed ? Why? This will be a backwards incompatible change :(

lzutao: `dyn` is keyword in 2018 edition. You still could use it by raw keyword `r#dyn`.

m4b: Oh yea. And I guess it makes sense we can’t use keywords in filenames because they become nodules ? :(

@philipc what do you think ? Probably rename to Dynamic or something :(

I’m sad

lzutao: Yes, we can. But when we import module, we must write something like `use parent::r#dyn::Object` instead.

src/elf/dyn.rs
```diff
@@ -660,27 +660,27 @@ macro_rules! elf_dynamic_info_std_impl {
...
 
-    elf_dyn_std_impl!(u32, ::elf32::program_header::ProgramHeader);
-    elf_dynamic_info_std_impl!(u32, ::elf::program_header::program_header32::ProgramHeader);
+    elf_dyn_std_impl!(u32, crate::elf32::program_header::ProgramHeader);
+    elf_dynamic_info_std_impl!(u32, crate::elf::program_header::program_header32::ProgramHeader);
 }
 
 pub mod dyn64 {
-    pub use elf::dyn::*;
+    pub use crate::elf::r#dyn::*;
```
lzutao: An example is this.

src/elf/compression_header.rs
```diff
@@ -71,9 +71,9 @@ macro_rules! elf_compression_header_std_impl { ($size:ty) => {
             #[cfg(feature = "std")]
             pub fn from_fd(fd: &mut File, offset: u64) -> Result<CompressionHeader> {
                 let mut chdr = CompressionHeader::default();
-                try!(fd.seek(Start(offset)));
+                r#try!(fd.seek(Start(offset)));
                 unsafe {
-                    try!(fd.read(plain::as_mut_bytes(&mut chdr)));
+                    r#try!(fd.read(plain::as_mut_bytes(&mut chdr)));
```
lzutao: I would update `try!` macro later. I hope I don't forget it.

src/elf/dyn.rs
philipc: We can keep deprecated aliases for the old names if you want to keep backwards compatibility. I do think we need to rename it though, using `r#dyn` in 2018 edition code isn't nice, that syntax is intended for interop with 2015 edition code. I don't have any better suggestions than `Dynamic`.

lzutao: I would do the rename in separate PR and update CHANGELOG in there.

m4b: Is Dyn also banned (capital C)?

So I suggest renaming to dyn_syms or dynamic (I don’t like dynamic personally), but perhaps we should bikeshed later

lzutao:
Should be ready to review again.

src/error.rs
```diff
@@ -24,27 +24,18 @@ pub enum Error {
...
 impl Error {
     pub fn description(&self) -> &str {
-        match *self {
-            #[cfg(feature = "std")]
-            Error::IO(_) => { "IO error" }
-            Error::Scroll(_) => { "Scroll error" }
-            Error::BadMagic(_) => { "Invalid magic number" }
-            Error::Malformed(_) => { "Entity is malformed in some way" }
-        }
+        "description() is deprecated; use to_string()"
```
lzutao: I don't know if any dependent crates using this method. But it might be deprecated in favor of `to_string`.

philipc: That's not a good way to do deprecations. They need to be handled during compilation.

lzutao: Fixed in 2a0c80a.

src/elf/dyn.rs
philipc: `Dyn` doesn't need to change.

The `dyn` module is for the `PT_DYNAMIC` segment/`.dynamic` section, not for dynamic symbols.

philipc:
I don't have time to give this a proper review currently, so please be patient, unless @m4b gets to it first.

m4b:
This is getting big :) I’ll review this weekend.

I think it’s already being done but best to punt the dyn issue to another pr (yes we’ll definitely rename to something but not now), along with other unusual changes.

This should just be getting the code compliant. And really appreciate your work here @Izutao!

tests/archive.rs
```diff
@@ -47,8 +46,11 @@ fn parse_archive() {
                 println!("could not get crt1.o");
                 assert!(false)
             }
-        },
-        Err(err) => {println!("could not parse archive: {:?}", err); assert!(false)}
+        }
+        Err(err) => {
+            println!("could not parse archive: {:?}", err);
+            assert!(false)
+        }
```
lzutao: Why this code? Isn't it just panic?

lzutao:
@m4b I pushed unusual changes to #135 .

m4b:
Let’s do this :) thanks for your hard work @izutao, and thanks for reviewing @philipc. This is a great milestone ! I’d like to seriously consider 1.0 in the near future, let’s discuss in another issue.

src/elf/dyn.rs
lzutao: Discussion will be moved to #135.

PR #135, opened 2019-03-28
Fix compliant after upgrading to Rust 2018 edition

lzutao:
This PR:
- [x] Update CHANGELOG to reflect breaking changes.
- [x] Discuss about renaming module `elf::dyn` due to `dyn` become keyword in 
      Rust 2018 edition.
- [x] Use `?` to replace `try!` macro.
- [x] Fix deprecated warning on `trim_{left|right}[matches]` methods.
- [x] error: Remove `Error::description`.

### DIFF
--- a/src/elf/dynamic.rs
+++ b/src/elf/dynamic.rs
@@ -323,7 +323,7 @@ if_alloc! {
         type Size = usize;
         fn try_from_ctx(bytes: &'a [u8], Ctx { container, le}: Ctx) -> result::Result<(Self, Self::Size), Self::Error> {
             use scroll::Pread;
-            let r#dyn = match container {
+            let dynamic = match container {
                 Container::Little => {
                     (bytes.pread_with::<dyn32::Dyn>(0, le)?.into(), dyn32::SIZEOF_DYN)
                 },
@@ -331,7 +331,7 @@ if_alloc! {
                     (bytes.pread_with::<dyn64::Dyn>(0, le)?.into(), dyn64::SIZEOF_DYN)
                 }
             };
-            Ok(r#dyn)
+            Ok(dynamic)
         }
     }
 
@@ -342,12 +342,12 @@ if_alloc! {
             use scroll::Pwrite;
             match container {
                 Container::Little => {
-                    let r#dyn: dyn32::Dyn = self.into();
-                    Ok(bytes.pwrite_with(r#dyn, 0, le)?)
+                    let dynamic: dyn32::Dyn = self.into();
+                    Ok(bytes.pwrite_with(dynamic, 0, le)?)
                 },
                 Container::Big => {
-                    let r#dyn: dyn64::Dyn = self.into();
-                    Ok(bytes.pwrite_with(r#dyn, 0, le)?)
+                    let dynamic: dyn64::Dyn = self.into();
+                    Ok(bytes.pwrite_with(dynamic, 0, le)?)
                 }
             }
         }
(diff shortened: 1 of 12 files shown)


src/error.rs
```diff
@@ -23,28 +23,26 @@ pub enum Error {
 }
 
 impl Error {
+    #[deprecated(since = "0.0.22", note = "Please use the to_string() method instead")]
```
philipc: Sorry, I didn't look into this properly last time. Just delete this method. The `Error` trait provides a default implementation.

src/error.rs
```diff
@@ -23,28 +23,26 @@ pub enum Error {
...
-    }
-    fn cause(&self) -> Option<&error::Error> {
+    fn source(&self) -> Option<&(dyn error::Error + 'static)> {
         match *self {
-            Error::IO(ref io) => { io.cause() }
-            Error::Scroll(ref scroll) => { scroll.cause() }
-            Error::BadMagic(_) => { None }
-            Error::Malformed(_) => { None }
+            Error::IO(ref io) => io.source(),
+            Error::Scroll(ref scroll) => scroll.source(),
```
philipc: Neither `io::Error` nor `scroll::Error` implement `source`, and I don't expect they ever will, so this code does nothing. Let's just delete this method.

philipc:
Need @m4b to tick off on the `dyn` rename too.

src/error.rs
lzutao: Did you mean to use blanket implementation of `error::Error`?

philipc: No. I don't think there is a blanket implementation? I meant to still implement `error::Error`, but use the provided default implementation for all methods. Note that I don't think there's anywhere in the crate that relies on the `error:Error` trait being implemented, but we still should. I don't know what best practices are since the description/cause changes, but as far as I can tell we're not gaining anything by implementing these methods currently.

philipc: Or, you could change this `source` method to return references to the io and scroll errors, instead of delegating. Still not sure that is useful though. I'm open to other opinions.

src/error.rs
lzutao: I think this would break compatibility if other dependent crates depends on it.
But more breaking change would be renaming `dyn` module.

philipc: You're right, and if it were version 1.0 then I would agree with keeping it. But it's version 0.0.22, and I think it's unlikely anyone is using this instead of `Display`.

m4b: Let me look at this, philipc might be right here

lzutao: Done.

src/elf/dynamic.rs
```diff
@@ -515,9 +515,9 @@ macro_rules! elf_dyn_std_impl {
             /// Gets the needed libraries from the `_DYNAMIC` array, with the str slices lifetime tied to the dynamic array/strtab's lifetime(s)
             pub unsafe fn get_needed<'a>(dyns: &[Dyn], strtab: *const Strtab<'a>, count: usize) -> Vec<&'a str> {
                 let mut needed = Vec::with_capacity(count);
-                for r#dyn in dyns {
-                    if r#dyn.d_tag as u64 == DT_NEEDED {
-                        let lib = &(*strtab)[r#dyn.d_val as usize];
+                for dynamic in dyns {
```
m4b: probably `d` was a better variable choice here given `d_tag`, and `for d in dyns` seems better, but whatever.

PR #136, opened 2019-03-29
Implement PE exception tables

jan-auer:
This implements parsing for the exception tables data directory and x64 unwind codes. Respective documentation from Microsoft is located here: https://docs.microsoft.com/en-us/cpp/build/exception-handling-x64?view=vs-2017.

The exception directory is only present on x64 builds, so the entire implementation can assume x64 exception handling.

Tests are still missing as I haven't taken the time to extract relevant bits from an actual executable for realistic testing yet.

### DIFF
--- a/src/pe/exception.rs
+++ b/src/pe/exception.rs
@@ -417,7 +417,17 @@ impl<'a> TryFromCtx<'a, UnwindOpContext> for UnwindCode {
                 let offset = bytes.gread_with::<u32>(&mut read, scroll::LE)?;
                 UnwindOperation::SaveXMM128(register, StackFrameOffset::with_ctx(offset, ctx))
             }
-            self::UWOP_PUSH_MACHFRAME => UnwindOperation::PushMachineFrame(operation_info == 1),
+            self::UWOP_PUSH_MACHFRAME => {
+                let is_error = match operation_info {
+                    0 => false,
+                    1 => true,
+                    i => {
+                        let msg = format!("invalid op info ({}) for UWOP_PUSH_MACHFRAME", i);
+                        return Err(error::Error::Malformed(msg));
+                    }
+                };
+                UnwindOperation::PushMachineFrame(is_error)
+            }
             op => {
                 let msg = format!("unknown unwind op code ({})", op);
                 return Err(error::Error::Malformed(msg));
--- a/src/pe/exception.rs
+++ b/src/pe/exception.rs
@@ -44,13 +44,13 @@ use core::fmt;
 use core::iter::FusedIterator;
 
 use scroll::ctx::TryFromCtx;
-use scroll::{self, Pread};
+use scroll::{self, Pread, Pwrite};
 
-use error;
+use crate::error;
 
-use pe::data_directories;
-use pe::section_table;
-use pe::utils;
+use crate::pe::data_directories;
+use crate::pe::section_table;
+use crate::pe::utils;
 
 /// The function has an exception handler that should be called when looking for functions that need
 /// to examine exceptions.
@@ -126,7 +126,7 @@ pub struct RuntimeFunctionIterator<'a> {
     data: &'a [u8],
 }
 
-impl<'a> Iterator for RuntimeFunctionIterator<'a> {
+impl Iterator for RuntimeFunctionIterator<'_> {
     type Item = error::Result<RuntimeFunction>;
 
     fn next(&mut self) -> Option<Self::Item> {
@@ -152,8 +152,8 @@ impl<'a> Iterator for RuntimeFunctionIterator<'a> {
     }
 }
 
-impl<'a> FusedIterator for RuntimeFunctionIterator<'a> {}
-impl<'a> ExactSizeIterator for RuntimeFunctionIterator<'a> {}
+impl FusedIterator for RuntimeFunctionIterator<'_> {}
+impl ExactSizeIterator for RuntimeFunctionIterator<'_> {}
 
 /// An x64 register used during unwinding.
 ///
@@ -444,7 +444,7 @@ pub struct UnwindCodeIterator<'a> {
     context: UnwindOpContext,
 }
 
-impl<'a> Iterator for UnwindCodeIterator<'a> {
+impl Iterator for UnwindCodeIterator<'_> {
     type Item = error::Result<UnwindCode>;
 
     fn next(&mut self) -> Option<Self::Item> {
@@ -463,7 +463,7 @@ impl<'a> Iterator for UnwindCodeIterator<'a> {
     }
 }
 
-impl<'a> FusedIterator for UnwindCodeIterator<'a> {}
+impl FusedIterator for UnwindCodeIterator<'_> {}
 
 /// A language-specific handler that is called as part of the search for an exception handler or as
 /// part of an unwind.
@@ -485,7 +485,7 @@ pub enum UnwindHandler<'a> {
 /// For unwinding, this link shall be followed until the root unwind info record has been resolved.
 ///
 /// [`unwind_codes`]: struct.UnwindInfo.html#method.unwind_codes
-/// [chained unwind handler]: struct.UnwindInfo.html#method.unwind_codes
+/// [chained unwind handler]: struct.UnwindInfo.html#structfield.chained_info
 #[derive(Clone)]
 pub struct UnwindInfo<'a> {
     /// Version of this unwind info.
@@ -618,7 +618,7 @@ impl<'a> UnwindInfo<'a> {
     }
 }
 
-impl<'a> fmt::Debug for UnwindInfo<'a> {
+impl fmt::Debug for UnwindInfo<'_> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         let count_of_codes = self.code_bytes.len() / UNWIND_CODE_SIZE;
 
@@ -634,7 +634,7 @@ impl<'a> fmt::Debug for UnwindInfo<'a> {
     }
 }
 
-impl<'a, 'b> IntoIterator for &'b UnwindInfo<'a> {
+impl<'a> IntoIterator for &'_ UnwindInfo<'a> {
     type Item = error::Result<UnwindCode>;
     type IntoIter = UnwindCodeIterator<'a>;
 
@@ -757,7 +757,7 @@ impl<'a> ExceptionData<'a> {
         &self,
         mut function: RuntimeFunction,
         sections: &[section_table::SectionTable],
-    ) -> error::Result<UnwindInfo> {
+    ) -> error::Result<UnwindInfo<'a>> {
         while function.unwind_info_address % 2 != 0 {
             let rva = (function.unwind_info_address & !1) as usize;
             function = self.get_function_at(rva, sections)?;
@@ -792,7 +792,7 @@ impl<'a> ExceptionData<'a> {
     }
 }
 
-impl<'a> fmt::Debug for ExceptionData<'a> {
+impl fmt::Debug for ExceptionData<'_> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         f.debug_struct("ExceptionData")
             .field("file_alignment", &self.file_alignment)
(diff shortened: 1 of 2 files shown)


src/pe/exception.rs
```diff
@@ -0,0 +1,815 @@
...
+    type Error = error::Error;
+    type Size = usize;
+
+    #[inline]
+    fn try_from_ctx(
+        bytes: &'a [u8],
+        ctx: UnwindOpContext,
+    ) -> Result<(Self, Self::Size), Self::Error> {
+        let mut read = 0;
+        let code_offset = bytes.gread_with::<u8>(&mut read, scroll::LE)?;
```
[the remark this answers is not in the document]
jan-auer: Yes, to my knowledge the only platform using big-endian was Xbox 360 and even then PEs were in little-endian (except for the code and data). However, since the rest of the PE code assumes little-endian too, I've opted for streamlining to the rest of the code base.

jan-auer:
Updated for compatibility with Rust 1.20.0. 

---

Even though it's not that much code, I'd like to offer a suggestion to sanity-check this:

 - `PE::exception_table` -> `ExceptionData::functions` -> `ExceptionData::get_unwind_info` -> `UnwindInfo::unwind_codes` -> `UnwindCode::operation`
 - `UnwindInfo::chained_info` -> `ExceptionData::get_unwind_info`. 
 - Comments on `StackFrameOffset` (used in some operations) regarding how to load register values

A usage example for this is my port of Breakpad's dump_syms [here](https://github.com/getsentry/symbolic/blob/bfe7e7bfc7d7f0e3fa6b4e34480f2551697331ff/minidump/src/cfi.rs#L469-L525).

---

There might be more, one could do to help interpreting the unwind info:

 - For one, I did not implement the new `UWOP_EPILOG` operation. Almost noone seems to interpret it correctly, and its behavior is quite complex. 
 - Next, there is nothing that "helps" with unwinding, e.g. an unwind context that keeps register state and restores register values. This could be a nice feature, but I think it's out of scope for this crate.
 - Finally, exception and termination handlers are only exposed as (unbounded) raw buffers. For my case, and probably most other cases, these won't be necessary. But if someone really wants to runtime-unwind, this would come in handy. Same applies as above, probably a good case for a separate crate.

For full transparency, here are two more thoughts that I had when implementing: 
 - The manually implemented binary search in `find_function` is not ideal. However, I found no ergonomic way in scroll to map a `&[u8]` to `&[T]`. Transmuting and implementing a custom `RuntimeFunction` struct that converts from little endian on the fly wasn't pretty either...
 - Only `RuntimeFunction` derives `Pwrite` at the moment, but it might be nice to also write the other unwind info eventually. I wasn't sure if that's necessary though and what the best pattern would be, especially for the `UnwindCodes` iterator.

philipc:
> Updated for compatibility with Rust 1.20.0.

Note that #134 is bumping to rust 1.31.1 and changing to 2018 edition.

src/pe/exception.rs
```diff
@@ -0,0 +1,810 @@
...
+/// no info, FP = RSP + UNWIND_INFO.FPRegOffset*16
+const UWOP_SET_FPREG: u8 = 3;
+/// info == register number, offset in next slot
+const UWOP_SAVE_NONVOL: u8 = 4;
+/// info == register number, offset in next 2 slots
+const UWOP_SAVE_NONVOL_FAR: u8 = 5;
+/// info == XMM reg number, offset in next slot
+const UWOP_EPILOG: u8 = 6;
+/// info == XMM reg number, offset in next 2 slots
+const UWOP_SPARE_CODE: u8 = 7;
```
philipc: These are confusing due to the version 1/2 differences and the mismatch between comments and constant names.

src/pe/exception.rs
```diff
@@ -0,0 +1,810 @@
...
+        }
+    }
+
+    /// Resolves unwind information for the given function entry.
+    pub fn get_unwind_info(
+        &self,
+        mut function: RuntimeFunction,
+        sections: &[section_table::SectionTable],
+    ) -> error::Result<UnwindInfo> {
+        while function.unwind_info_address % 2 != 0 {
```
philipc: What is this loop for? I didn't see anything in the Microsoft doc for `RUNTIME_FUNCTION`.

src/pe/exception.rs
```diff
@@ -0,0 +1,810 @@
...
+            let rva = (function.unwind_info_address & !1) as usize;
+            function = self.get_function_at(rva, sections)?;
+        }
+
+        let rva = function.unwind_info_address as usize;
+        let offset = utils::find_offset(rva, sections, self.file_alignment).ok_or_else(|| {
```
philipc: Can we store the RVA bias in `ExceptionData` instead of looking it up every time?

src/pe/exception.rs
```diff
@@ -0,0 +1,810 @@
...
+            error::Error::Malformed(format!("cannot map unwind rva ({:#x}) into offset", rva))
+        })?;
+
+        UnwindInfo::parse(self.bytes, offset)
+    }
+
+    fn get_function_at(
```
philipc: Rename to `get_function_by_rva`?

philipc:
> The manually implemented binary search in find_function is not ideal. However, I found no ergonomic way in scroll to map a &[u8] to &[T]. Transmuting and implementing a custom RuntimeFunction struct that converts from little endian on the fly wasn't pretty either...

Maybe transmute to `&[[u8; RUNTIME_FUNCTION_SIZE]]`?

src/pe/exception.rs
jan-auer: Funny enough, the original codes were deprecated and removed even in version one, and apparently not emitted anymore. I've updated the comments, that was oversight when copying stuff around.

src/pe/exception.rs
jan-auer: Yes, this seems to be missing from the Microsoft docs. Some other implementations do this -- for instance Breakpad's dump_syms. It seems that setting the last bit is a marker for a redirect. I've not tried to go to the bottom of it, though.

src/pe/exception.rs
jan-auer: Excellent point. I wasn't sure if the RVAs always point into the same section. This would make sense generally, but I don't trust this is the case for every PE. Especially the tooling Microsoft applies to their DLLs might invalidate this assumption. I'll try to investigate this.

jan-auer:
> Maybe transmute to `&[[u8; RUNTIME_FUNCTION_SIZE]]`?

Yeah, I was playing around with something similar for a moment... I'll let you choose what you like more 😄 

```rust
type Placeholder = [u32; 3];
let slice = unsafe { std::slice::from_raw_parts(
    self.bytes.as_ptr() as *const Placeholder,
    self.bytes.len() / RUNTIME_FUNCTION_SIZE,
) };

let index = match slice.binary_search_by_key(&rva, |p| u32::from_le(p[0])) {
    Ok(i) => i,
    Err(0) => return Ok(None),
    Err(i) => i - 1,
};

let function = self.get_function(index)?;
if function.end_address > rva {
    Ok(Some(function))
} else {
    Ok(None)
}
```

philipc:
Let's stick with the version that doesn't use unsafe.

src/pe/exception.rs
philipc: Yeah I think you're right. breakpad doesn't appear to do anything clever either.

m4b:
yes, I agree with @philipc , lets leave out unsafe.
>  However, I found no ergonomic way in scroll to map a &[u8] to &[T]. Transmuting and implementing a custom RuntimeFunction struct that converts from little endian on the fly wasn't pretty either...

I'm curious though, @jan-auer , so `gread_inout_with` allows you to pass a mutable buffer of type `&'mut [T]` where `T` is your type you want to deserialize out of the bytes you give it (and which has a `TryFromCtx` impl). It copies in the same host endianness case, but in practice i don't think this is generally much of an issue on most hardware these days.
I considered adding a pread method for essentially performing transmutes but several people convinced me not to, and i think it was a good idea not to in the end. If the user needs the extra perf boost, then they should transmute themselves, and guarantee the host <-> binary blob is the same endianness; on the other hand, if they're deserializing from arbitrary binary formats on disk, its generally not a great idea to be transmuting, and one should just copy out.
Anyway, great work here!

m4b:
> Only RuntimeFunction derives Pwrite at the moment, but it might be nice to also write the other unwind info eventually. I wasn't sure if that's necessary though and what the best pattern would be, especially for the UnwindCodes iterator.
If you or someone else wants to start on the pwrite implementation for a PE binary I would be super excited, or at least getting the ball rolling. https://github.com/m4b/faerie/issues/46 is be major reason for this :)
Imho, it's the last major thing I'd like to see done before we take this library 1.0 :D

m4b:
After this PR lands i'm going to release 0.22, got some great stuff in this release!

philipc:
@m4b This is already rebased after #134. I think all your review requests have been addressed?

jan-auer:
@m4b, It is rebased 👍 Tests are still missing, though. I hope to finish that today.

> `gread_inout_with` allows you to pass a mutable buffer of type 

Unfortunately, that reads the entire thing into a secondary buffer, which is not really in the spirit of memory mapped binary search :/ I was probably more looking for something along the lines of what we did [in the PDB crate](https://github.com/willglynn/pdb/blob/e365a6d71033c6d818fd2e5e107b562c24199242/src/omap.rs#L113-L151) (and [here](https://github.com/willglynn/pdb/blob/e365a6d71033c6d818fd2e5e107b562c24199242/src/omap.rs#L23-L42) the corresponding struct, which could be derived). As I've seen this pop a couple of times now, I might try to make some time for a PR to scroll.

> If you or someone else wants to start on the pwrite implementation for a PE binary

Phew 😅

jan-auer:
@m4b do you have an idea when you'll be able to run a release?

m4b:
Sorry got distracted I can try for this evening otherwise for sure this weekend!

m4b:
Ok 0.0.22 is published :)

src/pe/exception.rs
```diff
@@ -0,0 +1,826 @@
...
+//!     instruction is processed. After this, step 1 is repeated.
+//!
+//!  2. Case b) If the RIP lies within the prologue, then control has not entered the function,
+//!     there can be no exception handler associated with this exception for this function, and the
+//!     effects of the prolog must be undone to compute the context of the caller function. The RIP
+//!     is within the prolog if the distance from the function start to the RIP is less than or
+//!     equal to the prolog size encoded in the unwind info. The effects of the prolog are unwound
+//!     by scanning forward through the unwind codes array for the first entry with an offset less
+//!     than or equal to the offset of the RIP from the function start, then undoing the effect of
+//!     all remaining items in the unwind code array. Step 1 is then repeated.
```
bjorn3: *step 2?

src/pe/exception.rs
```diff
@@ -0,0 +1,826 @@
...
+    }
+}
+
+impl FusedIterator for RuntimeFunctionIterator<'_> {}
+impl ExactSizeIterator for RuntimeFunctionIterator<'_> {}
+
+/// An x64 register used during unwinding.
+///
+///  - `0` - `15`: General purpose registers
+///  - `17` - `32`: XMM registers
```
bjorn3: could you please add 16 (rip) to the list?

src/pe/exception.rs
```diff
@@ -0,0 +1,826 @@
...
+            24 => "$xmm7",
+            25 => "$xmm8",
+            26 => "$xmm9",
+            27 => "$xmm10",
+            28 => "$xmm11",
+            29 => "$xmm12",
+            30 => "$xmm13",
+            31 => "$xmm14",
+            32 => "$xmm15",
+            _ => "",
```
bjorn3: could you please return something like "<unknown>" here?

src/pe/exception.rs
philipc: No, it's step 1 from https://docs.microsoft.com/en-us/cpp/build/exception-handling-x64?view=vs-2017#unwind-procedure (the 1,2,3 here are actually case a,b,c from that doc).

src/pe/exception.rs
jan-auer: Can be added, but please note that for unwinding `$rip` can never occur as it cannot be represented.

PR #142, opened 2019-04-01
Fix unneeded unit return type

lzutao:
cc #137


lzutao:
Closed in favor of unique PR.

PR #148, opened 2019-04-03
Cleanup implementation of elf::gnu_hash::hash

### LINKED ISSUE #147: Is elf::gnu_hash::hash signature correct?
I personally think that `symbol` should be `&[u8]`. 
Look [at the code in glibc](https://code.woboq.org/userspace/glibc/elf/dl-lookup.c.html#552), it's just a C-string with no UTF-8 encoding requirements.

https://github.com/m4b/goblin/blob/e6f784eedd8499b44c05cde856fb38d14b5b0c84/src/elf/gnu_hash.rs#L26-L34

Also the implementation is not correct, 33 not 32.

I'm working on this issue right now. Just want to know what you guys think about this.

Btw, I don't think the hash function is in correct location. 
It shouldn't be in the macro because it's just an associated function
and it doesn't depend on `$size`.

lzutao:
Fix #147

### DIFF
--- a/src/elf/mod.rs
+++ b/src/elf/mod.rs
@@ -38,7 +38,7 @@
 //! `endian_fd` features if you disable `default`.
 
 #[macro_use]
-mod gnu_hash;
+pub(crate) mod gnu_hash;
 
 // These are shareable values for the 32/64 bit implementations.
 //
--- a/src/elf/gnu_hash.rs
+++ b/src/elf/gnu_hash.rs
@@ -12,7 +12,32 @@
 //!   3. maskwords
 //!   4. shift2
 //!
-//! See: https://blogs.oracle.com/ali/entry/gnu_hash_elf_sections
+//! See: https://blogs.oracle.com/solaris/gnu-hash-elf-sections-v2
+
+/// GNU hash function: takes a string and returns the u32 hash of that string
+pub fn hash(symbol: &str) -> u32 {
+    const HASH_SEED: u32 = 5381;
+    let mut hash = HASH_SEED;
+    for b in symbol.as_bytes() {
+        hash = hash
+            .wrapping_mul(33)
+            .wrapping_add(u32::from(*b));
+    }
+    hash
+}
+
+#[cfg(test)]
+mod tests {
+    use super::hash;
+    #[test]
+    fn test_hash() {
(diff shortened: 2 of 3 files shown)


src/elf/mod.rs
```diff
@@ -38,7 +38,7 @@
 //! `endian_fd` features if you disable `default`.
 
 #[macro_use]
-mod gnu_hash;
+pub mod gnu_hash;
```
lzutao: `pub(crate)` if needed.

philipc:
My concern with this one is that while symbols are technically `&[u8]`, in goblin I think we mostly treat them as `&str`, so this change would be inconsistent with the rest of goblin.

m4b:
> My concern with this one is that while symbols are technically `&[u8]`, in goblin I think we mostly treat them as `&str`, so this change would be inconsistent with the rest of goblin.

Yea I'd like to second this; I'm not saying 100% we shouldn't do this for, but I don't want to merge this before thinking about this more deeply. E.g., should the `StrTable` hand out `&[u8]` slices now? Honestly not sure that's really a win, and not sure what benefit it gets us really. At the end of the day majority of downstream users will end up converting it to `&str` anyway. OR, if they don't want/can't do that, they always have the option to access the raw bytes directly (since they gave us bytes to parse) and just use them for whatever purpose they need.

m4b:
I guess this would help if we had a direct usecase where interpreting the symbol as `&str` was actively harmful/troublesome/unhelpful.

lzutao:
[The ASM generated](https://godbolt.org/z/M18xSK) if we replace `&str` with `&[u8` is the same. So there
are performance benefices worth for this breaking change. I'll revert the 
change to use `&str` instead.

philipc:
So the changes in this PR now are:
- eliminate some duplicate code
- use `hash * 33` instead of `hash * 32 + hash` (Note that this code was already correct, but this change is fine since the compiler can do that optimisation itself.)
- add tests

Is that correct?

> The ASM generated if we replace &str with &[u8 is the same.

This is expected, because the first thing the function does with the symbol is convert it `&[u8]` (which is free).

lzutao:
@philipc You are correct. Now the PR is just a clean up.

PR #150, opened 2019-04-04
Running `cargo fmt` on whole project

lzutao:
Only run `cargo fmt` on project. No other changes is made.


philipc:
This is going to be up to @m4b to decide if the formatting is ok. If there's only some parts that aren't acceptable, then we can consider adding some `#[rustfmt::skip]` annotations before running `cargo fmt`.

m4b:
Sorry I don’t want to do this, at least not right now. We can ask for new patches to be formatted, but formatting the whole repo is very destructive to git history and doesn’t really get us a whole lot at the moment.

We can revisit this when 1.0 gets released but for now since this is a mechanical change we can run at any time and it’s basically effortless.

Don’t worry we’ll do this, but now is not the time (there’s no rush), so I’m going to close this.

m4b:
Clarification: we’ll run this _before_ 1.0 is released :)

PR #151, opened 2019-04-08
Refactor `elf::gnu_hash::GnuHash`

### LINKED ISSUE #149: casting from `*const u32` to a more-strictly-aligned pointer (`*const u64`)
This is the last error from clippy:
```rust
error: casting from `*const u32` to a more-strictly-aligned pointer (`*const u64`)
   --> src/elf/gnu_hash.rs:60:38
    |
60  |                 let bloomwords_ptr = hashtab.add(4) as *const $size;
    |                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    | 
   ::: src/lib.rs:365:9
    |
365 |         elf_gnu_hash_impl!(u64);
    |         ------------------------ in this macro invocation
    |
    = note: #[deny(clippy::cast_ptr_alignment)] on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#cast_ptr_alignment

error: aborting due to previous error
```

I don't know enough about implementation details of GNU Hash to know that this cast is safe or not. So I leave it to you (@m4b , @philipc ).

There are two things we could do:
- If this is safe to cast, we ignore the error with clippy attribute.
- If this is unsafe, how do we fix or recover from this cast?

lzutao:
Closes #149 .


CHANGELOG.md
```diff
@@ -15,6 +15,11 @@ I'm sorry, I will try my best to ease breaking changes.  We're almost to 1.0, do
   become a keyword in Rust 2018 edition.
 - BREAKING: Rename `mach::exports::SymbolKind::to_str(kind: SymbolKind)` -> `to_str(&self)`.
 - BREAKING: Rename `strtab::Strtab::to_vec(self)` -> `to_vec(&self).`
+- BREAKING: Changes in `elf::gnu_hash::GnuHash`:
+  + `new(hashtab: *const u32, total_dynsyms: usize, symtab: &'process [sym::Sym])`
+    to `new(hashtab: *const u32, total_dynsyms: usize)`.
+  + `find(&self, name: &str, hash: u32, strtab: &Strtab) -> Option<&sym::Sym>`
+    to `find(&self, symbol: &str, hash: u32, dynsym: &'a [sym::Sym], strtab: &Strtab)`.
```
lzutao: breaking change here

src/elf/gnu_hash.rs
```diff
@@ -47,95 +53,130 @@ macro_rules! elf_gnu_hash_impl {
...
-                let maskwords = *hashtab.add(2) as usize; // how many words our bloom filter mask has
+                let symindex = *hashtab.add(1);
+                let maskwords = *hashtab.add(2); // how many words our bloom filter mask has
+                assert!((maskwords & (maskwords - 1)) == 0);
                 let shift2 = *hashtab.add(3);
-                let bloomwords_ptr = hashtab.add(4) as *const $size;
-                let buckets_ptr = bloomwords_ptr.add(maskwords) as *const u32;
+                #[deny(clippy::cast_ptr_alignment)]
+                let bloom_filter_ptr = hashtab.add(4) as *const $size;
+                let buckets_ptr = bloom_filter_ptr.add((RATIO * maskwords) as usize) as *const u32;
```
lzutao: Same as https://github.com/bminor/glibc/blob/94e358f6d490650c714edb1ffc3a52f56ffe086e/elf/dl-lookup.c#L953 .

src/elf/gnu_hash.rs
```diff
@@ -47,95 +53,130 @@ macro_rules! elf_gnu_hash_impl {
...
+        impl<'a> GnuHash<'a> {
+            pub unsafe fn new(hashtab: *const u32, total_dynsyms: usize) -> GnuHash<'a> {
+                const WORD_SIZE: u32 = 32;
+                const RATIO: u32 = ELFCLASS_BITS / WORD_SIZE;
                 let nbuckets = *hashtab;
-                let symindex = *hashtab.add(1) as usize;
-                let maskwords = *hashtab.add(2) as usize; // how many words our bloom filter mask has
+                let symindex = *hashtab.add(1);
+                let maskwords = *hashtab.add(2); // how many words our bloom filter mask has
+                assert!((maskwords & (maskwords - 1)) == 0);
```
lzutao: Is `assert` accepted here?

src/elf/gnu_hash.rs
philipc: How did you test this?

lzutao:
I would like to write a test for this. But looking at glibc source tree, there aren't any particular
tests for GNU Hash or Hash table itself. I might be try to write a examples to test this
new PR.

CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ Before 1.0, this project does not adhere to [Semantic Versioning](http://semver.
 
 I'm sorry, I will try my best to ease breaking changes.  We're almost to 1.0, don't worry!
 
+## [Unreleased]
+### Changed
+- BREAKING: Changes in `elf::gnu_hash::GnuHash`:
+  + `new(hashtab: *const u32, total_dynsyms: usize, symtab: &'process [sym::Sym])`
+    to `new(hashtab: *const u32, total_dynsyms: usize)`.
+  + `find(&self, name: &str, hash: u32, strtab: &Strtab) -> Option<&sym::Sym>`
+    to `find(&self, symbol: &str, hash: u32, dynsym: &'a [sym::Sym], strtab: &Strtab)`.
```
lzutao: breaking change here

src/elf/gnu_hash.rs
```diff
@@ -1,143 +1,206 @@
...
-                let maskwords = *hashtab.add(2) as usize; // how many words our bloom filter mask has
+                let symindex = *hashtab.add(1);
+                let maskwords = *hashtab.add(2);
                 let shift2 = *hashtab.add(3);
-                let bloomwords_ptr = hashtab.add(4) as *const $size;
-                let buckets_ptr = bloomwords_ptr.add(maskwords) as *const u32;
+                assert!(
+                    (maskwords & (maskwords - 1)) == 0,
+                    "maskwords must be a power of two"
+                );
```
lzutao: Is assert accepted here?

lzutao:
@philipc I wrote a program that check for a symbol from a share object files
(and it ran Ok for now).
Which way do you prefer to write a test? Upload binary share object to 
`assets` folder and write test case which load that file?

philipc:
Have you fixed any bugs here then? It's really hard to see exactly what you have changed. In my opinion there's too many unnecessary changes.

It looks like @m4b has been including the binary for tests as inline code (e.g. https://github.com/m4b/goblin/blob/master/tests/macho.rs#L5).

lzutao:
@philipc Sorry for non relevant commits, I split them to another PR.

The old implementation cannot correctly check if symbol is exists or not 
in my test program.

m4b:
Please let me review before merging, busy right now might be a few days

src/elf/gnu_hash.rs
lzutao: I think we should rename this method to `from_raw` or something like that
and return a Result instead of panic in user code.

philipc:
>  Sorry for non relevant commits, I split them to another PR.

Splitting them to another PR is good, but that wasn't what prompted my comment. Even this single commit is changing too much for a bugfix. You've practically rewritten the file, and that means it is too hard for me to figure out what the significant changes are. Anyway, I'm going to let @m4b review this one, and he may feel differently about it.

src/elf/gnu_hash.rs
```diff
@@ -1,143 +1,206 @@
...
+            // Either 32 or 64-bit depending on the class of object
+            bloom_filter: &'a [$size], // => bloom_filter[maskwords]
+            /// GNU hash table bucket array; indexes start at 0. This array holds symbol
+            /// table indexes and contains the index of hashes in `chains`
+            buckets: &'a [u32], // => buckets[nbuckets]
+            /// Hash values; indexes start at 0. This array holds symbol table indexes.
+            chains: &'a [u32], // => chains[total_dynsyms - symindex]
+        }
+
+        impl<'a> core::fmt::Debug for GnuHash<'a> {
```
m4b: thank you for this!

src/elf/gnu_hash.rs
```diff
@@ -1,143 +1,206 @@
...
+        use crate::strtab::Strtab;
+        use core::mem;
+        use core::slice;
+
+        const BYTE_SIZE: u32 = 8;
+        /// Size of a bits mask in bloom filter
+        const ELFCLASS_BITS: u32 = mem::size_of::<$size>() as u32 * BYTE_SIZE;
 
-        pub struct GnuHash<'process> {
+        pub struct GnuHash<'a> {
```
m4b: thank you for documenting all these fields! Do note this is another good example though that makes this hard to read, almost every field changed in this struct...

lzutao:
I wonder why ` goblin::elf::sym::Sym` only available with `alloc` features.

src/elf/gnu_hash.rs
```diff
@@ -1,143 +1,206 @@
...
-                        -> Option<&sym::Sym> {
-                if self.filter(hash) {
-                    None
+            /// Given a symbol, a hash of that symbol, a dynamic string table and
+            /// a `dynstrtab` to cross-reference names, maybe returns a Sym.
+            pub fn find(
+                &self,
+                symbol: &str,
+                hash: u32,
+                dynsym: &'a [sym::Sym],
```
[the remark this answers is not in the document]
lzutao: I just move it from constructor to this method. I read gnu hash's implementation
in glibc project and found that we don't need to store that param in GnuHash struct,
instead we could pass that in `find` method.

src/elf/gnu_hash.rs
```diff
@@ -1,143 +1,206 @@
...
-                let filter = self.bloomwords[bloom_idx as usize] as usize; // FIXME: verify this is safe ;)
-                filter & (bitmask as usize) != (bitmask as usize) // if true, def _don't have_
+            /// Check if symbol maybe is in the hash table, or definitely not in it.
+            #[inline]
+            fn check_maybe_match(&self, hash: u32) -> bool {
+                #[inline(always)]
+                fn get_bitmask(hash: u32, hash2: u32) -> $size {
+                    // `hash & (ELFCLASS_BITS - 1)` is equivalent to `hash % ELFCLASS_BITS`
+                    // because `ELFCLASS_BITS` must be a power of 2.
+                    const MASK: u32 = ELFCLASS_BITS - 1;
```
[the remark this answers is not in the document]
lzutao: I translated some C codes from glibc implementation of GNU hash
and it feels natural to me to do this. If you don't like it, I will change
it to use modulus operation instead.

src/elf/gnu_hash.rs
lzutao: I added it in process of debugging this PR.

lzutao:
I added some tests for this PR. Hopefully those would prove this PR isn't doing wrong thing.

src/elf/gnu_hash.rs
lzutao: Sorry about that :)

lzutao:
Hm, networking failure.

lzutao:
Rebased on master.

lzutao:
Hi @m4b . Have you found time to review this PR? You could ask me
about anything changed here?

m4b:
Sorry I did not see you added tests, which was the major thing I wanted, will try a review this weekend.

m4b:
So I tried testing this on dryad, which is the only project I immediately know uses the gnu hash and depends on it being correct, but unfortunately it appears that something may have changed in the rust compiler, since it now segfaults when processing the kernel args, so i can't even test it :(

```
53	            // two null pointers mark end of envp
54	            // and beginning of the auxillary vectors
55	            while !(*p).is_null() {
   0x00007ffff7fa8b62 <+178>:	mov    0xa0(%rsp),%rax
   0x00007ffff7fa8b6a <+186>:	mov    (%rax),%rdi
   0x00007ffff7fa8b6d <+189>:	callq  *0x5389d(%rip)        # 0x7ffff7ffc410
=> 0x00007ffff7fa8b73 <+195>:	mov    %al,0x57(%rsp)
   0x00007ffff7fa8b77 <+199>:	jmp    0x7ffff7fa8b92 <dryad::kernel_block::KernelBlock::new+226>
   0x00007ffff7fa8b96 <+230>:	xor    $0xff,%al
   0x00007ffff7fa8b98 <+232>:	test   $0x1,%al
   0x00007ffff7fa8b9a <+234>:	jne    0x7ffff7fa8b9e <dryad::kernel_block::KernelBlock::new+238>
   0x00007ffff7fa8b9c <+236>:	jmp    0x7ffff7fa8b79 <dryad::kernel_block::KernelBlock::new+201>
   0x00007ffff7fa8beb <+315>:	jmpq   0x7ffff7fa8b62 <dryad::kernel_block::KernelBlock::new+178>
   0x00007ffff7fa8bf0 <+320>:	mov    0x48(%rsp),%rax

56	                p = p.offset(1);
```
I think that callq to an unknown address (which is where the segfault occurs) is supposed to be the pointer `is_null` call; not sure why this would have changed, which is sad. anyway.

src/elf/gnu_hash.rs
```diff
@@ -1,143 +1,194 @@
...
-            fn lookup(&self,
-                      symbol: &str,
-                      hash: u32,
-                      strtab: &Strtab)
-                      -> Option<&sym::Sym> {
-                let mut idx = self.buckets[(hash % self.nbuckets) as usize] as usize;
-                // println!("lookup idx = buckets[hash % nbuckets] = {}", idx);
-                if idx == 0 {
+            /// Locate the hash chain, and corresponding hash value element.
+            fn lookup<Sym: IntoIterator<Item = sym::Sym>>(
```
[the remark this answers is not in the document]
m4b: Actually, it's probably faster now since it's an iterator zipped with the chain, instead of indexing into the symtable :)

src/elf/sym.rs
```diff
@@ -320,114 +320,112 @@ pub mod sym64 {
...
-        pub st_other:    u8,
-        pub st_shndx:    usize,
-        pub st_value:    u64,
-        pub st_size:     u64,
-    }
+use scroll::ctx;
+use scroll::ctx::SizeWith;
+use core::fmt;
+use crate::container::{Ctx, Container};
+#[cfg(feature = "alloc")]
```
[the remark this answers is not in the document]
lzutao: I don't like it too. But since `gnu_hash` returns a `elf::sym::Sym`. Also
there are no cfgs around gnu_hash, so `elf::sym::Sym` shouldn't have that too.

tests/elf.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    match section {
+        Some(sym) => Ok(sym),
+        None => Err(String::from("cannot find symbol")),
+    }
+}
+
+#[test]
+fn test_parse_gnu_hash_section_64bit() {
+    let file = File::open("./tests/bins/elf/gnu_hash/hello.so").unwrap();
```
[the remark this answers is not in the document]
lzutao: Thanks for suggestion! I will change this.

src/elf/gnu_hash.rs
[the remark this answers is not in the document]
lzutao: I don't remember exactly. But probably some issues with lifetimes
and ownership happened. So I turned it into interator.

src/elf/gnu_hash.rs
```diff
@@ -1,143 +1,194 @@
...
-                        // println!("lookup match for {} at: 0x{:x}", symbol, symbol_.st_value);
-                        return Some(symbol_);
+                // Walk the chain until the symbol is found or the chain is exhausted.
+                let chain_idx = bucket - self.symindex;
+                let hash = hash & MASK_LOWEST_BIT;
+                for (hash2, symb) in self
+                    .chains
+                    .iter()
+                    .skip(chain_idx as usize)
+                    .zip(dynsyms.into_iter().skip(bucket as usize))
```
philipc: I think this iterator skip will be very inefficient compared with indexing like it used to. It defeats the purpose of using the hash in the first place.

m4b: @lzutao What do you think?

Effectively, we're talking about this code:
```rust
 101   │                 loop {
 102   │                     let symbol_ = &self.symtab[idx];
 103   │                     let h2 = self.hashvalues[hash_idx];
 104   │                     idx += 1;
 105   │                     hash_idx += 1;
 106   │                     let name = &strtab[symbol_.st_name as usize];
 107   │                     // println!("{}: h2 0x{:x} resolves to: {}", i, h2, name);
 108   │                     if hash == (h2 & !1) && name == symbol {
 109   │                         // println!("lookup match for {} at: 0x{:x}", symbol, symbol_.st_value);
 110   │                         return Some(symbol_);
 111   │                     }
 112   │                     if h2 & 1 == 1 {
 113   │                         break;
 114   │                     } // end of chain
 115   │                 }
```

@philipc why do you think it will be less efficient? I'm not super familiar with skip, but I just assumed it would begin the indexing operations at the given idx, e.g., it expands to something like:

```
for i in skip_value..end
```
Perhaps this is incorrect?

m4b: Fwiw, I used to benchmark the symbol resolution in dryad and it was getting close (but still somewhat far away) from the gnu linker. Didn't spend too much time, but `LD_DEBUG=statistics` is quite nice ballpark figures to test against. Unfortunately dryad is broken and i haven't had time to debug why the function call is now bad after there's been no code changes (except the rustc compiler) :(

philipc: `skip(bucket as usize)` uses `Iterator::nth`, and the default implementation of `Iterator::nth` runs `Iterator::next` in a loop, which is far less efficient than indexing (as in `let symbol_ = &self.symtab[idx];`). We could probably fix this by implementing `Iterator::nth` for `elf::SymIterator`, but personally I don't think this function should accept an iterator, to avoid these kinds of footguns.

m4b: K yes agreed, if that's how it is, skipping to index via iterating over everything (e.g., like accesing Nth element in a linked list) sucks; @lzutao sorry, needs to be fixed :( Did not know about this behavior of skip...

m4b:
@lzutao I would still like to see this epic PR land :) I think @philipc concern was last remaining issue.

There do appear to be some merge conflicts.  But if you have some spare time, let me know if you intend to work on this more or whether you ran out of steam (also this is totally ok, just want to get a feel for what state is here)

:) and of course, thanks for your work

lzutao:
Sorry for a long pending. I took a look at it again and it seems to me that the
API of GnuHash is not really safe or good.
I think we could instead construct GnuHash from Elf struct instead
of using public unsafe method.

m4b:
Ok, @lzutao do you recommend closing this or something else? There was a lot of good work done here, and hesitant to throw it away...

src/elf/gnu_hash.rs
```diff
@@ -1,143 +1,210 @@
...
-            hashvalues: &'process [u32],
-            symtab: &'process [sym::Sym],
+            /// 2 bit bloom filter on `chains`
+            // Either 32 or 64-bit depending on the class of object
+            bloom_filter: &'a [$IntTy],
+            /// GNU hash table bucket array; indexes start at 0. This array holds symbol
+            /// table indexes and contains the index of hashes in `chains`
+            buckets: &'a [u32],
+            /// Hash values; indexes start at 0. This array holds symbol table indexes.
+            chains: &'a [u32], // => chains[total_dynsyms - symindex]
```
lzutao: I should find a way to encode the dynsyms len.
Or should I save dynsyms in this struct?

src/elf/gnu_hash.rs
```diff
@@ -1,143 +1,210 @@
...
-                    let name = &strtab[symbol_.st_name as usize];
-                    // println!("{}: h2 0x{:x} resolves to: {}", i, h2, name);
-                    if hash == (h2 & !1) && name == symbol {
-                        // println!("lookup match for {} at: 0x{:x}", symbol, symbol_.st_value);
-                        return Some(symbol_);
+                // Walk the chain until the symbol is found or the chain is exhausted.
+                let chain_idx = bucket - self.symindex;
+                let hash = hash & MASK_LOWEST_BIT;
+                let chains = &self.chains[(chain_idx as usize)..];
+                let dynsyms = &dynsyms[(bucket as usize)..];
```
philipc: If you want to better tolerate corrupt data instead of panicking, you could use `slice::get` instead (for both `chains` and `dynsyms`). Maybe this is what you wanted the dynsyms length for, but you don't need that separately because slices already have a length.

src/elf/gnu_hash.rs
```diff
@@ -1,143 +1,194 @@
...
+                let bloom_filter_ptr = hashtab.add(4) as *const $size;
+                let bloom_filter = slice::from_raw_parts(bloom_filter_ptr, maskwords as usize);
+                let buckets_ptr = bloom_filter_ptr.add(maskwords as usize) as *const u32;
                 let buckets = slice::from_raw_parts(buckets_ptr, nbuckets as usize);
-                let hashvalues_ptr = buckets_ptr.add(nbuckets as usize);
-                let hashvalues = slice::from_raw_parts(hashvalues_ptr, total_dynsyms - symindex);
-                let bloomwords = slice::from_raw_parts(bloomwords_ptr, maskwords);
-                GnuHash {
+                let chains_ptr = buckets_ptr.add(nbuckets as usize);
+                let chains = slice::from_raw_parts(chains_ptr, total_dynsyms - symindex as usize);
```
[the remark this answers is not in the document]
lzutao: I added some simple check.

src/elf/gnu_hash.rs
philipc: So you moved `dynsyms` into this struct. Why? I don't really care if it is in this struct or not (@m4b might though), but I would like to understand what the problem is that you want to solve by doing this.

src/elf/gnu_hash.rs
```diff
@@ -1,143 +1,210 @@
...
+            pub fn find<'b>(&self, symbol: &str, dynsyms: &'b [Sym], dynstrtab: &Strtab) -> Option<&'b Sym> {
+                let hash = self::hash(symbol);
+                unsafe { self.find_with_hash(symbol, hash, dynsyms, dynstrtab) }
+            }
+
+            /// # Safety
+            ///
+            /// This function is unsafe because it will not check if the passed `hash` is really
+            /// the hash of `symbol`
+            pub unsafe fn find_with_hash<'b>(
```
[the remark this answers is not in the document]
philipc: You removed the `unsafe` in the signature, but you still need to ensure the body contains all the necessary safety checks. You need to check
- the [safety requirements](https://doc.rust-lang.org/1.30.0/std/ptr/fn.read.html#safety) of `ptr::read`
- the [safety requirements](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#safety) of `slice::from_raw_parts`

In particular, I think it is missing alignment checks, and zero length checks.

src/elf/gnu_hash.rs
```diff
@@ -1,143 +1,210 @@
...
+                    .field("bucket", &self.buckets.as_ptr())
+                    .field("chains", &self.chains.as_ptr())
+                    .finish()
+            }
+        }
+
+        impl<'a> GnuHash<'a> {
+            /// Initialize a GnuHash from a pointer to `.hash` (or `.gnu.hash`) section
+            /// and total number of dynamic symbols.
+            pub unsafe fn from_raw_table(hashtab: &'a [u8], total_dynsyms: usize) -> Result<Self, &'static str> {
```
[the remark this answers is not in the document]
philipc: You removed the `unsafe` in the signature, but you still need to ensure the body contains all the necessary safety checks. You need to check
- the [safety requirements](https://doc.rust-lang.org/1.30.0/std/ptr/fn.read.html#safety) of `ptr::read`
- the [safety requirements](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#safety) of `slice::from_raw_parts`

In particular, I think it is missing alignment checks, and zero length checks.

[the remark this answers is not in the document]
lzutao: I added the check that total length of slices constructed from hashtab must be
equal to hashtab.len().
About the alignment check, I don't think I could do any check because I just read it right
through hashtab. Perhaps I change the `pointer::read` to `pointer::read_unaligned` instead.

src/elf/gnu_hash.rs
lzutao: The `dynsyms` was in this struct before this PR. I moved it out because I want
to use `impl IntoIterator<Sym>` which you pointed out it may be slow in some
situations when combined with `Iterator::skip` method.
I move it in again because now we instead of passing `total_dynsyms` (which is `dynsyms.len()`)
to construct `chains` array. Just use `dynsyms` instead.

philipc: I see, thanks!

src/elf/gnu_hash.rs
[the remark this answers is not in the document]
philipc: It will be possible to do alignment checks (it's just a pointer, you can check the alignment of it), but let's leave this function as `unsafe` for now.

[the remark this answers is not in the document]
lzutao: I added the alignment check in 34232bb , but the test will failed
because the passed hashtab is not really aligned.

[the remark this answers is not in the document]
philipc: https://users.rust-lang.org/t/can-i-conveniently-compile-bytes-into-a-rust-program-with-a-specific-alignment/24049/6 talks about how to make the included bytes aligned.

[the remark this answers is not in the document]
lzutao: Thanks. But then does caller have to do that trick to call this function?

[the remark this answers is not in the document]
philipc: Probably not. I think the normal case is that the file will be memory mapped or at least loaded at a page alignment, and the hash table is normally designed to be aligned given that. So for that case the caller won't need to do it. So doing this trick in the test is just matching the normal use case.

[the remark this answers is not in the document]
lzutao: That's seems fair. Added it in 73a37c9248dc45cd37305346938a6d4c3261e93d .

philipc:
I think this is ok now, unless @m4b wants to take a look.

m4b:
I’m currently afk but I would like to do a quick review. I’ll try to review by Sunday.  Please feel free to ping if I haven’t reviewed and OK’d after :)

lzutao:
Thanks. Please let me squash or rebase this PR before merge.

m4b:
Once you’re finished, could you please squash everything related to gnu hash into single commit, and change log separate. If it’s too annoying just all a single commit, please :)

PR #153, opened 2019-04-14
mach: add Nlist::is_weak

philipc:


src/mach/symbols.rs
```diff
@@ -83,6 +83,48 @@ pub const NLIST_TYPE_MASK: u8 = 0xe;
...
+/// and is visible only to modules within this shared library.
+pub const REFERENCE_FLAG_PRIVATE_UNDEFINED_NON_LAZY: u16 = 0x4;
+/// This symbol is defined in another module in this file, is a lazy (function) symbol,
+/// and is visible only to modules within this shared library.
+pub const REFERENCE_FLAG_PRIVATE_UNDEFINED_LAZY: u16 = 0x5;
+
+// Additional flags of n_desc field.
+
+/// Must be set for any defined symbol that is referenced by dynamic-loader APIs
+/// (such as dlsym``` and NSLookupSymbolInImage```) and not ordinary undefined symbol
```
willglynn: I'd like to improve this formatting:

Screenshot

Mostly the triple backticks, but also the mixing of both "strip tool" and `strip` in the same message.

src/mach/symbols.rs
```diff
@@ -83,6 +83,48 @@ pub const NLIST_TYPE_MASK: u8 = 0xe;
...
+/// references. The strip tool uses this bit to avoid removing symbols that must
+/// exist: If the symbol has this bit set, `strip` does not strip it.
+pub const REFERENCED_DYNAMICALLY: u16 = 0x10;
+/// Sometimes used by the dynamic linker at runtime in a fully linked image. Do not
+/// set this bit in a fully linked image.
+pub const N_DESC_DISCARDED: u16 = 0x20;
+/// When set in a relocatable object file (file type MH_OBJECT) on a defined symbol,
+/// indicates to the static linker to never dead-strip the symbol. (Note that the
+/// same bit (0x20) is used for two nonoverlapping purposes.)
```
willglynn: Someone looking just at the docs wouldn't know that 0x20 also means `N_DESC_DISCARDED`. Please mention this fact in both places, and consider having them both refer to the other.

philipc: I changed it to not be a doc comment instead. The way that rustdoc presents these constants is not useful, and I would only ever look at the source for this info, so it's not worth spending time on.

I should have mentioned that all this doc is copied from the Apple documentation.

m4b:
LGTM, thanks @philipc and @willglynn !

PR #154, opened 2019-04-16
Minor documentation improvements

lzutao:


src/elf/reloc.rs
```diff
@@ -1,48 +1,70 @@
 //! # Relocation computations
+//!
+//! The following notation is used to describe relocation computations.
```
philipc: Note that these are specific to x86_64.

src/elf/reloc.rs
```diff
@@ -1,48 +1,70 @@
...
+//!
+//!  * A: The addend used to compute the value of the relocatable field.
+//!  * B: The base address at which a shared object is loaded into memory
+//!       during execution. Generally, a shared object file is built with a
+//!       base virtual address of 0. However, the execution address of the
+//!       shared object is different. See Program Header.
```
philipc: Here and below: delete the 'See XXX' statements.

src/elf/compression_header.rs
```diff
@@ -30,7 +30,7 @@ pub const ELFCOMPRESS_HIPROC: u32 = 0x7fff_ffff;
 macro_rules! elf_compression_header_std_impl { ($size:ty) => {
 
     #[cfg(test)]
-    mod test {
+    mod tests {
```
philipc: What is the problem with the old name?

Cargo.toml
```diff
@@ -14,15 +14,8 @@ include = ["src/**/*", "Cargo.toml", "CHANGELOG.md", "LICENSE", "README.md", "et
...
-[dependencies.log]
-version = "0.4"
-default-features = false
-optional = true
-
-[dependencies.scroll]
-version = "0.9"
-default_features = false
+log = { version = "0.4", default-features = false, optional = true }
+scroll = { version = "0.9", default-features = false }
```
philipc: Why? I'm not convinced that this is objectively better.

examples/dyldinfo.rs
```diff
@@ -126,8 +122,7 @@ fn main () {
                     "-bind" => { bind = true }
                     "-lazy_bind" => { lazy_bind = true }
                     other => {
-                        println!("unknown flag: {}", other);
-                        println!("");
+                        println!("unknown flag: {}\n", other);
```
philipc: Which clippy warning is this?

src/elf/compression_header.rs
lzutao: It's just a convention from [the Book](https://doc.rust-lang.org/book/ch11-03-test-organization.html#the-tests-module-and-cfgtest).

Cargo.toml
lzutao: I don't think one is better than another but the new one is shorter.
I'd revert if you prefer.

examples/dyldinfo.rs
lzutao: FYI, it is [print_with_newline](https://rust-lang.github.io/rust-clippy/master/index.html#print_with_newline).

src/elf/reloc.rs
lzutao: Done

src/elf/reloc.rs
lzutao: Done

Cargo.toml
philipc: Yes, please drop that commit.

examples/dyldinfo.rs
philipc: Ah okay, please use `println!()` instead. It's clearer and I don't think the performance is a concern.

philipc: Also I think you meant [println_empty_string](https://rust-lang.github.io/rust-clippy/master/index.html#println_empty_string).

lzutao: If we revert this change, every time clippy run it will warn about this. 
Kind of annoyance.

Cargo.toml
lzutao: Done.

examples/dyldinfo.rs
lzutao: Ok, I forget I could just drop the empty string.

PR #155, opened 2019-04-23
Basic COFF support

philipc:
Reuses the existing parsing for the COFF header and sections, and adds COFF symbol parsing.

Work for future PRs:
- auxiliary symbol records
- COFF relocations
- write support

### DIFF
--- a/src/pe/header.rs
+++ b/src/pe/header.rs
@@ -30,8 +30,6 @@ impl DosHeader {
 #[repr(C)]
 #[derive(Debug, PartialEq, Copy, Clone, Default)]
 pub struct CoffHeader {
-    /// COFF Magic: PE\0\0, little endian
-    pub signature: u32,
     /// The machine type
     pub machine: u16,
     pub number_of_sections: u16,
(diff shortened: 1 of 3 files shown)


src/pe/header.rs
```diff
@@ -30,8 +31,6 @@ impl DosHeader {
 #[repr(C)]
 #[derive(Debug, PartialEq, Copy, Clone, Default)]
 pub struct CoffHeader {
-    /// COFF Magic: PE\0\0, little endian
-    pub signature: u32,
```
m4b: How come this changed? Or put another way, how does this work now, or how did it work before ?
Is the data layout of coff/PE such that it just works ?
Also this is technically breaking so we should add a breaking changes note, though I doubt anyone used it.

**edit** looking at this now looks like this is the second element of the header just before the coff header which means they’d have the same layout if it’s packed. Wow good fix/find :)

philipc:
> so just curious is coff used very much anymore ? I recall it being what the mingw windows environment and its compiler will emit, but can’t remember ?

As far as I can tell, COFF is always used for object files on Windows (even msvc). PE is an executable (image) format, not an object file format.

src/pe/header.rs
philipc: Yeah I'm just moving this field around. I had to do this because COFF object files don't have any of the PE headers, including this magic value.

PR #159, opened 2019-05-01
pe: add write support for COFF object files

philipc:
See https://github.com/philipc/object-write/blob/master/src/coff.rs for an example of its usage.

### DIFF
--- a/src/pe/header.rs
+++ b/src/pe/header.rs
@@ -2,7 +2,7 @@ use crate::error;
 use crate::pe::{optional_header, section_table, symbol};
 use crate::strtab;
 use log::debug;
-use scroll::Pread;
+use scroll::{Pread, Pwrite, IOread, IOwrite, SizeWith};
 
 /// DOS header present in all PE binaries
 #[repr(C)]
@@ -29,7 +29,7 @@ impl DosHeader {
 
 /// COFF Header
 #[repr(C)]
-#[derive(Debug, PartialEq, Copy, Clone, Default)]
+#[derive(Debug, PartialEq, Copy, Clone, Default, Pread, Pwrite, IOread, IOwrite, SizeWith)]
 pub struct CoffHeader {
     /// The machine type
     pub machine: u16,
@@ -41,6 +41,7 @@ pub struct CoffHeader {
     pub characteristics: u16,
 }
 
+/// The size of `COFF_MAGIC` + `CoffHeader`.
 pub const SIZEOF_COFF_HEADER: usize = 24;
 /// PE\0\0, little endian
 pub const COFF_MAGIC: u32 = 0x0000_4550;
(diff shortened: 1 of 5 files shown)


src/pe/header.rs
```diff
@@ -41,6 +41,7 @@ pub struct CoffHeader {
     pub characteristics: u16,
 }
 
+/// The size of `COFF_MAGIC` + `CoffHeader`.
 pub const SIZEOF_COFF_HEADER: usize = 24;
```
philipc: Not sure if we should change this value to exclude COFF_MAGIC. It would be a silent breaking change.

m4b: I'm ok breaking the world if we get windows object file writers :) this is why i wanted something like this in before 1.0, because there's unknown unknowns that might break a stable api, which would otherwise have been seen

src/pe/section_table.rs
```diff
@@ -94,6 +131,29 @@ impl SectionTable {
...
+    type Units = usize;
+    fn size_with(_ctx: &scroll::Endian) -> usize {
+        SIZEOF_SECTION_TABLE
+    }
+}
+
+impl ctx::IntoCtx<scroll::Endian> for SectionTable {
+    fn into_ctx(self, bytes: &mut [u8], ctx: scroll::Endian) {
+        let offset = &mut 0;
+        bytes.gwrite(&self.name[..], offset).unwrap();
```
m4b: Normally one implements the TryIntoCtx variant, and just unwraps that implementation here; why isn't that done?

philipc: Because I'm not familiar enough with scroll. I'll fix it.

Along those lines, why do many of the goblin structures have a `fn parse` instead of or in addition to implementing TryFromCtx?

src/pe/header.rs
philipc: Added another commit with this breaking change.

src/pe/section_table.rs
m4b: It was for usually api reasons; I wanted the various parseable "things" to have a `pub fn parse`-esque method which made no mention of `scroll` in the signature, so someone wouldn't be forced to import both goblin + scroll. Probably in many cases, e.g., `Elf::parse` it started off _before_ scroll got written :D (I wrote scroll because I was sick of hand writing stupid byte order field populators, essentially, and adding new parsers or structs was tedious and boring, and refactors or bug fixes were extremely annoying - e.g., multiple times there was some typo read the wrong width out into a variable, or etc.).

I called it `parse`, because I like `new` to be infallible construction, so it was just parse. And adding the `try_from_ctx` impl in elf and to a certain extent mach was an afterthought. If they got added (or not) is just no one has had time to go back and them, or felt the need (since most people use the `Object::parse` endpoint afaics.

And that's the whole story about that :laughing:

m4b:
E.g., this example was one possible api usage (using raw `pread`), and is still possible, but it requires one to import scroll + goblin, which is annoying imho. But it sure seems cool to me to `pread` an elf object :) https://github.com//m4b/goblin/blob/779d0ce835bb325327f39157ea8fdb8882daa171/examples/scroll.rs#L22

philipc:
I've verified this is good enough to create an object file that can be linked and execute. May need more in future for exception handling, but I don't know enough about that yet.

PR #164, opened 2019-05-20
Get build artifact path reliable with std::env::current_exe

lzutao:
We sort the `package` section based on https://github.com/rust-dev-tools/fmt-rfcs/blob/master/guide/cargo.md

@m4b Do we need to include `examples`, `tests` and `fuzz` directories
in goblin crate?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,15 +2,15 @@
 name = "goblin"
 version = "0.1.3"
 authors = ["m4b <m4b.github.io@gmail.com>", "seu <seu@panopticon.re>", "Will Glynn <will@willglynn.com>", "Philip Craig <philipjcraig@gmail.com>"]
-edition = "2018"
-readme = "README.md"
-keywords = ["binary", "elf", "mach", "pe", "archive"]
-repository = "https://github.com/m4b/goblin"
-license = "MIT"
-description = "An impish, cross-platform, ELF, Mach-o, and PE binary parsing and loading crate"
-documentation = "https://docs.rs/goblin"
 categories = ["parsing", "development-tools::debugging"]
+documentation = "https://docs.rs/goblin"
+edition = "2018"
 include = ["src/**/*", "Cargo.toml", "CHANGELOG.md", "LICENSE", "README.md", "etc/*", "examples/*", "tests/*", "fuzz/**/*"]
+keywords = ["binary", "elf", "mach", "pe", "archive"]
+license = "MIT"
+readme = "README.md"
+repository = "https://github.com/m4b/goblin"
+description = "An impish, cross-platform, ELF, Mach-o, and PE binary parsing and loading crate"
 
 [dependencies]
 plain = "0.2.3"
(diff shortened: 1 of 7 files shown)


m4b:
Hmmm probably not ? Included is only for building from source via a registry (like crates.io right ?)

lzutao:
Yes, I [think so](https://doc.rust-lang.org/nightly/cargo/reference/manifest.html#the-exclude-and-include-fields-optional).

build.rs
```diff
@@ -0,0 +1,14 @@
+use std::env;
```
m4b: In general I’m against build files unless they’re absolutely necessary; could you give some motivation for why we need this ? I really don’t want to add a build.rs

lzutao:
Sorry. I has caught an new build error with this build.rs (`cargo package`). I will try to resolve it
by having help from cargo team.

build.rs
lzutao: The `parse_self` in `tests/archive.rs` requires to know the build artifact
path. The old method is incorrect. The build artifact could be
`target/debug`, `target/release`, `target/<target-triple>/debug`.
I find out that we could rely on `OUT_DIR` variable to know
exactly where the build artifact path is. But the `OUT_DIR` variable is 
not set if [there is no `build.rs`][1] script.

We could create an empty build.rs and parsing `OUT_DIR` in `parse_self`
test if you prefer.

[1]: https://doc.rust-lang.org/nightly/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates

m4b:
Did the build issue get resolve from cargo team (you mentioned)?

lzutao:
> Did the build issue get resolve from cargo team (you mentioned)?

Yeah, it was resolved as of Rust 1.36.0.

lzutao:
Can someone review this?

lzutao:
Could you review this @philipc ?

philipc:
Is it normal to exclude tests? I know some distros like to run tests to check they have packaged it correctly.

lzutao:
I don't know about that. But I think Debian maintains their own git fork of crates
instead of using packages on crates.io.

philipc:
I downloaded the debian package and it looks to me like they obtained their source from crates.io. It is missing the assets directory, and they have a changelog entry saying that the archive test is flaky because the assets are missing. That also confirms that they do run the tests.

philipc:
Maybe @infinity0 can tell us if Debian expects us to keep tests for the crates.io package.

infinity0:
In Debian (and also Fedora and I think most other distros) we generally run the tests per-crate rather than per-workspace. Sometimes this fails because as you noted the tests depend on files in the parent workspace that don't get uploaded to crates.io as part of the crate. However I do believe the rust ecosystem "in general" is trying to discourage this practise to be able to support distros like us better, but I'm not sure what the progress is like on this or actually where I even heard it from in the first place. CC @ignatenkobrain and @cuviper the Fedora rust people for discussion.

PR #166, opened 2019-05-30
Fix minor clippy warnings

lzutao:


src/pe/section_table.rs
```diff
@@ -227,14 +227,14 @@ mod tests {
     #[test]
     fn set_name_offset() {
         let mut section = SectionTable::default();
-        for (offset, name) in vec![
-            (0, b"/0\0\0\0\0\0\0"),
+        for &(offset, name) in [
+            (0usize, b"/0\0\0\0\0\0\0"),
```
philipc: The `usize` shouldn't have been needed.

PR #168, opened 2019-06-04
Add constants for ARM64_32 (ILP32 ABI on 64-bit ARM)

jan-auer:
This is used by the new Apple Watch. I have tried to get this as close as possible to what LLVM does, but I'm not 100% sure about the relocations mapping.


philipc:
> I'm not 100% sure about the relocations mapping.

LLVM treats `Triple::aarch64` and `Triple::aarch64_32` the same when printing relocation names.

PR #170, opened 2019-06-06
Fix alloc build

philipc:
I don't know if there's a better way to do the number printing without `io::Write`.

### DIFF
--- a/src/pe/section_table.rs
+++ b/src/pe/section_table.rs
@@ -2,7 +2,6 @@ use crate::alloc::string::{String, ToString};
 use scroll::{ctx, Pread, Pwrite};
 use crate::error::{self, Error};
 use crate::pe::relocation;
-use std::io::Write;
 
 #[repr(C)]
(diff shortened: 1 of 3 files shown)


src/pe/section_table.rs
```diff
@@ -93,7 +92,25 @@ impl SectionTable {
         if idx <= 9_999_999 { // 10^7 - 1
             self.name = [0; 8];
             self.name[0] = b'/';
-            write!(&mut self.name[1..], "{}", idx).unwrap();
+            if idx == 0 {
```
m4b: i literally have no idea what this is doing :laughing:

philipc: It's doing `write!(&mut self.name[1..], "{}", idx)` without using `io::Write`. I could add some comments to make it clearer.

m4b: If you want to. Probably what you said right here is sufficient :) feel free to merge after

Issue #173, opened 2019-07-05
Parsing of static Windows library archives fails


raindev:
I've got some Windows static libraries (`.lib` archives) that goblin fails to parse. One file is a toy example consisting of two "hello world" object files[1], another one is BoringSSL crypto module[2]. Both are built on Windows 10 using Visual Studio toolchain[3]. The error is `type is too big (4) for 3`.

Debugging revealed that the problem is caused by invalid symbol lookup table size read here: https://github.com/m4b/goblin/blob/master/src/archive/mod.rs#L200
Because the size is too large, the loop goes on reading 4-byte values until reaching the end of the buffer, failing if buffer size is not divisible by 4.

Taking `lib.lib` as an example there're two symbol lookup tables (member name is `/`, space padded to the right). Both members are 151 bytes large (hence 3 bytes unparsable into u32 left). One of them (offset 0x44) has correct table size: `00 00 00 06` (big endian) is parsed correctly, the other (offset 0x118) however starts from of `02 00 00 00` (way out of range for string table size). It turned out Windows extends System V archive format adding a second symbol lookup table (they call it Linker Member) named `/` which has a different format[4]. E.g. the first 4 bytes is the number of archive members little-endian, not the size of string table big-endian.

Since a second file named `/` could only be present in Windows library archives, it should be easy to implement support for without breaking System V compatibility. I hope to follow up with an implementation.

This might be the same problem from #133 but I'm yet to confirm it.

1. https://www.dropbox.com/s/df7su6n9r26472i/lib.lib?dl=0
2. https://www.dropbox.com/s/9zy9sa7k5bmf9cp/crypto.lib?dl=0
3. Microsoft C/C++ Optimizing Compiler Version 19.21.27702.2 for x64, Incremental Linker Version 14.21.27702
4. https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#second-linker-member

m4b:
Wow, thanks for the great write up ! Very interesting stuff :)

If you’ve got an idea for an implementation and some time, go for it! Looking forward to it :)

Issue #133, opened 2019-03-25
archive: memory allocation of 16378232832 bytes failed


leo-lb:
Hello,

I don't really have time now to debug this but I'd still like to share the test file.
It's being parsed as an archive.

The application exits after printing `memory allocation of 16378232832 bytes failed`.

https://github.com/leo-lb/random-files/blob/master/jvm.lib

Thanks.

jjunqueira:
This is happening to me as well, it unfortunately causes an abort panic and crashes my program. I can't figure out a way to catch the error and keep processing. I wonder if its the same issue from here [https://github.com/m4b/goblin/issues/120](url) but it is happening to me on PE files.

m4b:
@leo-lb thanks for the file; so on current master i get:
```
type is too big (4) for 3
```
when i run:
```
cargo run --example=rdr ~/Downloads/jvm.lib
```

`nm` can parse it though without failing so we should do better.

@jjunqueira the issue you linked seems to be a deadlink, but you are also getting OOM errors?

jjunqueira:
@m4b Oops sorry https://github.com/m4b/goblin/issues/120

I am using 0.0.22 so maybe I should use master? 

The part in issue 120 that I think is applying here is "Memory allocator runs out of virtual memory and the process is aborted.".  I am gathering pe file information from large batches of binaries and when I hit this error my whole process is aborted. In my case I would prefer it return an error that I could handle instead of abort.

I'll try doing a test on master and report back!

m4b:
Yes we should never oom or panic, user error should always be returned. This is a bug for sure.

m4b:
Can you upload the offending PE or is it sensitive ? Would like to poke at it and see what’s wrong.

jjunqueira:
@m4b Sorry got sidetracked on a few things and haven't made it back to this yet. I'll ask to see if I can share the sample. I also haven't gotten around to trying it on the master branch yet. I am hoping to get to it in the morning.

jjunqueira:
@m4b @leo-lb So I feel kinda dumb now but I just saw this in the README "Goblin requires rustc 1.31.1." and of course my version is:

▶ rustc --version
rustc 1.34.1 (fc50f328b 2019-04-24)

Then I was thinking that a major pretty recent change was moving away from jemalloc as the default allocator and I am having an abort on memory allocation. So I switched my project to include jemalloc and BOOM problem gone.

[...]

PR #174, opened 2019-07-12
Fix parsing of Windows library archives

raindev:
Implement parsing of Windows-specific Second Linker Member in SysV archives. Also fix the parser going out of boundaries if the archive is aligned.

I have verified that the parser succeeds on the samples from both #173 and #133. I was thinking about automated tests but it's quite ugly bringing binary data into the repository for testing. I think using cc crate to build binaries from source might be a better approach. With a Windows Travis pipeline the CI can verify parsing of PE binaries as well.

### DIFF
--- a/src/archive/mod.rs
+++ b/src/archive/mod.rs
@@ -279,6 +279,37 @@ impl<'a> Index<'a> {
             strtab: strings,
         })
     }
+
+    // Parses Windows Second Linker Member:
+    // number of members (m):   4
+    // member offsets:          4 * m
+    // number of symbols (n):   4
+    // symbol member indexes:   2 * n
+    // followed by SysV-style string table
+    // https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#first-linker-member
+    pub fn parse_windows_linker_member(buffer: &'a [u8]) -> Result<Self> {
+        let offset = &mut 0;
+        let members = buffer.gread_with::<u32>(offset, scroll::LE)? as usize;
+        let mut member_offsets = Vec::with_capacity(members);
+        for _ in 0..members {
+            member_offsets.push(buffer.gread_with::<u32>(offset, scroll::LE)?);
+        }
+        let symbols = buffer.gread_with::<u32>(offset, scroll::LE)? as usize;
+        let mut symbol_indexes = Vec::with_capacity(symbols);
+        for _ in 0..symbols {
+            symbol_indexes.push(buffer.gread_with::<u16>(offset, scroll::LE)? as usize);
+        }
+        let mut symbol_offsets = Vec::with_capacity(symbols);
+        for i in symbol_indexes {
+            symbol_offsets.push(member_offsets[i - 1]);
+        }
+        let strtab = strtab::Strtab::parse(buffer, *offset, buffer.len() - *offset, 0x0)?;
+        Ok(Index {
+            size: symbols,
+            symbol_indexes: symbol_offsets,
+            strtab: strtab.to_vec()?,
+        })
+    }
 }
 
 /// Member names greater than 16 bytes are indirectly referenced using a `/<idx` schema,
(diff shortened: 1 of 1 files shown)


src/archive/mod.rs
```diff
@@ -366,7 +398,13 @@ impl<'a> Archive<'a> {
             let name = member.raw_name();
             if name == INDEX_NAME {
                 let data: &[u8] = buffer.pread_with(member.offset as usize, member.size())?;
-                index = Index::parse_sysv_index(data)?;
+                index = if sysv_symbol_index_seen {
+                    // Second symbol index is Microsoft's extension of SysV format
+                    Index::parse_windows_linker_member(data)?
```
raindev: Accordingly to Microsoft's documentation, the first symbol index entry (First Linker Member) is not used by linkers and is kept for compatibility reasons. Given this, it seems reasonable to simply replace the index parsed previously rather than providing access to both.

src/archive/mod.rs
```diff
@@ -366,7 +398,13 @@ impl<'a> Archive<'a> {
...
+                } else {
+                    sysv_symbol_index_seen = true;
+                    Index::parse_sysv_index(data)?
+                }
```
philipc:
```suggestion
                };
```

src/archive/mod.rs
```diff
@@ -366,7 +398,13 @@ impl<'a> Archive<'a> {
             let name = member.raw_name();
             if name == INDEX_NAME {
                 let data: &[u8] = buffer.pread_with(member.offset as usize, member.size())?;
-                index = Index::parse_sysv_index(data)?;
+                index = if sysv_symbol_index_seen {
```
philipc: Maybe we should return an error if this is the third time or more.

src/archive/mod.rs
philipc: Sounds good to me.

src/archive/mod.rs
```diff
@@ -279,6 +279,37 @@ impl<'a> Index<'a> {
...
+            member_offsets.push(buffer.gread_with::<u32>(offset, scroll::LE)?);
+        }
+        let symbols = buffer.gread_with::<u32>(offset, scroll::LE)? as usize;
+        let mut symbol_indexes = Vec::with_capacity(symbols);
+        for _ in 0..symbols {
+            symbol_indexes.push(buffer.gread_with::<u16>(offset, scroll::LE)? as usize);
+        }
+        let mut symbol_offsets = Vec::with_capacity(symbols);
+        for i in symbol_indexes {
+            symbol_offsets.push(member_offsets[i - 1]);
```
philipc: Could this be done in the previous loop? That is, we don't actually need to create `symbol_indexes`.

src/archive/mod.rs
philipc: Or ignore them.

src/archive/mod.rs
raindev: That's right, done.

src/archive/mod.rs
raindev: I'm not sure about this. We'll have to track then number of occurrences of index, including BSD index to have the behaviour consistent and allow only one index to be present (exept if it's Windows linker member). I think it would be better to tackle it systematically in a separate pull-request.

raindev: The current behaviour would be to threat all the subsequent SysV symbol indexes as Windows linker members, similar to running into multiple BSD symbol indexes.

raindev:
Thanks for the quick feedback Philip!

src/archive/mod.rs
philipc: I'm fine with doing this in another PR, but I don't like treating subsequent SysV symbol index as Windows linker members because that's the sort of sloppy parsing that caused the errors this PR is fixing. It's better to not support something, rather than try to parse it incorrectly.

raindev: I agree that the parsing should be too strict. It's just the question of how much complexity is it worth introducing for. I've drafted a separate PR #175 to figure the best way forward.

raindev:
@philipc it would be great if you can release a version which contains the changes to crates.io.

philipc:
@m4b Can you do a release please?

PR #175, opened 2019-07-12
Be more strict about multiple indexes found in an archive

raindev:
Fail fast rather than attempting to parse and overwriting already initialized symbol index. Came up in
https://github.com/m4b/goblin/pull/174#discussion_r302866012.

### DIFF
--- a/src/archive/mod.rs
+++ b/src/archive/mod.rs
@@ -370,6 +370,14 @@ pub struct Archive<'a> {
 
 impl<'a> Archive<'a> {
     pub fn parse(buffer: &'a [u8]) -> Result<Archive<'a>> {
+        #[derive(PartialEq)]
+        enum IndexType {
+            NoIndex,
+            SysV,
+            WindowsLinkerMember,
+            BSD,
+        };
+
         let mut magic = [0u8; SIZEOF_MAGIC];
         let offset = &mut 0usize;
         buffer.gread_inout(offset, &mut magic)?;
@@ -378,9 +386,7 @@ impl<'a> Archive<'a> {
         }
         let mut member_array = Vec::new();
         let mut index = Index::default();
-        let mut sysv_index_seen = false;
-        let mut windows_linker_member_seen = false;
-        let mut bsd_index_seen = false;
+        let mut index_type = IndexType::NoIndex;
         let mut sysv_name_index = NameIndex::default();
         while *offset + 1 < buffer.len() {
             // realign the cursor to a word boundary, if it's not on one already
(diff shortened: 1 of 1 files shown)


philipc:
This is basically fine, but you could consider using a single enum instead of multiple bools, and exposing that enum value could be useful too if users want to know which type of archive it is.

raindev:
Using an enum to represent the state machine is definitely a clearer approach.

As for exposing the enum to the caller, I see the following options:

- change the return type of `Archive::parse` to `(ArchiveType, Result<Archive<'a>>)`
- include `type: IndexType` into `Archive` struct
- change `Archive` itself to be a enum

The second option seems preferable since the first incurs a breaking change. The third seems compelling conceptually since it'll allow to expose different data present in different  archives in a type safe way while maintaining a common interface but it's quite a large change which includes many breaking changes to the interface of the library.

philipc:
I prefer the second option. It makes sense to me for this to be an `Archive` property, rather than an additional return value, and I think changing `Archive` to an enum would lead to too much duplication.

src/archive/mod.rs
```diff
@@ -343,6 +374,14 @@ pub struct Archive<'a> {
 
 impl<'a> Archive<'a> {
     pub fn parse(buffer: &'a [u8]) -> Result<Archive<'a>> {
+        #[derive(PartialEq)]
+        enum IndexType {
+            NoIndex,
```
philipc:
```suggestion
            None,
```

src/archive/mod.rs
```diff
@@ -343,6 +374,14 @@ pub struct Archive<'a> {
...
+            SysV,
+            WindowsLinkerMember,
```
philipc: Seems a bit verbose, would simply `Windows` suffice?

src/archive/mod.rs
```diff
@@ -365,10 +405,27 @@ impl<'a> Archive<'a> {
...
+                    index_type = IndexType::WindowsLinkerMember;
+                    // second symbol index is Microsoft's extension of SysV format
+                    Index::parse_windows_linker_member(data)?
+                } else {
+                    index_type = IndexType::SysV;
+                    Index::parse_sysv_index(data)?
+                }
 
             } else if member.bsd_name == Some(BSD_SYMDEF_NAME) || member.bsd_name == Some(BSD_SYMDEF_SORTED_NAME) {
+                if index_type == IndexType::SysV {
```
philipc:
```suggestion
                if index_type != IndexType::None {
```

src/archive/mod.rs
```diff
@@ -365,10 +405,27 @@ impl<'a> Archive<'a> {
...
                 let data: &[u8] = buffer.pread_with(member.offset as usize, member.size())?;
-                index = Index::parse_sysv_index(data)?;
+                index = if index_type == IndexType::SysV {
+                    index_type = IndexType::WindowsLinkerMember;
+                    // second symbol index is Microsoft's extension of SysV format
+                    Index::parse_windows_linker_member(data)?
+                } else {
+                    index_type = IndexType::SysV;
+                    Index::parse_sysv_index(data)?
+                }
```
philipc: I think this whole block would be clearer as:
```rust
match index_type {
    IndexType::None => { ... }
    IndexType::SysV => { ... }
    _ => return Err(...),
}
```

src/archive/mod.rs
raindev: Sounds good to me.

src/archive/mod.rs
raindev: I agree, it's clearer. Kept separate clauses for Windows/BSD to give more specific error messages.

src/archive/mod.rs
raindev: Good catch!

PR #177, opened 2019-08-07
pe.export: fix reexport dll and ordinal

### LINKED ISSUE #160: reexport lib name may be truncated
Reviewing the exports of `kernel32.dll`, it looks like the `lib` field of reexports may be truncated by one character:

image

I believe the expected value should be `NTDLL`:

image

lzybkr:
Fix #160

### DIFF
--- a/src/pe/export.rs
+++ b/src/pe/export.rs
@@ -142,7 +142,7 @@ impl<'a> ExportData<'a> {
 /// PE binaries have two kinds of reexports, either specifying the dll's name, or the ordinal value of the dll
 pub enum Reexport<'a> {
   DLLName { export: &'a str, lib: &'a str },
-  DLLOrdinal { export: &'a str, ordinal: usize }
+  DLLOrdinal { ordinal: usize, lib: &'a str }
 }
 
 impl<'a> scroll::ctx::TryFromCtx<'a, scroll::Endian> for Reexport<'a> {
@@ -168,7 +168,7 @@ impl<'a> scroll::ctx::TryFromCtx<'a, scroll::Endian> for Reexport<'a> {
                 if rest[0] == b'#' {
                     let ordinal = rest.pread_with::<&str>(1, scroll::ctx::StrCtx::Length(len - 1))?;
                     let ordinal = ordinal.parse::<u32>().map_err(|_e| error::Error::Malformed(format!("Cannot parse reexport ordinal from {} bytes", bytes.len())))?;
-                    return Ok((Reexport::DLLOrdinal { export: dll, ordinal: ordinal as usize }, reexport_len + 1))
+                    return Ok((Reexport::DLLOrdinal { ordinal: ordinal as usize, lib: dll }, reexport_len + 1))
                 } else {
                     let export = rest.pread_with::<&str>(0, scroll::ctx::StrCtx::Length(len))?;
                     return Ok((Reexport::DLLName { export, lib: dll }, reexport_len + 1))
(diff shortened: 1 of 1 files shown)


src/pe/export.rs
```diff
@@ -157,23 +157,21 @@ impl<'a> scroll::ctx::TryFromCtx<'a, scroll::Endian> for Reexport<'a> {
...
+                    break;
+                }
+                let len = reexport_len - o - 1;
+                let rest: &'a [u8] = bytes.pread_with(o + 1, len)?;
                 debug!("rest: {:?}", &rest);
-                let len = rest.len() - 1;
                 if rest[0] == b'#' {
-                    // UNTESTED
-                    let ordinal = rest.pread_with::<&str>(1, scroll::ctx::StrCtx::Length(len))?;
+                    let ordinal = rest.pread_with::<&str>(1, scroll::ctx::StrCtx::Length(len - 1))?;
```
philipc: Have you tested this? It doesn't seem correct to me. `rest` is already starting at `o + 1`, so I don't think we need the additional `1` offset here.

src/pe/export.rs
```diff
@@ -157,23 +157,21 @@ impl<'a> scroll::ctx::TryFromCtx<'a, scroll::Endian> for Reexport<'a> {
...
                     let ordinal = ordinal.parse::<u32>().map_err(|_e| error::Error::Malformed(format!("Cannot parse reexport ordinal from {} bytes", bytes.len())))?;
-                    // FIXME: return size
-                    return Ok((Reexport::DLLOrdinal { export: dll, ordinal: ordinal as usize }, 0))
+                    return Ok((Reexport::DLLOrdinal { export: dll, ordinal: ordinal as usize }, reexport_len + 1))
```
philipc: This is an existing problem, but this enum variant makes no sense to me. I think it would be better as:
```
Reexport::DLLOrdinal { ordinal: ordinal as usize, lib: dll }
```

src/pe/export.rs
lzybkr: I did test. I changed `rest` to start after the `.` - so that's `o + 1`.

In the ordinal case, we also need to skip the `#` (start at offset 1), and the len includes the `#`, so subtract 1.

```
#175 PS> .\target\debug\pedump.exe C:\windows\System32\certcli.dll | sls mySan

Export { name: Some("mySanitizeName"), offset: 420153, rva: 423225, size: 0, reexport: Some(DLLOrdinal { export: "CERTCA", ordinal: 831 }) }
Export { name: Some("mySanitizedNameToDSName"), offset: 420189, rva: 423261, size: 0, reexport: Some(DLLOrdinal { export: "CERTCA", ordinal: 832 }) }
Export { name: Some("mySanitizedNameToShortName"), offset: 420228, rva: 423300, size: 0, reexport: Some(DLLOrdinal { export: "CERTCA", ordinal: 833 }) }

#176 PS> link /dump /exports C:\windows\System32\certcli.dll | sls mySan

    349   A0          mySanitizeName (forwarded to CERTCA.#831)
    350   A1          mySanitizedNameToDSName (forwarded to CERTCA.#832)
    351   A2          mySanitizedNameToShortName (forwarded to CERTCA.#833)
```

philipc: Ah great, my mistake.

src/pe/export.rs
lzybkr: That seems like a reasonable change, but it would be an api breaking change. Is that OK for this PR?

philipc: The off-by-one bug and this enum variant were both added in fe11bb05. Since this code doesn't work and hasn't ever worked since this enum variant was added, I think it is fine.

m4b: Yes ok with break and fix, thanks for this PR and fixing this very long-standing bug I added :D

lzybkr: Great, I made the change. And thanks for this very useful crate.

PR #183, opened 2019-10-02
pe: Fix resolution of redirect unwind info

jan-auer:
Fixes the resolution of redirected unwind information in the PE exception table.

The issue originated in https://github.com/m4b/goblin/pull/136. Runtime function entries usually specify an offset of a `UnwindInfo` struct, which is always 4-byte aligned. However, sometimes the last bit of this offset is flipped, with indicates that instead another `RuntimeFunction` must be resolved at the offset (with the bit flipped back). This logic is handled by `get_unwind_info`.

There was a bug in the initial implementation, that treated the offset as relative, where in fact it is absolute. To fix this, `get_function_by_offset` now expects an absolute offset.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,3 +41,7 @@ archive = ["alloc"]
 [badges.travis-ci]
 branch = "master"
 repository = "m4b/goblin"
+
+[dev-dependencies]
+reqwest = "0.9.21"
+lazy_static = "1.4.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,3 @@ archive = ["alloc"]
 [badges.travis-ci]
 branch = "master"
 repository = "m4b/goblin"
-
-[dev-dependencies]
-reqwest = "0.9.21"
-lazy_static = "1.4.0"
--- a/cache/.gitignore
+++ b/cache/.gitignore
@@ -1,2 +1,0 @@
-*
-!.gitignore
--- a/cache/.gitignore
+++ b/cache/.gitignore
@@ -1,0 +1,2 @@
+*
+!.gitignore
--- a/src/pe/exception.rs
+++ b/src/pe/exception.rs
@@ -824,3 +824,187 @@ impl<'a> IntoIterator for &'_ ExceptionData<'a> {
         self.functions()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::pe::PE;
+
+    macro_rules! microsoft_symbol {
+        ($name:literal, $id:literal) => {{
+            use std::fs::File;
+            use std::path::Path;
+
+            let path = Path::new(concat!("cache/", $name));
+            if !path.exists() {
+                let url = format!(
+                    "https://msdl.microsoft.com/download/symbols/{}/{}/{}",
+                    $name, $id, $name
+                );
+
+                let mut response = reqwest::get(&url).expect(concat!("get ", $name));
+                let mut target = File::create(path).expect(concat!("create ", $name));
+                response
+                    .copy_to(&mut target)
+                    .expect(concat!("download ", $name));
+            }
+
+            std::fs::read(path).expect(concat!("open ", $name))
+        }};
+    }
+
+    lazy_static::lazy_static! {
+        static ref PE_DATA: Vec<u8> = microsoft_symbol!("WSHTCPIP.DLL", "4a5be0b77000");
+    }
+
+    #[test]
+    fn test_parse() {
+        let pe = PE::parse(&PE_DATA).expect("parse PE");
+        let exception_data = pe.exception_data.expect("get exception data");
+
+        assert_eq!(exception_data.len(), 19);
+        assert!(!exception_data.is_empty());
+    }
+
+    #[test]
+    fn test_iter_functions() {
+        let pe = PE::parse(&PE_DATA).expect("parse PE");
+        let exception_data = pe.exception_data.expect("get exception data");
+
+        let functions: Vec<RuntimeFunction> = exception_data
+            .functions()
+            .map(|result| result.expect("parse runtime function"))
+            .collect();
+
+        assert_eq!(functions.len(), 19);
+
+        let expected = RuntimeFunction {
+            begin_address: 0x1355,
+            end_address: 0x1420,
+            unwind_info_address: 0x4019,
+        };
+
+        assert_eq!(functions[4], expected);
+    }
+
+    #[test]
+    fn test_get_function() {
+        let pe = PE::parse(&PE_DATA).expect("parse PE");
+        let exception_data = pe.exception_data.expect("get exception data");
+
+        let expected = RuntimeFunction {
+            begin_address: 0x1355,
+            end_address: 0x1420,
+            unwind_info_address: 0x4019,
+        };
+
+        assert_eq!(
+            exception_data.get_function(4).expect("find function"),
+            expected
+        );
+    }
+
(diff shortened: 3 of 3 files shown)


jan-auer:
Test failures on beta seem unrelated.

m4b:
Re CI failure, the sizes appeared to have changed, which seems pretty serious?

```
---- elf::dynamic::dyn32::tests::size_of stdout ----

thread 'elf::dynamic::dyn32::tests::size_of' panicked at 'assertion failed: `(left == right)`

  left: `16`,

 right: `8`', src/elf/dynamic.rs:676:5

note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

---- elf::section_header::section_header32::tests::size_of stdout ----

thread 'elf::section_header::section_header32::tests::size_of' panicked at 'assertion failed: `(left == right)`

  left: `72`,

 right: `40`', src/elf/section_header.rs:348:5

---- elf::section_header::section_header64::tests::size_of stdout ----

thread 'elf::section_header::section_header64::tests::size_of' panicked at 'assertion failed: `(left == right)`

  left: `72`,

 right: `64`', src/elf/section_header.rs:360:5
```

I agree this shouldn't be related to your PR, but it worries me.

jan-auer:
Added tests based on a 16KB Microsoft DLL that contains such a redirect. Unfortunately, since exception data references information across the entire file, it is too error-prone to extract this into a minimal subset. On the upside, we can use this file for other tests now.

Do you think there might be legal/copyright issues with checking this in?

m4b:
I generally don’t like binary data in goblin which makes things hard. Must checked in things are legacy. But also yea maybe not a great idea to check in a Microsoft dll ... Probably not worth the risk unless someone else has any ideas ?

Really we need a story for a large test suite we can run goblin against for conformance.

Would anyone like to explore such a thing ?

jan-auer:
In this case, the problem is that it's unclear when MSVC decides to generate such records, otherwise it would be super easy to create a minimal example and extract just the necessary bits for a unit test. We could fabricate such data, but that would be modelled after the current implementation, which defeats the purpose of tests.

[`symbolic`](https://github.com/getsentry/symbolic), `goblin` and probably soon more projects like Mozilla's [`dump_syms`](https://github.com/mozilla/dump_syms) rewrite all have the same issue. I'm all +1 on creating a central repo with handcrafted binaries for testing.

How would you like to proceed with this

On a different note, the compile errors on Beta are fixed in #184 btw.

jan-auer:
Ah, one more idea: in the [some tests](https://github.com/willglynn/pdb/blob/64757397c4aea7fcb4f045b89c05c25d7a84b42e/tests/omap_address_translation.rs#L8-L22) of the `pdb` crate, we're fetching Microsoft binaries on demand for testing. Since this file is on their symbol server and will probably not go away, we could do that here, too. Then it's not checked in at least.

philipc:
The fix looks good to me.

Using the symbol server would be fine for this test, but probably isn't a solution for the general need to have binaries for testing, so maybe a central repo with binaries would be better.

jan-auer:
Removed the Microsoft symbol from the history and changed tests to download it from the symbol server. 

@m4b You'll probably have opinions on the implementation, especially since it pulls in the quite heavy `reqwest` library. Let me know if that's okay or whether you prefer a different download client.

m4b:
Wow ok before this escalates maybe the test should be in another PR. I really appreciate the work you did, but I’m really uncomfortable adding download logic into tests which mean CI can fail for random network outages.

It might be better to have a broader story for goblin and binary tests as this keeps coming up.

So for now, sorry to be a pain, could you remove the test commit and then we’ll merge ? 

Really appreciate your work :)

willglynn:
@m4b CI currently retrieves data to clone the repo (`github.com`), install Rust (`build.travis.org` => `rust-lang.org`), and install `goblin`'s dependencies (`crates.io`). It's true that the proposed change causes `goblin`'s tests to download something instead of downloading nothing, but it's also true that CI has a long list of network dependencies as it stands right now.

m4b:
Yes but those are effectively CI prefetch/setup rules and it’s easy to distinguish what failed. The test in goblin itself can now fail due to network outage or whatever, which I don’t think is ok (or normal).

Also perhaps downstream clients wants to run tests but they are on a locked down network, or rustc wants to run tests but has a whitelist on CI or etc.

I just think that the test processing data should be separate from how the data arrives, and the two shouldn’t be conflated.

(Also speaking from personal experience I have seen many problems when they are combined)

willglynn:
@m4b The test downloads the file only if it doesn't exist, which benefits local development and provides a path for locked down networks. If the download fails, it provides an error message to that effect. ¯\\\_(ツ)_/¯ I agree this shouldn't be considered a good solution though, and #185 is needed.

m4b:
I really don’t think it’s a good idea for a test to being downloading <X>, whether it caches it or not. However if everyone thinks this is the best path forward, and we don’t have a better solution then I will begrudgingly merge.

I don’t want reqwest though; since (imho) this is already a hack, I’m ok with being even more hacky and shelling out to curl for the download.

@jan-auer how do you feel about using curl instead ? We can make the test run only on Linux, which generally has curl available ? I assume the CI does too, though may have to add it.

Alternatively can still move this test part into another PR and we can fight about it there.

To summarize:

1. I don’t want network io in init tests (but I will add it if I’m outvoted 😬)
2. I definitely don’t want reqwest even as a dev dependency (although I’m flexible if there’s really compelling arguments)

jan-auer:
Let me back out the tests so that we can get the actual bug fix merged and then attempt tests in a separate PR. I'll check in a small manual test instead if I can get it working somehow.

 - I'd also prefer tests that don't require network access or an HTTP client to run. It means less dependencies and more options to retrieve the files.
 - That means that the responsibility of pulling the files should probably be outside of the actual tests. Then we also don't have to rely on `curl` being present on Windows and can get the tests to run somehow on Windows.
 - If there were an external repository (#185) or a script to pull files **before** actually running the tests, then we could easily make sure the files can be downloaded on any kind of infrastructure. Here, that could be a `before_script` in Travis, I suppose?
 - We can still feature flag the tests out by default and then just activate them on CI for convenience.

m4b:
@jan-auer let me know when you backout the test and I'll merge this ASAP, thanks for your hard work!

I'm ok with merging even with beta CI failing since we/you have another patch streamlined

jan-auer:
@m4b commented out the tests now. I believe we want to bring them back as soon as we've decided on #185, so it's probably good to keep the code around for reference. 

Also added one to assert the size of runtime functions, just in case :)

m4b:
@jan-auer I merged your other fix, just to be paranoid, could you rebase this PR to just 100% verify we're good to go? Thanks so much for your work here and the beta fix!

Cargo.toml
```diff
@@ -41,3 +41,7 @@ archive = ["alloc"]
 [badges.travis-ci]
 branch = "master"
 repository = "m4b/goblin"
+
+[dev-dependencies]
+reqwest = "0.9.21"
```
m4b: this is still a dev-dependency

Cargo.toml
```diff
@@ -41,3 +41,7 @@ archive = ["alloc"]
...
+lazy_static = "1.4.0"
```
m4b: So is this; can you remove both of these for now?

jan-auer:
D'oh, fixed. Also removed the cache directory for now.

jan-auer:
Likewise, thanks for the thorough review(s).

PR #184, opened 2019-10-03
Fix errors, warnings and lints on Beta

jan-auer:
Taking a closer look at the test failures on beta it seems like `repr(C)` is broken now on Beta. The compiler introduces additional padding around `u32` fields that are already aligned. There were a bunch of other warnings that are now fixed in this PR, but the core problem still exists.

### DIFF
--- a/src/elf/dynamic.rs
+++ b/src/elf/dynamic.rs
@@ -1,11 +1,14 @@
-
 macro_rules! elf_dyn {
     ($size:ty) => {
-        #[cfg(feature = "alloc")]
-        use scroll::{Pread, Pwrite, SizeWith};
+        // XXX: Do not import scroll traits here.
+        // See: https://github.com/rust-lang/rust/issues/65090#issuecomment-538668155
+
         #[repr(C)]
         #[derive(Copy, Clone, PartialEq, Default)]
-        #[cfg_attr(feature = "alloc", derive(Pread, Pwrite, SizeWith))]
+        #[cfg_attr(
+            feature = "alloc",
+            derive(scroll::Pread, scroll::Pwrite, scroll::SizeWith)
+        )]
         /// An entry in the dynamic array
         pub struct Dyn {
             /// Dynamic entry type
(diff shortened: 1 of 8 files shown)


jan-auer:
Okay, still not quite sure what the correct fix for this is, but here's the reason for the regression on the example of the failing `elf::dynamic` test.

The problem is that `Dyn` somehow ends up in `goblin::elf::dynamic`, which gets publicly imported in `dyn32` and shadows the 32-bit `Dyn` implementation.

jan-auer:
I'm just not able to read code. Of course, `Dyn` was declared explicitly in `goblin::elf::dynamic`. 

Due to some change in the beta compiler, the outer `Dyn` now shadows the inner one in `dyn32`. I still think this is to be considered a regression in the compiler, but independently, it's not clean do rely on this here. I updated the code to avoid shadowing now.

petrochenkov:
Less invasive workaround for the glob shadowing issue - https://github.com/rust-lang/rust/issues/65090#issuecomment-538668155.

m4b:
Weird 😲 for either workaround id like a reference to the rustc issue so people aren’t confused and if it gets fixed we can go back to what it was ?

petrochenkov:
>id like a reference to the rustc issue
The link in the previous comment refers to that issue - https://github.com/rust-lang/rust/issues/65090.

m4b:
This is a huge change with a huge diff can we just do petrochenkovs change ? I can push it

m4b:
I just meant a link in code for the fix :)

jan-auer:
Uh yeah, sorry for the big diff. Contains lots of indentation. I agree that it’s a little large. If you prefer the other fix, I can change it. 

On a personal note, I’m not a big fan of shadowing. It does not seem too clean and can lead to situations that are hard to debug (as seen in this instance).

jan-auer:
Or feel free to change this PR :)

Please just retain the first commit so that we get the warnings fixed, too.

m4b:
@jan-auer I have applied @petrochenkov fix but I think you need to allow me to push onto your branch? Not sure; I tried various variations on:

```
git push jan-auer fix/beta
```

but my git skills probably need work. Do you know a quick command to push to your remote branch? Nothing seems to work for me:

(and `jan-auer` is `git remote add jan-auer https://github.com/jan-auer/goblin`)

jan-auer:
@m4b not sure, sounds like this _should_ have worked :)

Rebased the branch now.

Issue #181, opened 2019-08-18
PE: meet invalid name rva in ImportDirectoryEntry and panic


wyxloading:
I have met some nasty exe that have a invalid .idata section, which will panic.
Something like:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Malformed("Cannot find name from rva 0x43720004 in sections ...
```
code: https://github.com/m4b/goblin/blob/master/src/pe/import.rs#L147-L148

In this moment i actually don't know why these exe have such .idata. I just want to parse the unwind info.
Now `pe::PE::parse` will parse the whole pefile.
I am wondering would it be nice to pass some options to enable/disable parse partially.
Like i would disable parsing the import table as i just want to get the exception table (for the unwind info).

BTW, i use gimli/object to help parsing object files. Thanks for the great job.
I don't know where i should fire this issue (goblin or gimli/object?)

wyxloading:
I think this issue more like discussion https://github.com/m4b/goblin/issues/120#issuecomment-470402558. That's what i want.

philipc:
How sure are you that the `.idata` section is invalid? Can you provide the binary?

I think we shouldn't fail for this error though, so if you can fix it to at least keep going then that'd be great.

Partial parsing would be a larger change, and is not something I want to comment on at this time, but maybe others have thoughts on that.

wyxloading:
I looked into the .idata section, think it's not stick to what [pe format](https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#import-directory-table) has wrote.
But llvm just skip all the import table entries if meet any entry without an valid rva.

Anyway, here the sample exe
[xy2.zip](https://github.com/m4b/goblin/files/3518129/xy2.zip)

PR #188, opened 2019-10-06
Added a feature to disable pe rva resolve for already mapped modules

ko1N:
It would be extremely helpful to use the capabilities of goblin when working with memory/process dumps as well as files on disk. Since memory dumps are already mapped we can just skip over the rva resolve function and all of the other functionality will work still.

I'm not sure however if it would be best to implement this as a compile-time feature (as this PR does it) or as a runtime feature (maybe splitting the parse function into parse_image / parse_mapped or something similiar.

### DIFF
--- a/src/pe/debug.rs
+++ b/src/pe/debug.rs
@@ -2,6 +2,7 @@ use crate::error;
 use scroll::{Pread, Pwrite, SizeWith};
 
 use crate::pe::data_directories;
+use crate::pe::options;
 use crate::pe::section_table;
 use crate::pe::utils;
 
@@ -18,10 +19,26 @@ impl<'a> DebugData<'a> {
         sections: &[section_table::SectionTable],
         file_alignment: u32,
     ) -> error::Result<Self> {
+        Self::parse_with_opts(
+            bytes,
+            dd,
+            sections,
+            file_alignment,
+            &options::ParseOptions::default(),
+        )
+    }
+
+    pub fn parse_with_opts(
+        bytes: &'a [u8],
+        dd: data_directories::DataDirectory,
+        sections: &[section_table::SectionTable],
+        file_alignment: u32,
+        opts: &options::ParseOptions,
+    ) -> error::Result<Self> {
         let image_debug_directory =
-            ImageDebugDirectory::parse(bytes, dd, sections, file_alignment)?;
+            ImageDebugDirectory::parse_with_opts(bytes, dd, sections, file_alignment, opts)?;
         let codeview_pdb70_debug_info =
-            CodeviewPDB70DebugInfo::parse(bytes, &image_debug_directory)?;
+            CodeviewPDB70DebugInfo::parse_with_opts(bytes, &image_debug_directory, opts)?;
 
         Ok(DebugData {
             image_debug_directory,
@@ -59,14 +76,31 @@ pub const IMAGE_DEBUG_TYPE_FIXUP: u32 = 6;
 pub const IMAGE_DEBUG_TYPE_BORLAND: u32 = 9;
 
 impl ImageDebugDirectory {
+    #[allow(unused)]
     fn parse(
         bytes: &[u8],
         dd: data_directories::DataDirectory,
         sections: &[section_table::SectionTable],
         file_alignment: u32,
     ) -> error::Result<Self> {
+        Self::parse_with_opts(
+            bytes,
+            dd,
+            sections,
+            file_alignment,
+            &options::ParseOptions::default(),
+        )
+    }
+
+    fn parse_with_opts(
+        bytes: &[u8],
+        dd: data_directories::DataDirectory,
+        sections: &[section_table::SectionTable],
+        file_alignment: u32,
+        opts: &options::ParseOptions,
+    ) -> error::Result<Self> {
         let rva = dd.virtual_address as usize;
-        let offset = utils::find_offset(rva, sections, file_alignment).ok_or_else(|| {
+        let offset = utils::find_offset(rva, sections, file_alignment, opts).ok_or_else(|| {
             error::Error::Malformed(format!(
                 "Cannot map ImageDebugDirectory rva {:#x} into offset",
                 rva
@@ -94,6 +128,14 @@ pub struct CodeviewPDB70DebugInfo<'a> {
 
 impl<'a> CodeviewPDB70DebugInfo<'a> {
     pub fn parse(bytes: &'a [u8], idd: &ImageDebugDirectory) -> error::Result<Option<Self>> {
+        Self::parse_with_opts(bytes, idd, &options::ParseOptions::default())
+    }
+
+    pub fn parse_with_opts(
+        bytes: &'a [u8],
+        idd: &ImageDebugDirectory,
+        opts: &options::ParseOptions,
+    ) -> error::Result<Option<Self>> {
         if idd.data_type != IMAGE_DEBUG_TYPE_CODEVIEW {
             // not a codeview debug directory
             // that's not an error, but it's not a CodeviewPDB70DebugInfo either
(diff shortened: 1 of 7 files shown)


m4b:
Hi @ko1N thanks for the PR! The use of features is a nice quick fix, but from what I understand, cargo features are meant to be "additive"; e.g., the semantics/ behavior of a function shouldn't typically change depending on which feature is added, e.g., someone activating this feature on top of someone else not, can break everyone's build, or make it behavior strangely, so-called action at a distance.

That being said, I'm not sure what the best solution is here; parse_image/parse_mapped sounds like a lot of boilerplate and a lot of work, whereas your current solution is quite clean, though it has the problems above.

Does anyone else have any suggestions here?

ko1N:
Hello @m4b! I gave this issue a bit more thought and ye, you're right. We shouldn't change functionality through a feature as it would prevent people from using it in both ways in the same program.

I think a clean solution would be to change the parse() function to additionally accept something like a ParseOptions struct. We could also add another function which wraps the new parse function with some default options to not introduce an api change.
Maybe this could also be beneficial for other use cases (e.g. adding an option which will make the parser instantly fail on error, opt-out/opt-in for specific parts of the header being parsed, etc).

m4b:
Parser options depending on how they’re implemented could be interesting; might also be a segue into partial parsing or other effects on how goblin parses ?

Eg https://github.com/m4b/goblin/issues/181

ko1N:
I removed the feature and added a runtime checked struct which will be passed through the relevant parse() calls. I also added wrappers (parse / parse_with_opts) to keep all the tests working and the API kind of stable.

Should we also add wrappers to all the other functions in pe::utils to keep its API consistent (try_name, find_offset_or, etc)?

ko1N:
Yes, you are totally right. I forgot to include the other parse() functions in my earlier commit. I'm not sure if we should write wrappers for the pe::utils though?

src/pe/options.rs
```diff
@@ -0,0 +1,12 @@
+#[derive(Debug, Copy, Clone)]
+pub struct ParseOptions {
+    pub rva: bool,
```
m4b: the field for this seems a little misleading; isn't it more like `resolve_rva`, or even `not_memory_mapped` (latter not preferable because it encodes booleanness, but i think you get the point?)

m4b:
Could this change be used for a windows backtrace functionality, e.g., in backtrace-rs with gimli feature? @philipc or @willglynn maybe you'd know?

m4b:
@ko1N you may have to rebase on master, I just merged another change for PE stuff

philipc:
@m4b Are you referring to unwinding or symbolication? The gimli feature in backtrace-rs is only for symbolication, and it already works on windows AFAIK.

m4b:
symbolification; interesting maybe this comment needs to be updated (I understood that it only works on linux atm): https://github.com/rust-lang/backtrace-rs/blob/master/Cargo.toml#L94

could this change help in unwinding? I'm asking random questions, haven't thought about it at all :P

philipc:
Yeah I think that comment needs updating, there's window support [here](https://github.com/rust-lang/backtrace-rs/blob/816a9559c116f1224c9d3ac968bdc1e6be14fcf1/src/symbolize/gimli.rs#L85). I don't think this has been needed there because we don't need to use RVAs to find sections or symbols.

For unwinding it's probably helpful since my understanding is the unwind info contains RVAs that point to other parts of the unwind info.

src/pe/exception.rs
```diff
@@ -766,29 +778,49 @@ impl<'a> ExceptionData<'a> {
...
         let rva = function.unwind_info_address as usize;
-        let offset = utils::find_offset(rva, sections, self.file_alignment).ok_or_else(|| {
+        let offset = utils::find_offset(rva, sections, self.file_alignment, opts).ok_or_else(|| {
             error::Error::Malformed(format!("cannot map unwind rva ({:#x}) into offset", rva))
         })?;
 
         UnwindInfo::parse(self.bytes, offset)
     }
 
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
ko1N: Internally those parse() functions are not called anymore and would lead to a dead code warning.

ko1N:
If I see it correctly libunwind-rs uses its own name to parse the file on disk with goblin right now. See [https://github.com/rust-lang/backtrace-rs/blob/b7fa382eb82ea2dbec36e7b58d63d0b1d5cdd011/src/symbolize/gimli.rs#L508](here) @m4b .

By removing the find_offset function to resolve the RVA we can now simply use a memory copy of a running process and parse the PE header with it. E.g. you can run a tool like pedump to dump the entire process memory on disk and parse the resulting file with goblin. Or you can use it do parse the PE header of a running process. You can even use it to parse a kernel memory dump now (as the kernel is just a PE executable as well).

src/pe/options.rs
```diff
@@ -0,0 +1,12 @@
+#[derive(Debug, Copy, Clone)]
+pub struct ParseOptions {
```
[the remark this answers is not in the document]
ko1N: Added those

ko1N:
When playing around I found that the debug guid wasnt reported with disabled rva resolve. I added a runtime switch where it would select either pointer_to_raw_data or address_of_raw_data in the codeview debugging parser depending on the configuration.

I'm not entirely sure if we need this in the CoffHeader parser as well. Maybe someone could double-check on this.

m4b:
@ko1n did you need help to push this through ?

What’s the status, your last comment is remaining issue before can be merged ?

I like this change. Is there anyway to tell at parse time whether the PE would require rva resolution or no? Would be nice if Eg, bingrep, just “worked” if I passed it a process dump or executable.

ko1N:
I checked back on it but I don't see any obvious issue holding the merge back right now. I'm already using goblin to parse some of those mapped executables and it works fairly well.
There is also very minimal risk of breaking tools which make use of goblin right now.

However it would still be a good idea to test this functionality. Do you think I should compile and dump a test executable so we can write an automated test for it. Or do you think we should wait on the binary test suite for that matter?

ko1N:
Do you have any update on the testing situation?

m4b:
@ko1N major apologies; i'm not sure why this wasn't merged; i don't see any backwards incompatible changes here, and it seems like a reasonable api to have, albeit somewhat niche (but that's fine!)

If you're still interested, i can help rebase this, and push; if not i may close this, let me know whenever you find the time, thanks!

I'll leave this open for now

ko1N:
I squashed and rebased the commit to the current master.
It's a niche feature but I intend to use exactly this in my [memflow](https://github.com/memflow/memflow) project.

src/pe/debug.rs
```diff
@@ -59,14 +76,31 @@ pub const IMAGE_DEBUG_TYPE_FIXUP: u32 = 6;
 pub const IMAGE_DEBUG_TYPE_BORLAND: u32 = 9;
 
 impl ImageDebugDirectory {
+    #[allow(unused)]
```
m4b, line 79: nit: i think it would just be better to delete this and the other unused private function, but not important

m4b (approved):
Looks great! Sorry this took so long; so i ran your patch with `rdr` example on a bunch of windows system libraries, e.g., in `System32` and nothing had a non-zero error code, so i'm good with merging this. thanks for being so patient!

m4b:
@ko1N would you like a release? There are a couple changes coming, one of them breaking change, but i can branch and cherry-pick this and release a 0.3.2 so you can get off of git or whatever :) you've waited long enough

m4b:
ok released as 0.3.3 :)

src/pe/debug.rs
ko1N: My intention was to keep the API compatible here (even if existing code calls parse on something like ImageDebugDirectory manually). Maybe a better idea would be to mark these functions as deprecated and remove them in a next major release.

ko1N:
Thanks so much! 👍

PR #191, opened 2019-11-05
Format code using 'cargo fmt'

Atul9:


Atul9:
closing since 1.0 isn't released yet - https://github.com/m4b/goblin/pull/150

PR #195, opened 2019-11-30
mach: don't return data for zerofill sections

philipc:
I'll also guess that the big comment I'm indenting here was a result of these sections (it's still good to have correct handling for truncated files of course).


m4b:
Yup you’re probably right :)

m4b:
released in 0.1.2

PR #196, opened 2019-12-12
Delete `feature(alloc)`

philipc:
The feature `alloc` has been stable since 1.36.0.

The main change is in lib.rs, the rest is cleanup.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,7 +10,7 @@
 //! * A PE32/PE32+ (64-bit) parser, and raw C structs
 //! * A Unix archive parser and loader
 //!
-//! Goblin requires at least `rustc` 1.31.1, uses the 2018 rust edition, and is developed on stable.
+//! Goblin requires at least `rustc` 1.36.0, uses the 2018 rust edition, and is developed on stable.
 //!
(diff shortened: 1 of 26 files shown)


src/lib.rs
```diff
@@ -79,24 +79,14 @@
...
 #![cfg_attr(not(feature = "std"), no_std)]
-#![cfg_attr(all(feature = "alloc", not(feature = "std")), feature(alloc))]
 
 #[cfg(feature = "std")]
 extern crate core;
 
-#[cfg(all(feature = "alloc", not(feature = "std")))]
+#[cfg(feature = "alloc")]
 #[macro_use]
 extern crate alloc;
```
m4b: So this is like a “magical” external crate now, similar to core ? I just realized we had this before. And I don’t really know how it works haha

So In 1.36 what changer is that we can use this now without the alloc mod simulating it below for std targets ? (I guess stabilization means it’s available for both no_std and std configurations ?)

I should probably just read the rfc. 

Anyway thanks for the refactor !

philipc: Yes, it's the same as core, and it's always available, so we can use it all the time even for std builds on stable rust.

PR #198, opened 2019-12-21
split elf::Sym definition out of alloc gate

lzutao:
Also derive Copy trait because sym{32,64}::Sym have it
See the non-whitespace diff: https://github.com/m4b/goblin/pull/198/files?utf8=%E2%9C%93&diff=unified&w=1

### DIFF
--- a/src/elf/sym.rs
+++ b/src/elf/sym.rs
@@ -163,84 +163,81 @@ macro_rules! elf_sym_std_impl {
             }
         }
 
-        if_alloc! {
-            use crate::elf::sym::Sym as ElfSym;
+        use crate::elf::sym::Sym as ElfSym;
 
-            use core::fmt;
-            use core::slice;
+        use core::fmt;
+        use core::slice;
 
-            if_std! {
-                use crate::error::Result;
-
-                use std::fs::File;
-                use std::io::{Read, Seek};
-                use std::io::SeekFrom::Start;
-            }
-
-            impl Sym {
-                /// Checks whether this `Sym` has `STB_GLOBAL`/`STB_WEAK` bind and a `st_value` of 0
-                #[inline]
-                pub fn is_import(&self) -> bool {
-                    let bind = self.st_info >> 4;
-                    (bind == STB_GLOBAL || bind == STB_WEAK) && self.st_value == 0
-                }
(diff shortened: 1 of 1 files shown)


src/elf/sym.rs
```diff
@@ -320,112 +318,111 @@ pub mod sym64 {
...
+                sym32::SIZEOF_SYM
+            },
+            Container::Big => {
+                sym64::SIZEOF_SYM
+            },
         }
     }
+}
 
+if_alloc! {
```
m4b: How come the TryFromCtx impls require alloc ?

lzutao: The error::Error has variant `Malformed(String)` which requires allocation.

m4b: Ah yea. Hmmm, might be worth investigating whether static strs could replace or more variants. Anyway thanks for PR!

philipc:
This is fine if it is useful to you, but it seems pointless without `impl From<Sym> for ElfSym` and `impl From<ElfSym> for Sym`. Can you give your motivation for why this is useful, and why only for `Sym` (there's other unified structs that are currently alloc only too)?

lzutao:
I am trying to unlock GnuHash in #151 , the GnuHash::find return a Sym
but GnuHash itself is not under alloc gate. So the test failed.

> it seems pointless without `impl From<Sym> for ElfSym` and `impl From<ElfSym> for Sym`

I would move that out of alloc gate too.

src/elf/sym.rs
```diff
@@ -163,93 +163,92 @@ macro_rules! elf_sym_std_impl {
...
-            pub unsafe fn from_raw<'a>(symp: *const Sym, count: usize) -> &'a [Sym] {
-                slice::from_raw_parts(symp, count)
-            }
+        #[inline]
+        pub unsafe fn from_raw<'a>(symp: *const Sym, count: usize) -> &'a [Sym] {
+            slice::from_raw_parts(symp, count)
+        }
+
+        if_alloc! {
+            if_std! {
```
philipc: I think you can delete the `if_alloc! {}` if everything inside it is gated by `if_std!`.

lzutao: Yeah, that's right

PR #200, opened 2019-12-23
elf: handle zero offset in SectionHeader::parse

### LINKED ISSUE #199: Some problems with parsing core dumps
Hello!
I decided to continue my experiments with parsing core dumps. And the first thing I've done: I updated the goblin. Unfortunately, the current version returns a error, when i tried to parse some core dumps: **parse object: Scroll(TooBig { size: 1905784752, len: 258048 })**.  I was debugging a little  trying to locate a problem.
Screenshot at 2019-12-22 16:57:59
Here, we've got bad sizes and received an error. Maybe problem is that there is no sections in core file?
Here  is the output of readelf tool:
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00 
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              CORE (Core file)
  Machine:                           Advanced Micro Devices X86-64
  Version:                           0x1
  Entry point address:               0x0
  Start of program headers:          64 (bytes into file)
  Start of section headers:          0 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
  Size of program headers:           56 (bytes)
  Number of program headers:         18
  Size of section headers:           0 (bytes)
  Number of section headers:         0
  Section header string table index: 0

There are no sections in this file.

There are no sections to group in this file.

Readelf and old version of goblin could parse my core file. Any ideas?

philipc:
Regression in 779d0ce8 (which wasn't reviewed...).
Fixes #199, cc @xcoldhandsx

### DIFF
--- a/src/elf/section_header.rs
+++ b/src/elf/section_header.rs
@@ -435,16 +435,22 @@ if_alloc! {
         }
         /// Parse `count` section headers from `bytes` at `offset`, using the given `ctx`
         #[cfg(feature = "endian_fd")]
-        pub fn parse(bytes: &[u8], mut offset: usize, count: usize, ctx: Ctx) -> error::Result<Vec<SectionHeader>> {
+        pub fn parse(bytes: &[u8], mut offset: usize, mut count: usize, ctx: Ctx) -> error::Result<Vec<SectionHeader>> {
             use scroll::Pread;
-            let mut section_headers = Vec::with_capacity(count);
-            let mut nsection_headers = count;
+            // Zero offset means no section headers, not even the null section header.
+            if offset == 0 {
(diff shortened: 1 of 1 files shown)


xcoldhandsx:
I've tested on my example and it seems, that it works, thanks.

m4b:
Ugh m4b breaking shit. Thanks @xcoldhandsx for reporting and @philipc for fixing this !

m4b:
1.3 is released with this regression fixed

PR #201, opened 2019-12-24
elf.note: added new constants for structs: prstatus, prpsinfo, siginfo_t, mapped files.

xcoldhandsx:
Some constants for very useful notes.


m4b:
just curious, where did you find these constants?

xcoldhandsx:
I found these constants in readelf sources. Here is the link: https://github.com/bminor/binutils-gdb/blob/8dc433a0fb04f8fa37530d0788053dd9bea5c37d/include/elf/common.h

m4b:
Ok we can add them, are you encountering them in the wild ?

xcoldhandsx:
Yes. These constants are always in core dumps, representing notes for corresponding structs. I use these constants to check note types, when i try to retrieve some interesting data from core file (signal, registers and its values, process map, etc.)

m4b:
thanks for the PR ! :)

PR #202, opened 2020-01-04
use inner cfg attribute instead of macro

lzutao:
[Without whitespace diff](https://github.com/m4b/goblin/pull/202/files?utf8=%E2%9C%93&diff=unified&w=1)


lzutao:
Looks like we still imports type not witin the cfg attribute.

PR #218, opened 2020-05-05
Fix panic in parsing UnwindInfo

jan-auer:
The exception handler address incorrectly shadowed the `offset` for reading exception handler data. This could lead to out-of-bounds access which and results in a panic.

Instead, the address should be treated as an opaque pointer (relative, probably RVA), while the data follows the pointer. See https://docs.microsoft.com/en-us/cpp/build/exception-handling-x64

### DIFF
--- a/src/pe/exception.rs
+++ b/src/pe/exception.rs
@@ -587,13 +587,13 @@ impl<'a> UnwindInfo<'a> {
         // whenever flags UNW_FLAG_EHANDLER or UNW_FLAG_UHANDLER are set. The language-specific
         // handler is called as part of the search for an exception handler or as part of an unwind.
         } else if flags & (UNW_FLAG_EHANDLER | UNW_FLAG_UHANDLER) != 0 {
-            let offset = bytes.gread_with::<u32>(&mut offset, scroll::LE)? as usize;
+            let address = bytes.gread_with::<u32>(&mut offset, scroll::LE)?;
             let data = &bytes[offset..];
 
             handler = Some(if flags & UNW_FLAG_EHANDLER != 0 {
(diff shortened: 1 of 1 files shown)


src/pe/exception.rs
```diff
@@ -587,13 +587,13 @@ impl<'a> UnwindInfo<'a> {
         // whenever flags UNW_FLAG_EHANDLER or UNW_FLAG_UHANDLER are set. The language-specific
         // handler is called as part of the search for an exception handler or as part of an unwind.
         } else if flags & (UNW_FLAG_EHANDLER | UNW_FLAG_UHANDLER) != 0 {
-            let offset = bytes.gread_with::<u32>(&mut offset, scroll::LE)? as usize;
+            let address = bytes.gread_with::<u32>(&mut offset, scroll::LE)?;
             let data = &bytes[offset..];
```
m4b, line 591: should data use `offset` or `address` here?

m4b:
Once confirm offset is ok @jan-auer ill merge this and push a new release with the other changes from @jackcmay ; i think i turned off the travis-ci job too, so hopefully that won't be annoying anymore.

src/pe/exception.rs
jan-auer: Thanks for asking, data should in fact use `offset`. 

For posterity: This is documented [here](https://docs.microsoft.com/en-us/cpp/build/exception-handling-x64?view=vs-2019#struct-unwind_info). With the handler flags set, `struct UNWIND_INFO` contains a `u32` address of the exception handler, followed immediately by variable length data.

m4b:
thanks for the PR @jan-auer

PR #220, opened 2020-05-05
Fix build and clippy warnings

jackcmay:
Fix build and clippy warnings

Would be nice to enforce this in CI :-)

### DIFF
--- a/src/elf/note.rs
+++ b/src/elf/note.rs
@@ -49,10 +49,10 @@ pub const NT_PRSTATUS: u32 = 1;
 pub const NT_PRPSINFO: u32 = 3;
 
 ///Fields of siginfo_t.
-pub const NT_SIGINFO: u32 = 0x53_494_749;
+pub const NT_SIGINFO: u32 = 0x5349_4749;
 
 ///Description of mapped files.
-pub const NT_FILE: u32 = 0x46_494_c45;
+pub const NT_FILE: u32 = 0x4649_4c45;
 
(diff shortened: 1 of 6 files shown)


src/elf/note.rs
```diff
@@ -49,10 +49,10 @@ pub const NT_PRSTATUS: u32 = 1;
 pub const NT_PRPSINFO: u32 = 3;
 
 ///Fields of siginfo_t.
-pub const NT_SIGINFO: u32 = 0x53494749;
+pub const NT_SIGINFO: u32 = 0x53_494_749;
```
jan-auer, line 52: Hex literals are probably more readable in groups of 4 (same below):

```suggestion
pub const NT_SIGINFO: u32 = 0x5349_4749;
```

jackcmay: @jan-auer Agreed, fixed

jackcmay:
@m4b 👀

src/elf/program_header.rs
```diff
@@ -314,6 +314,9 @@ macro_rules! elf_program_header_std_impl { ($size:ty) => {
                 phdrs
             }
 
+            /// # Safety
+            ///
+            /// This function creates a `ProgramHeader` directly from a raw pointer
```
m4b, line 319: nit: this should probably a note that the phnum value is blindly trusted and not validated in anyway (though not sure how you could)? but also maybe people should already know that if they're using this? dunno

PR #222, opened 2020-05-10
Replaced CoffHeader::strings unwrap with try.

ExPixel:
The `unwrap` was causing some problems in my code and it seems unnecessary since the function returns the same Result type as what is being unwrapped anyway.

### DIFF
--- a/src/pe/header.rs
+++ b/src/pe/header.rs
@@ -99,7 +99,7 @@ impl CoffHeader {
         let offset = self.pointer_to_symbol_table as usize
             + symbol::SymbolTable::size(self.number_of_symbol_table as usize);
         let length = bytes.pread_with::<u32>(offset, scroll::LE)? as usize;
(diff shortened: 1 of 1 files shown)


src/pe/header.rs
```diff
@@ -99,7 +99,7 @@ impl CoffHeader {
         let offset = self.pointer_to_symbol_table as usize
             + symbol::SymbolTable::size(self.number_of_symbol_table as usize);
         let length = bytes.pread_with::<u32>(offset, scroll::LE)? as usize;
-        Ok(strtab::Strtab::parse(bytes, offset, length, 0).unwrap())
+        Ok(strtab::Strtab::parse(bytes, offset, length, 0)?)
```
m4b, line 102: interesting; not sure why this was ever unwrapped, parsing the strtab should be fallible here, so it must have gone unnoticed

PR #223, opened 2020-05-10
Added more COFF machine constants.

ExPixel:
Added all of the COFF machine constants found here:
https://docs.microsoft.com/en-us/windows/win32/debug/pe-format

(sorry about the multiple pull requests)

### DIFF
--- a/src/pe/header.rs
+++ b/src/pe/header.rs
@@ -56,8 +56,56 @@ pub const SIZEOF_COFF_HEADER: usize = 20;
 /// PE\0\0, little endian
 pub const PE_MAGIC: u32 = 0x0000_4550;
 pub const SIZEOF_PE_MAGIC: usize = 4;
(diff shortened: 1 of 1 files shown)


ExPixel:
The constants can be found here: https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#machine-types . Seems to match up with what you already had for x86 and x86_64.

m4b:
Awesome, appreciate the PR :)

m4b:
ok this along with unwrap is in 0.2.3 on crates.io, thanks for the PRs :)

jan-auer:
FWIW, we also have the [`MachineType`](https://github.com/willglynn/pdb/blob/40029549cc038988b358b02e27b3cc960355ded2/src/dbi.rs#L250) enum in `pdb`. If there are more cases like this, we could create a minimal `pe` crate that exposes these to both `goblin` and `pdb`.

Issue #231, opened 2020-07-28
missing arch names on OSX


blinsay:
I've been playing around with Goblin on OSX and noticed that the arch names for my multi-arch binaries aren't being output correctly. `file` seems to get it correct, so I don't think there's anything strange going on with my system.

Thanks for a cool project, let me know if I can provide anything else that's helpful.

```
 $ file /usr/bin/ruby
/usr/bin/ruby: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [i386:Mach-O executable i386]
/usr/bin/ruby (for architecture x86_64):        Mach-O 64-bit executable x86_64
/usr/bin/ruby (for architecture i386):  Mach-O executable i386
$ cargo run -q /usr/bin/ruby
/usr/bin/ruby
- unknown (cputype=0x01000007, cpusubtype=0x80000003)
- i386 (cputype=0x00000007, cpusubtype=0x00000003)
```

```rust
use goblin::mach::constants::cputype::get_arch_name_from_types;
use goblin::mach::Mach;
use std::env::args;
use std::fs::read;

fn main() {
    for path in args().skip(1) {
        println!("{}", path);
        let buf = read(path).unwrap();

        if let Ok(Mach::Fat(m)) = Mach::parse(&buf) {
            for i in 0..m.narches {
                let a = m.get(i).unwrap();
                let cputype = get_arch_name_from_types(a.header.cputype, a.header.cpusubtype)
                    .unwrap_or("unknown");
                println!(
                    "- {} (cputype=0x{:08x}, cpusubtype=0x{:08x})",
                    cputype, a.header.cputype, a.header.cpusubtype
                );
            }
        }
    }
}
```

blinsay:
Expanded the repro a bit to scan for binaries where the arch name is `None` and it seems like it's always the `cputype=0x01000007 cpusubtype=0x80000003` pair.

Evian-Zhang:
When using `get_arch_name_from_types` for Mach-O files, we should use `a.header.cputype()` and `a.header.cpusubtype()` instead of `a.header.cputype` and `a.header.cpusubtype`. This may help you solve this problem. I would like to open a PR to add more documentation for this method and add some explanation.

m4b:
@blinsay I believe @Evian-Zhang comment + documentation changes answers your issue? If yes, would you mind closing? And thanks again for opening the issue!

blinsay:
Using the method worked, thanks for clarifying!

[...]

PR #232, opened 2020-08-16
mach.constant: add more document for get_arch_name_from_types to avoid misuse

Evian-Zhang:
Related issue: #231 

# Problem

Some users may use `cputype` and `cpusubtype` directly to get architecture name, which may leads to wrong answer.

For example:

```rust
let buf = read("path/to/macho").unwrap();
if let Ok(Mach::Binary(a)) = Mach::parse(&buf) {
    println!("arch name: {}", get_arch_name_from_types(a.header.cputype, a.header.cpusubtype).unwrap());
}
```

When apply this to [arm64e](https://developer.apple.com/documentation/security/preparing_your_app_to_work_with_pointer_authentication) arch Mach-O file (which may be the main arch of iOS app and Apple Silicon Mac's app), the program crashes because the `get_arch_name_from_types` returns `None`.

# Reason

From Apple's open source [lipo](https://opensource.apple.com/source/cctools/cctools-949.0.1/misc/lipo.c.auto.html) code, we can find the code to display an executable's arch name is:

```c
case CPU_TYPE_ARM64:
    switch(cpusubtype & ~CPU_SUBTYPE_MASK){
        case CPU_SUBTYPE_ARM64_ALL:
            printf("arm64");
            break;
        case CPU_SUBTYPE_ARM64_V8:
            printf("arm64v8");
            break;
        case CPU_SUBTYPE_ARM64E:
            printf("arm64e");
            break;
        default:
            goto print_arch_unknown;
    }
    break;
```

(This branch is for CPU of type ARM64, for other CPU types, the code is similar)

It tests against the `cpusubtype & ~CPU_SUBTYPE_MASK` instead of `cpusubtype`.

And if we look at the code in [`src/mach/header.rs`](https://github.com/m4b/goblin/blob/master/src/mach/header.rs), we can find provided methods

```rust
impl Header {
    /// Returns the cpu type
    pub fn cputype(&self) -> CpuType {
        self.cputype
    }
    /// Returns the cpu subtype with the capabilities removed
    pub fn cpusubtype(&self) -> CpuSubType {
        self.cpusubtype & !CPU_SUBTYPE_MASK
    }
}
```

# Solution

As a result, we should use `cputype()` and `cpusubtype()` method instead of `cputype` and `cpusubtype` field to pass to `get_arch_name_from_types` method to  get the right architecture name. So I add some documentation for this method.

### DIFF
--- a/src/mach/constants.rs
+++ b/src/mach/constants.rs
@@ -367,10 +367,12 @@ pub mod cputype {
             /// use goblin::mach::constants::cputype::get_arch_name_from_types;
             /// use goblin::mach::Mach;
             ///
-            /// let buf = read("path/to/macho").unwrap();
-            /// if let Ok(Mach::Binary(a)) = Mach::parse(&buf) {
-            ///     println!("arch name: {}", get_arch_name_from_types(a.header.cputype(), a.header.cpusubtype()).unwrap());
-            /// }
+            /// read("path/to/macho").and_then(|buf| {
+            ///     if let Ok(Mach::Binary(a)) = Mach::parse(&buf) {
+            ///         println!("arch name: {}", get_arch_name_from_types(a.header.cputype(), a.header.cpusubtype()).unwrap());
+            ///     }
+            ///     Ok(())
+            /// });
             /// ```
             pub fn get_arch_name_from_types(cputype: CpuType, cpusubtype: CpuSubType)
                 -> Option<&'static str> {
--- a/src/mach/constants.rs
+++ b/src/mach/constants.rs
@@ -351,6 +351,23 @@ pub mod cputype {
             }
 
             /// Get the architecture name from cputype and cpusubtype
+            ///
+            /// When using this method to determine the architecture
+            /// name of an instance of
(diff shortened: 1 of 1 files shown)


src/mach/constants.rs
```diff
@@ -351,6 +351,27 @@ pub mod cputype {
...
+            /// instead of corresponding field `cputype` and `cpusubtype`.
+            ///
+            /// For example:
+            ///
+            /// ```rust
+            /// use std::fs::read;
+            /// use goblin::mach::constants::cputype::get_arch_name_from_types;
+            /// use goblin::mach::Mach;
+            ///
+            /// let buf = read("path/to/macho").unwrap();
```
m4b, line 370: I believe this is causing CI failure. I’d suggest to use and_then in example so unwrap doesn’t fall or experiment at using one of the checked in macho binaries. Former probably better for example.

Evian-Zhang:
> Cool thanks for great write up and the example plus documentation ! I commented about CI failure, otherwise once fixed LGTM, and thanks again !

OK, fixed it.

m4b:
I'm going to squash and merge, thanks for this!

PR #236, opened 2020-08-19
pe: add basic validations

Evian-Zhang:
From the Microsoft [PE format](https://docs.microsoft.com/en-us/windows/win32/debug/pe-format) spec, we can do some basic validations for the parsed PE and COFF object, such as some fields should be set to zero for PE image but not for COFF object.

The validation is very basic and can be further completed in the future.

### DIFF
--- a/src/pe/characteristic.rs
+++ b/src/pe/characteristic.rs
@@ -73,6 +73,7 @@ let show_type characteristics =
   else "MANY"                   (* print all *)
  */
 use crate::error;
+use alloc::borrow::ToOwned;
 
 pub const IMAGE_FILE_RELOCS_STRIPPED: u16 = 0x0001;
 pub const IMAGE_FILE_EXECUTABLE_IMAGE: u16 = 0x0002;
--- a/src/pe/characteristic.rs
+++ b/src/pe/characteristic.rs
@@ -72,6 +72,7 @@ let show_type characteristics =
   else if (has IMAGE_FILE_EXECUTABLE_IMAGE characteristics) then "EXE"
   else "MANY"                   (* print all *)
  */
+use crate::error;
 
 pub const IMAGE_FILE_RELOCS_STRIPPED: u16 = 0x0001;
 pub const IMAGE_FILE_EXECUTABLE_IMAGE: u16 = 0x0002;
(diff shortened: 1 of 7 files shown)


Evian-Zhang:
The CI error is very strange. I use `cargo build --no-default-features --features="pe32"` in my own Ubuntu 20.04 and no error is turned out.

philipc:
Are you trying to reproduce the error with rust 1.36? That error is in the build for the Minimum Supported Rust Version (MSRV).

Evian-Zhang:
OK, I have re-write some code and managed to pass the CI. (The rust mirror I use does not have the so-long-ago rust version, so I can't test it in my own computer.)

src/pe/mod.rs
```diff
@@ -176,6 +176,26 @@ impl<'a> PE<'a> {
             exception_data,
         })
     }
+
+    /// Validate the `PE`
+    pub fn validate(&self) -> error::Result<()> {
```
m4b, line 181: This isn't turned on by default, right? I.e., it's a method the user can (optionally) call after to do validation after the parse occurs, and PE object created (assuming it didn't fail from more serious errors during initial parse), yes?

src/pe/optional_header.rs
```diff
@@ -234,6 +235,40 @@ impl From<WindowsFields32> for WindowsFields {
...
+                .push("FileAlignment should be a power of 2 between 512 and 64K.".to_owned());
+        }
+        if self.size_of_image % self.section_alignment != 0 {
+            error_messages.push("SizeOfImage must be a multiple of SectionAlignment.".to_owned());
+        }
+        if error_messages.is_empty() {
+            Ok(())
+        } else {
+            Err(error::Error::Malformed(
+                super::utils::organize_validation_error_messages("", error_messages),
```
m4b, line 266: nit: This was bit confusing, seeing the `""` in most calls, but i see now it's a prefix. Not sure if `Option<&str>` is better? Also since it's `pub(crate)` doesn't matter that much I suppose, since it's more of an internal function to PE. Perhaps `pub(super)` is even better, I can't remember if this would work w.r.t. privacy requirements though?

src/pe/header.rs
```diff
@@ -149,6 +161,41 @@ impl CoffHeader {
         let length = bytes.pread_with::<u32>(offset, scroll::LE)? as usize;
         Ok(strtab::Strtab::parse(bytes, offset, length, 0)?)
     }
+
+    /// Validate the `CoffHeader`.
+    ///
+    /// `is_image` being `true` indicates the `CoffHeader` is in a `PE`,
+    /// being `false` indicates the `CoffHeader` is in a `Coff`.
+    pub fn validate(&self, is_image: bool) -> error::Result<()> {
```
m4b, line 169: this suffers from boolean blindness, and is particularly difficult because there's nothing really boolean about how it affects the semantics of validation here.  Not sure what the best solution is, but because it's a (new) pub method, would like to think about this better.

src/pe/characteristic.rs
```diff
@@ -90,10 +92,64 @@ pub const IMAGE_FILE_DLL: u16 = 0x2000;
...
 pub fn is_exe(characteristics: u16) -> bool {
-    characteristics & IMAGE_FILE_EXECUTABLE_IMAGE == IMAGE_FILE_EXECUTABLE_IMAGE
+    has_flag(characteristics, IMAGE_FILE_EXECUTABLE_IMAGE)
+}
+
+/// Validate the characteristics.
+///
+/// `is_image` being `true` indicates the characteristics is in a `PE`,
+/// being `false` indicates the characteristics is in a `Coff`.
+pub fn validate(characteristics: u16, is_image: bool) -> error::Result<()> {
```
m4b, line 111: ditto same issue with boolean blindness here.

Also does this really need to be `pub`; I don't see many people calling this.

src/pe/section_table.rs
```diff
@@ -163,6 +164,57 @@ impl SectionTable {
         let number = self.number_of_relocations as usize;
         relocation::Relocations::parse(bytes, offset, number)
     }
+
+    /// Validate the `SectionTable`.
+    ///
+    /// `is_image` being `true` indicates the `SectionTable` is in a `PE`,
+    /// being `false` indicates the `SectionTable` is in a `Coff`.
+    pub fn validate(&self, is_image: bool) -> error::Result<()> {
```
m4b, line 172: ditto boolean blindness concern

src/pe/optional_header.rs
```diff
@@ -249,6 +284,24 @@ impl OptionalHeader {
             magic => Err(error::Error::BadMagic(u64::from(magic))),
         }
     }
+
+    /// Validate the Optional Header.
+    pub fn validate(&self) -> error::Result<()> {
+        let mut error_messages = vec![];
+        if let Err(error::Error::Malformed(error_message)) = self.windows_fields.validate() {
```
not-wlan, line 291: While you're here you might as well validate IMAGE_OPTIONAL_HEADER.Magic == IMAGE_NT_OPTIONAL_HDR_MAGIC

https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-image_optional_header32

src/pe/optional_header.rs
```diff
@@ -234,6 +235,40 @@ impl From<WindowsFields32> for WindowsFields {
...
 
+impl WindowsFields {
+    /// Validate the Windows-Specific Fields
+    ///
+    /// This method DOES NOT validate `check_sum` field since
+    /// the algorithm for computing the checksum is incorporated
+    /// into IMAGEHELP.DLL, which is os-dependent.
+    pub fn validate(&self) -> error::Result<()> {
+        let mut error_messages = vec![];
+        if self.image_base % 0x1000 != 0 {
```
not-wlan, line 246: Shouldn't this be 0x1000**0** to reflect the 64KiB requirement?

src/pe/optional_header.rs
```diff
@@ -234,6 +235,40 @@ impl From<WindowsFields32> for WindowsFields {
...
+            error_messages.push("ImageBase must be a multiple of 64K.".to_owned());
+        }
+        if self.section_alignment < self.file_alignment {
+            error_messages.push(
+                "SectionAlignment must be greater than or equal to FileAlignment.".to_owned(),
+            );
+        }
+        if self.file_alignment < 512 || self.file_alignment > 0x1000 || self.file_alignment % 2 != 0
```
not-wlan, line 254: This check validates e.g. 514 despite not being a power of two. In addition to that it should also be 0x1000**0** instead of 0x1000.

It should be `if self.file_alignment < 0x200 || self.file_alignment > 0x10000 || (self.file_alignment & (self.file_alignment - 1)) == 0` to validate all requirements.

m4b:
closing this for now as no review comments were addressed; feel free to reopen and rebase, address comments if desired, and thank you for the PR!

PR #237, opened 2020-10-06
elf.parse: resolve overflow issue in hash sections

quake:
I found an overflow error when running the fuzz test ([test case](https://github.com/quake/goblin/commit/5dbbf95e130aa38aced419f0a3c4dd1676530ded)), this PR try to fix the issue by using `wrapping` operator.

### DIFF
--- a/src/elf/mod.rs
+++ b/src/elf/mod.rs
@@ -348,18 +348,18 @@ if_sylvan! {
 
     fn gnu_hash_len(bytes: &[u8], offset: usize, ctx: Ctx) -> error::Result<usize> {
         let buckets_num = bytes.pread_with::<u32>(offset, ctx.le)? as usize;
-        let min_chain = bytes.pread_with::<u32>(offset + 4, ctx.le)? as usize;
-        let bloom_size = bytes.pread_with::<u32>(offset + 8, ctx.le)? as usize;
+        let min_chain = bytes.pread_with::<u32>(offset.wrapping_add(4), ctx.le)? as usize;
+        let bloom_size = bytes.pread_with::<u32>(offset.wrapping_add(8), ctx.le)? as usize;
         // We could handle min_chain==0 if we really had to, but it shouldn't happen.
         if buckets_num == 0 || min_chain == 0 || bloom_size == 0 {
             return Err(error::Error::Malformed(format!("Invalid DT_GNU_HASH: buckets_num={} min_chain={} bloom_size={}",
                                                        buckets_num, min_chain, bloom_size)));
         }
         // Find the last bucket.
-        let buckets_offset = offset + 16 + bloom_size * if ctx.container.is_big() { 8 } else { 4 };
+        let buckets_offset = offset.wrapping_add(16).wrapping_add(bloom_size.wrapping_mul(if ctx.container.is_big() { 8 } else { 4 }));
         let mut max_chain = 0;
         for bucket in 0..buckets_num {
-            let chain = bytes.pread_with::<u32>(buckets_offset + bucket * 4, ctx.le)? as usize;
+            let chain = bytes.pread_with::<u32>(buckets_offset.wrapping_add(bucket.wrapping_mul(4)), ctx.le)? as usize;
             if max_chain < chain {
                 max_chain = chain;
             }
@@ -368,7 +368,7 @@ if_sylvan! {
             return Ok(0);
         }
         // Find the last chain within the bucket.
-        let mut chain_offset = buckets_offset + buckets_num * 4 + (max_chain - min_chain) * 4;
+        let mut chain_offset = buckets_offset.wrapping_add(buckets_num.wrapping_mul(4)).wrapping_add((max_chain - min_chain).wrapping_mul(4));
         loop {
             let hash = bytes.pread_with::<u32>(chain_offset, ctx.le)?;
             max_chain += 1;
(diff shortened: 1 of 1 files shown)


m4b:
Hmmm looks like macos CI is failing (for likely unrelated reasons ??)

quake:
weird, trying revert commit and trigger the CI again

quake:
> Hmmm looks like macos CI is failing (for likely unrelated reasons ??)

I'm sure the CI error has nothing to do with this commit, I reverted the commit and got same error.

m4b:
it looks like `/Library/Developer/CommandLineTools/usr/bin/dyldinfo` doesn't exist anymore, so that's the failure

willglynn:
Ah, I believe that's moved from `…/dyldinfo args` to `/usr/bin/xcrun dyldinfo args`.

m4b:
Oh you sweet, sweet cherub @willglynn , swooping in and saving the day; @quake this is annoying, but would you mind changing https://github.com/m4b/goblin/blob/e94ca6b81b3b3a87ed5efc94493b790fd1085355/tests/compare_dyldinfos.rs#L4 to `xcrun dyldinfo` ?

m4b:
actually it may be more complicated, maybe a separate PR is better

quake:
I will create a separate PR to try `xcrun`, but I suspect that's not the causing, since test output reports:
```
cargo dyldinfo ["-lazy_bind", "-bind", "/Library/Developer/CommandLineTools/usr/bin/dyldinfo"] output:
err: BadMagic(3405691582)
```
and `/Library/Developer/CommandLineTools/usr/bin/dyldinfo` runs well.

it looks like the magic header of `dyldinfo` is 0xcafebabe (3405691582), and it's ignored in `parse_magic_and_ctx` fn.

m4b:
Oh wow you're right; fwiw, it should probably be updated to not hard code anyway, since it can fail on other machines, something perhaps like:
```diff
diff --git a/tests/compare_dyldinfos.rs b/tests/compare_dyldinfos.rs
index 0c9dcaf..806e004 100644
--- a/tests/compare_dyldinfos.rs
+++ b/tests/compare_dyldinfos.rs
@@ -1,7 +1,19 @@
 use std::process;
 
+fn get_realpath(cmd: &str) -> String {
+    let output = process::Command::new("/usr/bin/xcrun")
+        .arg("-f")
+        .arg(cmd)
+        .output()
+        .expect("can get realpath");
+    std::str::from_utf8(&output.stdout)
+        .expect("output is valid utf8")
+        .to_string()
+}
+
 pub fn compare(args: Vec<&str>) {
-    let apple = process::Command::new("/Library/Developer/CommandLineTools/usr/bin/dyldinfo")
+    let apple = process::Command::new("/usr/bin/xcrun")
+        .arg("dyldinfo")
         .args(&args)
         .output()
         .expect("run Apple dyldinfo");
@@ -39,39 +51,28 @@ pub fn compare(args: Vec<&str>) {
 #[cfg(target_os = "macos")]
 #[test]
 fn compare_binds() {
-    compare(vec![
-        "-bind",
-        "/Library/Developer/CommandLineTools/usr/bin/dyldinfo",
-    ]);
-    compare(vec![
-        "-bind",
-        "/Library/Developer/CommandLineTools/usr/bin/clang",
-    ]);
+    let dyldinfo = get_realpath("dyldinfo");
+    let clang = get_realpath("dyldinfo");
+    compare(vec!["-bind", &dyldinfo]);
+    compare(vec!["-bind", &clang]);
     compare(vec!["-bind", "/usr/bin/tmutil"]);
 }
 
 #[cfg(target_os = "macos")]
 #[test]
 fn compare_lazy_binds() {
-    compare(vec![
-        "-lazy_bind",
-        "/Library/Developer/CommandLineTools/usr/bin/dyldinfo",
-    ]);
-    compare(vec![
-        "-lazy_bind",
-        "/Library/Developer/CommandLineTools/usr/bin/clang",
-    ]);
+    let dyldinfo = get_realpath("dyldinfo");
+    let clang = get_realpath("dyldinfo");
+    compare(vec!["-lazy_bind", &dyldinfo]);
+    compare(vec!["-lazy_bind", &clang]);
     compare(vec!["-lazy_bind", "/usr/bin/tmutil"]);
 }
 
 #[cfg(target_os = "macos")]
 #[test]
 fn compare_combined_options() {
-    compare(vec![
-        "-lazy_bind",
-        "-bind",
-        "/Library/Developer/CommandLineTools/usr/bin/dyldinfo",
-    ]);
+    let dyldinfo = get_realpath("dyldinfo");
+    compare(vec!["-lazy_bind", "-bind", &dyldinfo]);
 }
 
 #[cfg(not(target_os = "macos"))]
```

src/elf/mod.rs
```diff
@@ -348,18 +348,18 @@ if_sylvan! {
 
     fn gnu_hash_len(bytes: &[u8], offset: usize, ctx: Ctx) -> error::Result<usize> {
         let buckets_num = bytes.pread_with::<u32>(offset, ctx.le)? as usize;
-        let min_chain = bytes.pread_with::<u32>(offset + 4, ctx.le)? as usize;
-        let bloom_size = bytes.pread_with::<u32>(offset + 8, ctx.le)? as usize;
+        let min_chain = bytes.pread_with::<u32>(offset.wrapping_add(4), ctx.le)? as usize;
```
m4b, line 351: ok back to this PR; so what are the semantics if this overflows? This seems like a subtle bug now (I guess it was always present) Also, won't this only crash in debug mode, but will wrap overflow in release? Ditto for the other explicit overflow wraps

m4b:
@lzutao perhaps you have some thoughts on behavior of overflow wrapping, you did the work on the gnu hash rewrite

src/elf/mod.rs
quake: I'm using `wrapping_xxx` here because I've seen other code do it too, for example, 
https://github.com/m4b/goblin/blob/2188699dcc242bae0e58b052b029521b4c8a100a/src/mach/imports.rs#L234, semantically, maybe `overflowing_xxx` is better?

m4b: yes, i think your intuition is right, was kind of just wondering out loud what pros/cons are of the explict wrap, and what it means for the gnu hash table calculations to overflow (i just assume it means the binary is bad, better not to crash then i suppose?)

lzutao: Does `gnu_hash_len` calculate total size of GNU_HASH table?
Then performing some checks like https://github.com/m4b/goblin/blob/ece3ce60fbe5e7e6f31c69fd462d1b37521b0dc8/src/elf/gnu_hash.rs#L116-L135 is Ok.

m4b:
@quake are you still interested in getting this merged? let me know whenever you have a chance, just triaging old PRs here, thanks!

m4b:
5 year old PR probably ain't getting merged :D

PR #240, opened 2020-10-11
mach/load_command: Add some missing load commands

woodruffw:
First of all, thanks for this library! It's a pleasure to use.

I'm the maintainer of [ruby-macho](https://github.com/Homebrew/ruby-macho), and I noticed that these load commands were missing from your Mach-O implementation. I've taken the liberty of adding them; please let me know if there's anything else I can do.

---

Commit message:

Adds support for `LC_VERSION_MIN_{TV,WATCH}OS`, as well as
`LC_VERSION_EXPORTS_TRIE` and `LC_VERSION_CHAINED_FIXUPS`.
Each of the above maps to a pre-existing load command structure,
so no additional structure definitions were necessary.

Also adds constant support for `LC_NOTE` and `LC_BUILD_VERSION`.
These require new load command structures (and substructures, in the
case of `LC_BUILD_VERSION`), so I left them out of this PR in
the interest of brevity. I can do them in a follow-up PR,
if desired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ https://docs.rs/goblin/
 
 ### Usage
 
-Goblin requires `rustc` 1.36.0.
+Goblin requires `rustc` 1.40.0.
 
 Add to your `Cargo.toml`
 
--- a/src/mach/load_command.rs
+++ b/src/mach/load_command.rs
@@ -946,7 +946,8 @@ pub const SIZEOF_RPATH_COMMAND: usize = 12;
 #[repr(C)]
 #[derive(Debug, Clone, Copy, Pread, Pwrite, IOread, IOwrite, SizeWith)]
 pub struct LinkeditDataCommand {
-    /// LC_CODE_SIGNATURE, LC_SEGMENT_SPLIT_INFO, LC_FUNCTION_STARTS, LC_DATA_IN_CODE, LC_DYLIB_CODE_SIGN_DRS or LC_LINKER_OPTIMIZATION_HINT.
+    /// LC_CODE_SIGNATURE, LC_SEGMENT_SPLIT_INFO, LC_FUNCTION_STARTS, LC_DATA_IN_CODE,
+    /// LC_DYLIB_CODE_SIGN_DRS, LC_LINKER_OPTIMIZATION_HINT, LC_DYLD_EXPORTS_TRIE, or LC_DYLD_CHAINED_FIXUPS.
     pub cmd: u32,
     /// sizeof(struct linkedit_data_command)
     pub cmdsize: u32,
@@ -1001,10 +1002,10 @@ pub const SIZEOF_ENCRYPTION_INFO_COMMAND_64: usize = 24;
 /// The version_min_command contains the min OS version on which this
 /// binary was built to run.
 ///
-/// LC_VERSION_MIN_MACOSX or LC_VERSION_MIN_IPHONEOS
 #[repr(C)]
 #[derive(Debug, Clone, Copy, Pread, Pwrite, IOread, IOwrite, SizeWith)]
 pub struct VersionMinCommand {
+    /// LC_VERSION_MIN_MACOSX, LC_VERSION_MIN_IPHONEOS, LC_VERSION_MIN_TVOS, or LC_VERSION_MIN_WATCHOS.
     pub cmd: u32,
     pub cmdsize: u32,
     /// X.Y.Z is encoded in nibbles xxxx.yy.zz
@@ -1014,6 +1015,7 @@ pub struct VersionMinCommand {
 }
 
 impl VersionMinCommand {
+    // TODO: Needs a discriminant for TvOS, WatchOS.
     pub fn new(is_ios: bool) -> Self {
         VersionMinCommand {
             cmd: if is_ios {
@@ -1184,6 +1186,8 @@ pub const LC_REEXPORT_DYLIB: u32 = 0x1f | LC_REQ_DYLD;
 pub const LC_DYLD_INFO_ONLY: u32 = 0x22 | LC_REQ_DYLD;
 pub const LC_LOAD_UPWARD_DYLIB: u32 = 0x23 | LC_REQ_DYLD;
 pub const LC_MAIN: u32 = 0x28 | LC_REQ_DYLD;
+pub const LC_DYLD_EXPORTS_TRIE: u32 = 0x33 | LC_REQ_DYLD;
+pub const LC_DYLD_CHAINED_FIXUPS: u32 = 0x34 | LC_REQ_DYLD;
 pub const LC_SEGMENT: u32 = 0x1;
 pub const LC_SYMTAB: u32 = 0x2;
 pub const LC_SYMSEG: u32 = 0x3;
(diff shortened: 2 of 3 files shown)


src/mach/load_command.rs
```diff
@@ -1014,6 +1015,7 @@ pub struct VersionMinCommand {
 }
 
 impl VersionMinCommand {
+    // TODO: Needs a discriminant for TvOS, WatchOS.
```
willglynn, line 1018: Yeah, we ought to break this interface. Maybe add a new `#[non_exhaustive] enum Platform` and add a `fn platform(&self) -> Platform` while we're at it.

willglynn (approved):
LGTM, thanks! I would welcome followup PRs that bring `goblin` back to parity with current `#include <mach-o/*.h>`.

src/mach/load_command.rs
woodruffw: SGTM! I can make that part of this PR.

woodruffw:
Okay, I've added a `Platform` enum and the appropriate conversion traits. One slightly messy API detail: `platform()` currently returns a `Result<Platform>` instead of a `Platform` directly, since `u32 -> Platform` is implemented with the lossy `TryFrom`.

It's technically safe to remove that `Result` with an `unwrap` (since the only ways to create a `VersionMinCommand` ensure that `self.cmd` is always valid), but it's difficult to prove that at the type level with the current API. Thoughts? I'm happy to go either way :slightly_smiling_face:

woodruffw:
Also, TIL that `non_exhaustive` was only stabilized in `1.40`. That means bumping the MSRV from the current (`1.36`), or just dropping that attribute. Again, either's fine by me, I figure this is a question for the maintainers :slightly_smiling_face:

m4b:
Going to review in a bit, thanks for this! So this is a breaking change right (which is fine, might as well do this now)?

woodruffw:
> So this is a breaking change right (which is fine, might as well do this now)?

Yep, correct -- this changes the public API for `VersionMinCommand::new`.

src/mach/load_command.rs
```diff
@@ -1014,18 +1056,18 @@ pub struct VersionMinCommand {
...
 impl VersionMinCommand {
-    pub fn new(is_ios: bool) -> Self {
+    pub fn new(platform: Platform) -> Self {
         VersionMinCommand {
-            cmd: if is_ios {
-                LC_VERSION_MIN_IPHONEOS
-            } else {
-                LC_VERSION_MIN_MACOSX
-            },
+            cmd: platform as u32,
```
m4b, line 1061: Is this a new feature ? How is this cast working since Platform isn’t repr(u32).

src/mach/load_command.rs
```diff
@@ -1596,6 +1672,10 @@ impl CommandVariant {
             DylibCodeSignDrs(comm) => comm.cmdsize,
             LinkerOption(comm) => comm.cmdsize,
             LinkerOptimizationHint(comm) => comm.cmdsize,
+            VersionMinTvos(comm) => comm.cmdsize,
+            VersionMinWatchos(comm) => comm.cmdsize,
+            DyldExportsTrie(comm) => comm.cmdsize,
+            DyldChainedFixups(comm) => comm.cmdsize,
```
m4b, line 1678: We don’t need to do this now but might be a good idea to add non exhaustive annotations to these enums as much well

src/mach/load_command.rs
m4b, line 1061: Is this a new feature ? How is this as cast working since Platform isn’t repr(u32)

src/mach/load_command.rs
```diff
@@ -998,13 +1000,53 @@ pub struct EncryptionInfoCommand64 {
 
 pub const SIZEOF_ENCRYPTION_INFO_COMMAND_64: usize = 24;
 
+/// An enumeration of platforms currently identifiable within a version_min_command.
+#[non_exhaustive]
+#[derive(Debug)]
+pub enum Platform {
```
m4b, line 1006: If these are tied to the LC values why not make it repr(u32) and set the values to the constants ? Maybe you can’t do that with constants in enums can’t remember

README.md
```diff
@@ -20,7 +20,7 @@ https://docs.rs/goblin/
 
 ### Usage
 
-Goblin requires `rustc` 1.36.0.
+Goblin requires `rustc` 1.40.0.
```
m4b, line 23: I don’t really like this but it’s fine I guess.

src/mach/load_command.rs
```diff
@@ -1540,7 +1598,25 @@ impl<'a> ctx::TryFromCtx<'a, Endian> for CommandVariant {
...
+            LC_VERSION_MIN_WATCHOS => {
+                let comm = bytes.pread_with::<VersionMinCommand>(0, le)?;
+                Ok((VersionMinWatchos(comm), size))
+            }
+            LC_DYLD_EXPORTS_TRIE => {
+                let comm = bytes.pread_with::<LinkeditDataCommand>(0, le)?;
+                Ok((DyldExportsTrie(comm), size))
+            }
+            LC_DYLD_CHAINED_FIXUPS => {
+                let comm = bytes.pread_with::<LinkeditDataCommand>(0, le)?;
```
m4b, line 1614: This and above are both LinkEditDataCommand? Interesting

m4b (approved):
Only real question I have is whether the Platform should be repr(u32) (doesn’t have to be maybe there’s reasons to keep it away from the backing u32 consts?)

m4b:
I’m ok with bumping msrv. Re The unwrapping issue, id day if it really is safe to unwrap then do it (but use expect); but it’s the possible, Eg with a malformed binary, etc, to have that unwrap fail then leave it as is :)

src/mach/load_command.rs
woodruffw: Yep, this is part of the API breakage -- `VersionMinCommand` can be produced from one of four different `LC_*` constants, which are represented by that enum (since they might be useful elsewhere).

As for why it's working: I *thought* it was because I had implemented `From<Platform> for u32`, but I tried removing that trait `impl` and it still compiled :worried:. So I'll change it to `platform.into()` to make sure it's using the trait explicitly.

src/mach/load_command.rs
woodruffw: Yep, they both reside in `__linkedit`, like a few others.

src/mach/load_command.rs
woodruffw: I thought you couldn't either, but I just tried and it works :shrug:. So I'll remove the trait usage in favour of `repr(u32)`.

woodruffw:
> Re The unwrapping issue, id day if it really is safe to unwrap then do it (but use expect); but if it’s the possible, e.g. with a malformed binary, etc, to have that unwrap fail, then leave it as is :)

Sounds good! It should always be safe since the only ways to construct a `VersionMinCommand` constrain `cmd` to the variants we expect, so I'll use `expect`.

m4b (approved):
Awesome, thank you for this PR! :heart:

m4b:
@woodruffw I hope you don't mind, but I squashed and merged; i find this makes PR's cleaner generally, easier to bisect, known passing commit, etc. It's not a hard and fast rule but I prefer it more and more. People do get less "commit credit" so i do give them a heads up so they don't get sad :D

woodruffw:
I don’t mind at all! I usually do the same. Thanks for the quick review and merge!

m4b:
ok this is wrapped up in upcoming 0.3 release

PR #243, opened 2020-11-16
elf: prevent overflow

jackcmay:
Fuzzier found an overflow if bad section address or size.  

- Update `check_size` to also check addr + size
- Use saturating addition to prevent overflow or a wrap-around region bound if `check_size` is not called on a created section.

### DIFF
--- a/src/elf/section_header.rs
+++ b/src/elf/section_header.rs
@@ -428,11 +428,11 @@ if_alloc! {
         }
         /// Returns this section header's file offset range
         pub fn file_range(&self) -> Range<usize> {
-            self.sh_offset as usize..self.sh_offset as usize + self.sh_size as usize
+            self.sh_offset as usize..(self.sh_offset as usize).saturating_add(self.sh_size as usize)
         }
         /// Returns this section header's virtual memory range
         pub fn vm_range(&self) -> Range<usize> {
-            self.sh_addr as usize..self.sh_addr as usize + self.sh_size as usize
+            self.sh_addr as usize..(self.sh_addr as usize).saturating_add(self.sh_size as usize)
         }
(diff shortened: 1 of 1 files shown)


jackcmay:
Looks like CI is failing on something unrelated to my changes :-(

m4b:
Yes looks like it's this: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

aside: i find it a bit sad that for nearly 4 years travis CI did not have a single CI involved regression with any of the github projects i had, and now, with **not even 1 year into github actions**, it's already breaking my code, and wasting developer time upgrading to some new api... This does not bode well.

jackcmay:
:-(. Should I rebase once you have applied a fix?

m4b:
I don't have the time right now to fix the issue; maybe this weekend; i can just merge without CI if you want

m4b (approved):
LGTM thanks for this!

m4b:
If you think you know how to fix, feel free to go for it in this PR, otherwise we can just merge

m4b:
Ok i think i fixed it actually here: https://github.com/m4b/goblin/pull/244

can you rebase on top of that ?

jackcmay:
Rebased and CI passed, woot!

m4b:
sorrrryyyyy for spamming :D and thanks for the PR :heart:

jackcmay:
@m4b When do you think these changes will be released so we can pick them up?

m4b:
whoops, forgot about this, will release now!

m4b:
ok this is released in 0.3, which rolls up several things from past few months; it's a breaking change because one part of the mach-o api changed, but otherwise it should be fine

PR #247, opened 2021-01-03
Fix tests for multi-arch macOS binaries

keith:
As of Big Sur, even on Intel macOS machines, developer tools contain
both x86_64 and arm64 slices. This updates the dyldinfo example / tests
to only output the binds for x86_64.

### DIFF
--- a/tests/compare_dyldinfos.rs
+++ b/tests/compare_dyldinfos.rs
@@ -12,6 +12,8 @@ fn get_realpath(cmd: &str) -> String {
 pub fn compare(args: Vec<&str>) {
     let apple = process::Command::new("/usr/bin/xcrun")
         .arg("dyldinfo")
+        .arg("-arch")
+        .arg("x86_64")
         .args(&args)
         .output()
         .expect("run Apple dyldinfo");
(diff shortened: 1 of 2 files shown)


tests/compare_dyldinfos.rs
```diff
@@ -12,6 +12,8 @@ fn get_realpath(cmd: &str) -> String {
 pub fn compare(args: Vec<&str>) {
     let apple = process::Command::new("/usr/bin/xcrun")
         .arg("dyldinfo")
+        .arg("-arch")
+        .arg("x86_64")
```
m4b, line 16: This won’t run on a Mojave or Catalina system right ? There won’t be an arch flag ? (Just wondering if the example will work on non Big Sur systems or only Big Sur systems)

m4b:
This is great thanks for updating the example ! Just had the one question otherwise LGTM, thanks for doing this :)

tests/compare_dyldinfos.rs
keith: It looks like `-arch` was a flag back in Xcode 11.3 https://github.com/keith/xcode-man-pages/blob/db47aee1c9b7493f6af64bfdf467b68706ce60b8/docs/dyldinfo.1.html so I think it would still work on those machines. Which makes sense to me given you would have needed it for i386 + x86_64 on macOS as well. Unless you're thinking that passing this would fail on non multi-arch binaries? (quick test on big sur still on a binary with just x86_64 works as expected)

m4b: Awesome thank you for checking !

m4b (approved):
LGTM thank you for this !

PR #248, opened 2021-01-03
Add rpaths to MachO

keith:


keith:
I'm not super familiar with rust but when I was trying to load these from the public API I couldn't find a way to without access to `bytes`. Maybe I'm missing something? Otherwise I think everything using `LcStr` would have to have some public API like this? It doesn't seem like the other load commands, besides those stored in `libs` have any callers so I'm not sure.

m4b:
@keith thanks for the PR! This looks good to me, I’m surprised the list of rpaths hasn’t been added yet (or complained about as missing before)!

re your question about LcStr and load commands I didn’t follow exactly

m4b (approved):

keith:
I mean that I think all the load commands that have `LcStr`s have to have some convenience API like this currently, since otherwise there's no way to fetch that value. Looks to me like that's mostly handled by `libs` though

PR #249, opened 2021-01-18
Fix size check for section headers

jan-auer:
The section header size check introduced in #243 compares the virtual memory
address range `sh_addr` with the physical file size. This leads to a
false-positive invariant violation and fails to parse valid ELF files.

See also https://github.com/m4b/bingrep/issues/28

cc @jackmay


m4b (approved):
LGTM, thanks for this!

jan-auer:
Super quick turnaround, thanks a bunch @m4b!

PR #251, opened 2021-01-20
Fix MultiArch Debug impl panic

### LINKED ISSUE #250: MultiArch parsing non fat binary panicked with Scroll(BadOffset())
Ideally it should return an `Err` instead of panicking?

```toml
[dependencies]
goblin = "0.3.1"
```

```rust
use goblin::mach::MultiArch;

fn main() {
    let fat = MultiArch::new(include_bytes!("thin_x86_64")).unwrap();   //  this returns Ok
    println!("{:?}", fat);  // panicked here
}
```

`thin_x86_64` file  [thin_x86_64.zip](https://github.com/m4b/goblin/files/5840377/thin_x86_64.zip)

```bash
$ cargo run
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Scroll(BadOffset(8))', goblin-0.3.1/src/mach/mod.rs:413:45
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
MultiArch
```

https://github.com/m4b/goblin/blob/8d91dd76a1950882301481d92f053e8649faaae7/src/mach/mod.rs#L413

messense:
Fixes #250

### DIFF
--- a/src/mach/mod.rs
+++ b/src/mach/mod.rs
@@ -410,7 +410,7 @@ impl<'a> MultiArch<'a> {
 impl<'a> fmt::Debug for MultiArch<'a> {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
         fmt.debug_struct("MultiArch")
-            .field("arches", &self.arches().unwrap())
+            .field("arches", &self.arches().unwrap_or_default())
             .field("data", &self.data.len())
             .finish()
     }


src/mach/mod.rs
```diff
@@ -410,7 +410,7 @@ impl<'a> MultiArch<'a> {
 impl<'a> fmt::Debug for MultiArch<'a> {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
         fmt.debug_struct("MultiArch")
-            .field("arches", &self.arches().unwrap())
+            .field("arches", &self.arches().unwrap_or_default())
```
messense, line 413: It requires a `Vec<FatArch>`, can't write it to output `NO_ARCHES`

m4b:
argh not sure why i didn't see this, sorry for the delay and thanks for the fix!

PR #253, opened 2021-01-24
Return an empty file range for SHT_NOBITS sections

### LINKED ISSUE #252: Incorrect file_range for ELF sections with type SHT_NOBITS
I've noticed that goblin returns a non-empty range when `file_range()` is called on a section header with type `SHT_NOBITS`.

From what I understand, these sections don't have any data in the file itself, and only exist in memory. I would expect that the returned range is empty in that case. The returned file range also overlaps with the file range of the next section.

I would suggest that the `file_range` function should check the type, and return an empty range in case its NOBITS. If that's ok I can do a PR.

Tiwalun:
ELF sections with type SHT_NOBITS only exist in memory, and
don't contain any data in the ELF file itself. The range
returned by `file_range()` should be empty for such sections.

The previously returned range was wrong, and overlapped
the file range of the next section.

Fixes #252.

### DIFF
--- a/src/elf/section_header.rs
+++ b/src/elf/section_header.rs
@@ -428,7 +428,13 @@ if_alloc! {
         }
         /// Returns this section header's file offset range
         pub fn file_range(&self) -> Range<usize> {
-            self.sh_offset as usize..(self.sh_offset as usize).saturating_add(self.sh_size as usize)
+            // Sections with type SHT_NOBITS have no data in the file itself,
+            // they only exist in memory.
+            if self.sh_type == SHT_NOBITS {
+                self.sh_offset as usize..self.sh_offset as usize
+            } else {
+                self.sh_offset as usize..(self.sh_offset as usize).saturating_add(self.sh_size as usize)
+            }
         }
         /// Returns this section header's virtual memory range
         pub fn vm_range(&self) -> Range<usize> {
--- a/src/elf/section_header.rs
+++ b/src/elf/section_header.rs
@@ -426,14 +426,15 @@ if_alloc! {
                 sh_entsize: 0,
             }
         }
-        /// Returns this section header's file offset range
-        pub fn file_range(&self) -> Range<usize> {
+        /// Returns this section header's file offset range,
+        /// if the section occupies space in fhe file.
+        pub fn file_range(&self) -> Option<Range<usize>> {
             // Sections with type SHT_NOBITS have no data in the file itself,
             // they only exist in memory.
             if self.sh_type == SHT_NOBITS {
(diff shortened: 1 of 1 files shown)


src/elf/section_header.rs
```diff
@@ -428,7 +428,13 @@ if_alloc! {
         }
         /// Returns this section header's file offset range
         pub fn file_range(&self) -> Range<usize> {
-            self.sh_offset as usize..(self.sh_offset as usize).saturating_add(self.sh_size as usize)
+            // Sections with type SHT_NOBITS have no data in the file itself,
+            // they only exist in memory.
+            if self.sh_type == SHT_NOBITS {
+                self.sh_offset as usize..self.sh_offset as usize
```
m4b, line 434: is returning a 0 range going to cause issues? ideally we'd return an `Option<Range<usize>>` i think, but that would be breaking change. Might be worth it though?

m4b: ok i don't think this should be `self.sh_offset..self.sh_offset`; if sh_offset is non-zero you can get e.g., `10..10` or whatever, that is an empty range but will cause a panic if it's larger than buffer, i.e., this panics:
```rust
    let foo = [1, 2, 3];
    let x = &foo[10..10];
```

So i think it's better to return `0..0` ? or Option.

Tiwalun: Oh, I didn't expect the empty range to panic when it's out of bounds.

I think `0..0`would be confusing as well, if someone looks at the start field of the range to decide where the data in the file starts.

An Option would be the solution, in my opinion.

m4b: Ok let's make it option, it's the right thing to do :) this will be a breaking change so unless you are in urgent need of a release, i may wait a bit to roll up other potential changes, or make other breaking changes

Tiwalun: I don't need it urgently, manually checking the section type is an easy workaround.

src/elf/section_header.rs
```diff
@@ -426,9 +426,16 @@ if_alloc! {
                 sh_entsize: 0,
             }
         }
-        /// Returns this section header's file offset range
-        pub fn file_range(&self) -> Range<usize> {
-            self.sh_offset as usize..(self.sh_offset as usize).saturating_add(self.sh_size as usize)
+        /// Returns this section header's file offset range,
+        /// if the section occupies space in fhe file.
```
m4b, line 430: typo: `fhe` -> `the`

m4b (approved):
if you want to fix the typo that would be great, otherwise LGTM

m4b:
thanks for this PR ! :)

PR #254, opened 2021-01-26
elf.parse: added lazy_parse function

jessehui:
We are developing a libOS kernel (https://github.com/occlum/occlum) and would like to load ELF as lazy as possible. However, now goblin does all the read and parse in memory so we must read the whole ELF file to kernel memory (we can't mmap the file to memory due to some constraint).

Thus, I came up with this method to load lazily:
read ELF header only -> parse header -> get program header table -> only read in PT_LOAD and PT_INTERP segments

And by adding these two functions, I can achieve this. The implementation can be found [here](https://github.com/occlum/occlum/pull/315). I would like to see this get merged to master so that people with similar requirements can do this. But I am not sure if it is a common need. If it is, I can spend some time to reduce the repeated code.

Comments are welcome. Thank you.

### DIFF
--- a/src/elf/mod.rs
+++ b/src/elf/mod.rs
@@ -202,22 +202,46 @@ if_sylvan! {
         pub fn is_object_file(&self) -> bool {
             self.header.e_type == header::ET_REL
         }
+
+        /// Parses the contents to get the Header only. This `bytes` buffer should contain at least the length for parsing Header.
+        pub fn parse_header(bytes: &'a [u8]) -> error::Result<Header> {
+            bytes.pread::<Header>(0)
+        }
+
+        /// Lazy parse the ELF contents. This function mainly just assembles an Elf struct. Once we have the struct, we can choose to parse whatever we want.
+        pub fn lazy_parse(header: Header) -> error::Result<Self> {
+            let misc = parse_misc(&header)?;
+
+            Ok(Elf {
+                header,
+                program_headers: vec![],
+                section_headers: Default::default(),
+                shdr_strtab: Default::default(),
+                dynamic: None,
+                dynsyms: Default::default(),
+                dynstrtab: Strtab::default(),
+                syms: Default::default(),
+                strtab: Default::default(),
+                dynrelas: Default::default(),
+                dynrels: Default::default(),
+                pltrelocs: Default::default(),
+                shdr_relocs: Default::default(),
+                soname: None,
+                interpreter: None,
+                libraries: vec![],
+                is_64: misc.is_64,
+                is_lib: misc.is_lib,
+                entry: misc.entry,
+                little_endian: misc.little_endian,
+                ctx: misc.ctx,
+            })
+        }
+
         /// Parses the contents of the byte stream in `bytes`, and maybe returns a unified binary
         pub fn parse(bytes: &'a [u8]) -> error::Result<Self> {
-            let header = bytes.pread::<Header>(0)?;
-            let entry = header.e_entry as usize;
-            let is_lib = header.e_type == header::ET_DYN;
-            let is_lsb = header.e_ident[header::EI_DATA] == header::ELFDATA2LSB;
-            let endianness = scroll::Endian::from(is_lsb);
-            let class = header.e_ident[header::EI_CLASS];
-            if class != header::ELFCLASS64 && class != header::ELFCLASS32 {
-                return Err(error::Error::Malformed(format!("Unknown values in ELF ident header: class: {} endianness: {}",
-                                                           class,
-                                                           header.e_ident[header::EI_DATA])));
-            }
-            let is_64 = class == header::ELFCLASS64;
-            let container = if is_64 { Container::Big } else { Container::Little };
-            let ctx = Ctx::new(container, endianness);
+            let header = Self::parse_header(bytes)?;
+            let misc = parse_misc(&header)?;
+            let ctx = misc.ctx;
 
             let program_headers = ProgramHeader::parse(bytes, header.e_phoff as usize, header.e_phnum as usize, ctx)?;
 
@@ -329,11 +353,11 @@ if_sylvan! {
                 soname,
                 interpreter,
                 libraries,
-                is_64,
-                is_lib,
-                entry: entry as u64,
-                little_endian: is_lsb,
-                ctx,
+                is_64: misc.is_64,
+                is_lib: misc.is_lib,
+                entry: misc.entry,
+                little_endian: misc.little_endian,
+                ctx: ctx,
             })
         }
     }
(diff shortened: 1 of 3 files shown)


m4b:
Cool! Yea this seems like a great compromise to get a lazy parsing approach, I like it! Reducing some of the boilerplate into common function would be good, as you said, but i understand that this is just an initial sketch/PR?

What api were you thinking for filling in the rest of the structures, e.g., section headers, etc.?

And thanks for the PR! :)

jessehui:
Thank you for your reply. Glad you are interested. I push a new version to reduce the repeated code, make API more general and add a unit test which can also be an example.
  
> What api were you thinking for filling in the rest of the structures, e.g., section headers, etc.?

With goblin's well-designed APIs for each part (program headers, section headers ...), user can read in the contents they need and parse the specific part with corresponding `parse` function. And then, they can put it to the dummy `Elf` struct returned from `lazy_parse`. I think the unit test can be seen as an example ([here](https://github.com/m4b/goblin/pull/254/files#diff-f16e6a530c687929915c5d53dc772f052b5220d4ea8efe605b4aba587cf52232R37)). 
We will do the parsing similar for the libOS. Just we only care about PT_INTERP and PT_LOAD segments.

You may start the review. Thank you!

src/elf/mod.rs
```diff
@@ -202,22 +202,45 @@ if_sylvan! {
         pub fn is_object_file(&self) -> bool {
             self.header.e_type == header::ET_REL
         }
+
+        pub fn parse_elf_hdr(bytes: &'a [u8]) -> error::Result<Header> {
```
m4b, line 206: why is this a pub fn (or a function at all actually?)

src/elf/mod.rs
```diff
@@ -346,7 +369,7 @@ if_sylvan! {
         }
     }
 
-    fn gnu_hash_len(bytes: &[u8], offset: usize, ctx: Ctx) -> error::Result<usize> {
+    pub fn gnu_hash_len(bytes: &[u8], offset: usize, ctx: Ctx) -> error::Result<usize> {
```
m4b, line 372: why does this need to be `pub` now? I'd like to not increase api scope unless absolutely necessary

src/elf/mod.rs
```diff
@@ -379,7 +402,7 @@ if_sylvan! {
         }
     }
 
-    fn hash_len(bytes: &[u8], offset: usize, machine: u16, ctx: Ctx) -> error::Result<usize> {
+    pub fn hash_len(bytes: &[u8], offset: usize, machine: u16, ctx: Ctx) -> error::Result<usize> {
```
m4b, line 405: ditto

src/elf/mod.rs
```diff
@@ -388,6 +411,38 @@ if_sylvan! {
         };
         Ok(nchain)
     }
+
+    struct Misc {
```
m4b, line 415: i like this!

src/elf/mod.rs
m4b: this also feels very lowlevel as a pub function; on argument is a `u16` machine value, which doesn't seem great for user to think about; if this is needed for some reason, i think we can provide something better here (still not sure why it's needed exactly, hoping you can fill that in)

m4b (commented):
This looks actually really great! Want some clarification on the new pub functions, and also the new `parse_elf_hdr`, otherwise I think we can merge this very soon, and thanks again for the PR! I didn't think a lazy elf parser would be so easy;

Oh it might also be good to add a new example showing how to parse out portions in a lazy manner, etc. the test is good, but example filling in other parts of structs, just as proof of concept would be good (i assume the user has all the facilities to do this at the moment?)

The only other concern i have is the "semantics" of a lazy elf parse; i assume it's safe to use, and no unusual things will happen if the user creates an elf struct in the lazy manner (i.e., no other public methods on the struct will panic if called and it was constructed in the lazy manner, yes?)

vfsfitvnm:
I second this.
In my specific use case, I just need to parse the elf file header in order to get `e_ident[EI_CLASS]` and `e_machine`.
It would be a lot nicer if there was a straightforward way to do that (https://github.com/m4b/goblin/pull/254/files#diff-696bbf0d8b3f75177e0fa931a45c3e85057e2078598ee1cf8040393730b30633R206), instead of include `scroll` as a direct dependency.

src/elf/mod.rs
m4b: Can we rename this to parse_header then since elf is redundant a bit since user would write `Elf::parse_elf_hdr`

jessehui (commented):
@m4b Thank you for your review. I changed `gnu_hash_len` and `hash_len` back to private functions. I can add an example later.

src/elf/mod.rs
```diff
@@ -202,22 +202,45 @@ if_sylvan! {
         pub fn is_object_file(&self) -> bool {
             self.header.e_type == header::ET_REL
         }
+
+        pub fn parse_hdr(bytes: &'a [u8]) -> error::Result<Header> {
```
m4b, line 206: One last thing: we need doc comment on this function just for consistency, and i think it should be named `parse_header`; the type is `Header`, and the field is `header` so it is unusual to use the `elf.h` spelling here (but nowhere else).

tests/elf.rs
```diff
@@ -33,8 +33,49 @@ fn parse_gnu_hash_section(base: &[u8], symbol_name: &str) -> Result<Sym, &'stati
...
+    let ctx = Ctx {
+        le: scroll::Endian::Little,
+        container: Container::Big,
+    };
+
+    // get section headers
+    let section_headers =
+        SectionHeader::parse(base, header.e_shoff as usize, header.e_shnum as usize, ctx)
+            .map_err(|_| "parse section headers error")?;
+    obj.section_headers = section_headers;
```
m4b, line 55: very minor nit: i would have assigned directly into field here.

m4b (commented):
Once doc comment and renamed I think this is ready to merge, thank you for the PR! This is great stuff.
It's also occurred to me in future, we could add e.g., `elf.parse_section_headers()` which takes `&mut self`; then the `Elf::parse` code could do:

```rust
  let mut elf = Elf::lazy_parse()?;
  elf.parse_section_headers();
  elf.parse_whatever()?;
```

and ditto for users of lazy parsing, easier for them to fill out which portions they want without doing the freestanding parse functions filling a field up, etc. just a thought ( definitely don't do that now) :)

src/elf/mod.rs
```diff
@@ -202,22 +202,45 @@ if_sylvan! {
...
+            bytes.pread::<Header>(0)
+        }
+
+        // Lazy parse the ELF contents. This function mainly just assembles an Elf struct. Once we have the struct, we can choose to parse whatever we want.
```
m4b, line 210: this is also missing an extra /, just noticed it

m4b:
Example getting interpreter would be great! Re gnu hash, let's leave it private for now, that way we don't commit to an api; if someone needs this functionality, we can always make a PR/take a PR and review how best to expose it.

```rust
                let mut num_syms = if let Some(gnu_hash) = dyn_info.gnu_hash {
                    gnu_hash_len(bytes, gnu_hash as usize, ctx)?
                } else if let Some(hash) = dyn_info.hash {
                    hash_len(bytes, hash as usize, header.e_machine, ctx)?
                } else {
                    0
                };
                let max_reloc_sym = dynrelas.iter()
                    .chain(dynrels.iter())
                    .chain(pltrelocs.iter())
                    .fold(0, |num, reloc| cmp::max(num, reloc.r_sym));
                if max_reloc_sym != 0 {
                    num_syms = cmp::max(num_syms, max_reloc_sym + 1);
                }
                dynsyms = Symtab::parse(bytes, dyn_info.symtab, num_syms, ctx)?;
```

as far as i can see, this is only code that uses it; we could just a a helper function that does all this and takes whichever minimal arguments requierd, seems like `dyn_info, ctx, bytes, and dyn_relas and header` :shrug: 

For now i think it's better to leave private, as someone can always do as you say and parse whole thing :)

src/elf/mod.rs
jessehui: Got it. Done.

tests/elf.rs
jessehui: Done.

src/elf/mod.rs
jessehui: Done.

m4b (approved):
This is awesome, thanks so much for this!

m4b:
I can add this change onto the 0.3 branch and do (another :D ) release, 0.3.4 if you like?

jessehui:
@m4b Thank you for your review. Yeah sure, I was just going to ask whether there will be a recent release with this feature that can be used directly from crates.io?

m4b:
You got it, 0.3.4 is out!

jessehui:
Great! Thank you very much!

PR #256, opened 2021-01-29
Avoid overflow when header.n_namesz is 0

glandium:


src/elf/note.rs
```diff
@@ -205,7 +205,7 @@ if_alloc! {
             };
             debug!("{:?} - {:#x}", header, *offset);
             // -1 because includes \0 terminator
-            let name = bytes.gread_with::<&'a str>(offset, ctx::StrCtx::Length(header.n_namesz - 1))?;
+            let name = bytes.gread_with::<&'a str>(offset, ctx::StrCtx::Length(header.n_namesz.saturating_sub(1)))?;
             *offset += 1;
```
glandium, line 209: Come to think of it, this is probably wrong when n_names was 0 in the first place?

src/elf/note.rs
```diff
@@ -205,7 +205,7 @@ if_alloc! {
             };
             debug!("{:?} - {:#x}", header, *offset);
             // -1 because includes \0 terminator
-            let name = bytes.gread_with::<&'a str>(offset, ctx::StrCtx::Length(header.n_namesz - 1))?;
+            let name = bytes.gread_with::<&'a str>(offset, ctx::StrCtx::Length(header.n_namesz.saturating_sub(1)))?;
```
m4b, line 208: won't this read a zero-length name? What does this mean exactly, that the note is empty itself, or that it doesn't have a note name, just a desc? I'm not super familiar with elf notes lately, and what the expectations are (probably none?)

src/elf/note.rs
m4b: maybe? i don't know haha; maybe the desc starts after a null-delimited empty string

src/elf/note.rs
glandium: I got a crash in goblin where when reaching here the value of header is:
```
goblin::elf::note::NoteHeader {n_namesz: 0, n_descsz: 0, n_type: 0}
```
I suppose this means the not section is empty.

glandium: What does gread_with do when given a length of 0?

m4b: i don't know offhand :sweat_smile: ; i suspect it'll just return an empty string, let me check

m4b: it does this (test passes):
```rust
#[test]
fn pread_zero_length() {
    let bytes = b"hello";
    let s = bytes.pread_with::<&str>(0, ctx::StrCtx::Length(0));
    assert!(s.is_ok());
    assert_eq!(s.unwrap(), "");
}
```

glandium: my guess is that gread_with increases offset with the length read, so the *offset += 1 is to go past the \0 terminator when there is one to continue reading the desc.

m4b: yes i think this is correct, assuming an empty note is `\0` in the bytes (which sort of makes sense to me); are you expecting something in desc? if this fixes it for you we can just add it; it won't hurt anything and it's certainly better than panicking. my experience with notes is they're just kind of a super unspecified thing, hah. if someone comes and complains that they have an empty note name but their desc is off by 1, then they weren't using this lib before since it would have panicked, and we can just cross that bridge or fix it when it happens

m4b (approved):

glandium:
Note the overflow only crashes on debug builds.

m4b:
the only thing that worries me slightly is the offset is returned:

```rust
            Ok((Note {
                name,
                desc,
                n_type: header.n_type,
            }, *offset))
```
so we did technically increment it by 1 even if we got empty strings and empty desc, so their len is all 0, but we manually bump by +1 for the null terminator. now i'm kind of thinking we shouldn't bump by +1?

I can't remember offhand what's after notes, i think they're just byte streams in the file and the offset isn't used in gread/pread chaining; iirc, each note has a particular offset and goblin preads it out from there?

if not, and the notes are read by gread chaining with offsets, then the +1 could be trouble, i'm not sure. this is definitely a weird corner case hah

m4b:
So like, iiuc, your binary has a note section with sh_offset, but the notes have empty names? lol

glandium:
The new version of the patch doesn't make it increase the offset. The binary is the in-memory vdso (we use goblin to generate crash reports in Firefox, that's part of what we dump).

src/elf/note.rs
```diff
@@ -205,8 +205,10 @@ if_alloc! {
             };
             debug!("{:?} - {:#x}", header, *offset);
             // -1 because includes \0 terminator
-            let name = bytes.gread_with::<&'a str>(offset, ctx::StrCtx::Length(header.n_namesz - 1))?;
-            *offset += 1;
+            let name = bytes.gread_with::<&'a str>(offset, ctx::StrCtx::Length(header.n_namesz.saturating_sub(1)))?;
+            if (header.n_namesz > 0) {
+                *offset += 1;
```
m4b, line 210: yup i think this is the right thing to do, the offset is used in subsequent notes, via NoteIterator

m4b (approved):

m4b:
do you need a new version immediately? there's a minor breaking change (#253) that will be merged for returning option for file size ranges, i can bump patch version and push this before merging that

glandium:
We're actually on 0.1 currently for some reason, so a bump in version number is not affecting us more than upgrading to 0.3 would already do.

m4b:
well i published 0.3.2 anyway with this change, so it's there now; thanks for the patch!

PR #258, opened 2021-02-01
example: add lazy parse example

jessehui:
This example shows how to use a lazy parse manner to reduce the read IOs and to save time by parsing only parts we care about.


m4b (approved):
This is great, thanks for this!

Issue #110, opened 2018-10-14
Fuzz: Round 2


m4b:
I want to fuzz goblin's parser again, this time with:

1. elf seeds using what we have now + seeds based on bsd elf binaries, etc.
2. mach-o seeds
3. PE seeds.
4. object file seeds

4 is new but it's also very important. My guess is mach-o and PE will have the most issues, as they haven't been fuzzed at all yet.

I would love for someone to help set this up, and even just file issues as the errors start emerging :)

sanxiyn:
I feel like trying [Angora](https://github.com/AngoraFuzzer/Angora), a fuzzer with very promising algorithms (also written in Rust!), and isn't our friend goblin a trusty good target?! Stay tuned!

The insight of AFL/LibFuzzer is that a little smart (coverage guiding) combined with speed beats full smart, symbolic approaches like KLEE. Angora shows it can be combined with a little more smart (taint guiding). Two instrumented executables are produced for coverage and taint, run taint a little (it's slow!), run coverage a lot, but do not mutate randomly but use taint info!, it all seems beautiful. Let's see how it works in practice.

m4b:
Super excited to see what you uncover here :) The mach-o and and PE backends could use some fuzz-love too :)

Shnatsel:
In the meanwhile I've done a pass with plain old AFL. I've targeted the elf parser since it contains a suspicious amount of unsafe code. I've also thrown in some FreeBSD binaries as seeds. Here's the generated corpus that can be used for subsequent fuzzing: [min-elf-corpus.zip](https://github.com/m4b/goblin/files/2933543/min-elf-corpus.zip)

Fuzzing harness targets ELF decoding specifically, this usually produces better results than all-in-one: https://github.com/Shnatsel/goblin/blob/master/fuzz-afl/src/main.rs

PR #260, opened 2021-02-11
Fix arithmetic debug assertions in elf

Mrmaxmeier:
This adds a new fuzzing harness that covers most of the ELF module and fixes a few overflow assertions.

`saturating_add`s should be fine since all of the edge cases concern sizes that are bounds-checked later.

(Xref #110)

### DIFF
--- a/src/elf/mod.rs
+++ b/src/elf/mod.rs
@@ -145,7 +145,7 @@ if_sylvan! {
                     iters.push(note::NoteDataIterator {
                         data,
                         offset,
-                        size: offset + phdr.p_filesz as usize,
+                        size: offset.saturating_add(phdr.p_filesz as usize),
                         ctx: (alignment, self.ctx)
                     });
                 }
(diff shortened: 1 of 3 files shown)


src/elf/mod.rs
```diff
@@ -406,9 +406,9 @@ if_sylvan! {
     fn hash_len(bytes: &[u8], offset: usize, machine: u16, ctx: Ctx) -> error::Result<usize> {
         // Based on readelf code.
         let nchain = if (machine == header::EM_FAKE_ALPHA || machine == header::EM_S390) && ctx.container.is_big() {
-            bytes.pread_with::<u64>(offset + 4, ctx.le)? as usize
+            bytes.pread_with::<u64>(offset.saturating_add(4), ctx.le)? as usize
```
m4b, line 409: i think there was a concern about this before about what this would mean if it overflows, but i suppose saturating add is best here

m4b (approved):
Thanks for adding another fuzz target, this is great!

PR #261, opened 2021-02-14
Remove unneeded unsafe in data_directories.rs

nico-abram:
This pull request replaces get_unchecked calls with direct indexing ( [] ). This removes a bunch of unsafe blocks

As far as I can tell, this should not affect optimized code: Since rust statically knows the size of the array, it should be able to elide the bounds checks

This [godbolt link](https://rust.godbolt.org/z/Pd19W6) shows how both the version with and without the unsafe are compiled to the same function with opt-level=3. [This](https://rust.godbolt.org/z/Wrb3eT) is the same example without the derives adding noise to the output

What I found interesting is, according to [this godbolt link](https://rust.godbolt.org/z/4dzs53) when compiling without optimizations, rust also elides the bounds checks for the indexing version, and the get_unchecked version has to call the get_unchecked function which does not get inlined
imagen

So I would expect the new version to outperform the unsafe version on debug builds, and match the unsafe version with optimizations

### DIFF
--- a/src/pe/data_directories.rs
+++ b/src/pe/data_directories.rs
@@ -45,62 +45,62 @@ impl DataDirectories {
     }
     pub fn get_export_table(&self) -> &Option<DataDirectory> {
         let idx = 0;
-        unsafe { self.data_directories.get_unchecked(idx) }
+        &self.data_directories[idx]
     }
     pub fn get_import_table(&self) -> &Option<DataDirectory> {
         let idx = 1;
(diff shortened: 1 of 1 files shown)


m4b (approved):
@nico-abram thank you so much for doing the legwork here, including checking assembly outputs; this has come up before in context of a a "safety audit" (which never materialized), but i'm glad to see the research done to see that switching to regular indexing has no perf regression (this was stated generally, without evidence). Much appreciated!

PR #262, opened 2021-02-15
Replace unchecked unsafe slicing with regular slicing in elf test

nico-abram:
[This test](https://github.com/m4b/goblin/blob/48f6ad968309cb0acfc0364d269763572cc35c6a/tests/elf.rs#L20-L23) performs the equivalent of subslicing a slice through unsafe code and avoids the bounds checking:

```rs
    let hashtab: &[u8] = unsafe {
        let addr = base.as_ptr().add(hash_section.sh_offset as usize);
        let size = hash_section.sh_size as usize;
        slice::from_raw_parts(addr, size)
    };
```
Since it's a test, I expect the bounds checking to not matter (I also expect that any potential overhead is dwarfed by everything else this function is doing like the vec allocation or the Elf parsing)

### DIFF
--- a/tests/elf.rs
+++ b/tests/elf.rs
@@ -1,5 +1,3 @@
-use std::slice;
-
 use goblin::elf::section_header::SHT_GNU_HASH;
 use goblin::elf::sym::Sym;
 use goblin::elf::Elf;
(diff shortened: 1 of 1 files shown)


m4b (approved):
yup, agreed, not needed, thanks for this!

PR #271, opened 2021-04-27
WIP implement macos compact unwinding info

Gankra:
This is a port of https://github.com/getsentry/symbolic/pull/372/ to move the core of the implementation into goblin, per @jan-auer's suggestion that it should live in this crate.

I haven't really properly implemented the main entry point that's hanging off of the Macho type, as I'm not sure what the right way to grab the target architecture is.

### DIFF
--- a/src/mach/compact_unwind_info.rs
+++ b/src/mach/compact_unwind_info.rs
@@ -1,0 +1,1503 @@
+//! Support for the "compact unwinding format" used by Apple platforms,
+//! which can be found in __unwind_info sections of binaries.
+//!
+//! The primary type of interest is CompactUnwindInfoIter, which can be
+//! constructed from a CompactUnwindInfo, which can be constructed from
+//! a UnwindInfoFrame.
+//!
+//! The CompactUnwindInfoIter lets you iterate through all of the mappings
+//! from instruction addresses to unwinding instructions, or lookup a specific
+//! mapping by instruction address.
+//!
+//! Example:
+//!
+//! ```rust,ignore
+//! fn process_dwarf<'d: 'o, 'o, O>(&mut self, object: &O) -> error::Result<(), CfiError>
+//!     where
+//!         O: ObjectLike<'d, 'o> + Dwarf<'o>,
+//! {
+//!     let endian = object.endianity();
+//!     if let Some(section) = object.section("unwind_info") {
+//!         let frame = UnwindInfoFrame::new(&section.data, endian);
+//!         let info = CompactUnwindInfo::new(object, section.address, frame);
+//!
+//!         let mut iter = info.iter()?;
+//!         while let Some(entry) = iter.next()? {
+//!             let base_address = entry.instruction_address;
+//!             let len = entry.len;
+//!             if let Some(instructions) = entry.cfi_instructions(&iter) {
+//!                 for instruction in instructions {
+//!                     // Interpret the CFI instructions
+//!                 }
+//!             }
+//!         }
+//!     }
+//! }
+//! ```
+//!
+//! # Unimplemented Features (TODO)
+//!
+//! * ARM64 opcode decoding (and writing the section on that format)
+//! * Personality/LSDA lookup (for runtime unwinders)
+//! * Entry lookup by address (for runtime unwinders)
+//! * x86/x64 Stackless-Indirect mode decoding (for stack frames > 2KB)
+//! * x86/x64 DWARF mode decoding (for when compact unwinding defers to the DWARF in eh_frame)
+//!
+//!
+//! # The Compact Unwinding Format
+//!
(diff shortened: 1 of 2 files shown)


src/mach/mod.rs
```diff
@@ -114,6 +115,37 @@ impl<'a> MachO<'a> {
...
+                    }
+                }
+            }
+            None
+        };
+
+        if let Some((_header, data)) = get_unwind_info_section() {
+            compact_unwind_info::CompactUnwindInfoIter::new(data, self.little_endian, self.is_64)
+        } else {
+            Err(error::Error::Malformed(String::from("No unwind_info section found")))
```
m4b, line 146: Maybe this should just return option instead of malformed? Or are unwind_info sections required to be present in a macho binary?

src/mach/mod.rs
```diff
@@ -114,6 +115,37 @@ impl<'a> MachO<'a> {
...
         }
     }
+    pub fn compact_unwind_info(&self) -> error::Result<compact_unwind_info::CompactUnwindInfoIter> {
+        let get_unwind_info_section = || {
+            let section_name = "unwind_info";
+            for segment in &self.segments {
+                for section in segment {
+                    if let Ok((header, data)) = section {
+                        if let Ok(sec) = header.name() {
+                            if sec.len() >= 2 && &sec[2..] == section_name {
```
m4b, line 125: why `&sec[2..]` is there some ascii prefix it is stripping here? If it's always the same, maybe it's better to have `section_name` contain that prefix and remove this part?

src/mach/compact_unwind_info.rs
```diff
@@ -0,0 +1,1503 @@
...
+}
+
+// x86/x64 implementation
+impl Opcode {
+    fn x86_cfi_instructions(
+        &self,
+        iter: &CompactUnwindInfoIter,
+    ) -> Option<std::vec::IntoIter<CfiOp>> {
+        let pointer_size = self.pointer_size(iter) as i32;
+        // TODO: don't allocate for this (use ArrayVec..?)
```
m4b, line 1165: I'd prefer to take on no more dependencies

src/mach/compact_unwind_info.rs
```diff
@@ -0,0 +1,1503 @@
+//! Support for the "compact unwinding format" used by Apple platforms,
```
m4b, line 1: I love the documentation! So this file will take me a bit of time to review, thankfully it looks like half of it is documentation. It might also help to get another expert to review, any help appreciated :) I will probably hold off doing a review until this PR is ready to be non WIP though, so please do let me know when you think it's ready :)

m4b (commented):
It's a good chunk of code! :D Let me know when it's no longer WIP, and ready to be considered for merging/review, and I'll get to the rest then; otherwise, lgtm generally speaking; as I mentioned in a comment, I would greatly prefer no more dependencies being added than what we already have

src/mach/mod.rs
jan-auer: That's the two underscores, for example `"__unwind_info"`. This function is structured similarly to `object` and `symbolic` which allow to search in both ELF and MachO for a section name regardless of the specific prefix.

src/mach/mod.rs
Gankra: Yes that's a good idea. At the time it felt strange to have `Result<Option<...>>` but now it's totally reasonable to me.

src/mach/mod.rs
Gankra: (yea this is a straight-up copy-paste of `symbolic` to resolve the section by name, since that functionality seemed to be absent.)

src/mach/compact_unwind_info.rs
Gankra: Ok! In the original PR on symbolic I have now replaced the Vec with a little adhoc ArrayVec impl to get rid of the Vec without introducing that dependency.

src/mach/compact_unwind_info.rs
Gankra: I have been iterating on this back in [the original PR on symbolic](https://github.com/getsentry/symbolic/pull/372/) and I believe it is in its ~final form, modulo what everyone wants me to do with [the outstanding TODOs](https://github.com/getsentry/symbolic/pull/372/#issuecomment-831357054) that can potentially be deferred for followups.

Just to avoid constantly synching things over and over, I'd like all review to go on there and then when everyone's happy I'll port the impl back over to goblin and we can discuss the implementation/signature of the entry point in `mach`.

The large outstanding question for me for integrating what I have into goblin is how to, using only goblin's API, detect the architecture (x86, x64, ARM64) which is otherwise not specified in the unwind_info section but determines how opcodes are interpreted.

PR #275, opened 2021-06-02
Parse Strtab only once on Creation

Lichtso:
Together with @jackcmay I am working on a [BPF VM](https://github.com/solana-labs/rbpf), which utilizes goblin for loading user provided ELF files.

So, last week we encountered a problem with the way `Strtab::get()` works, tried to fix it outside, which did not work and now have found a fix which would require upstream changes. Furthermore, I think these changes can also help other projects depending on goblin which are exposed to potentially hand-crafted adversary / worst case inputs too.

### The problem
A `get` request performs a linear scan to find the separators / `NULL` termination and possibly invalid UTF-8 variable length encodings on every single request. So it is possible to construct e.g. an ELF file which is only a few MiB in size but has massive overlap in the `Strtab`. This way everything involving the `Strtab` suddenly becomes O(n^2) in runtime complexity because of the overlap (as opposed to independent strings which can be amorphized to be constant length) and just reading the file can take minutes to hours.

### What did not work
We tried to fix it on our side by caching the results of `Strtab::get()`. But
* there are places like this: https://github.com/m4b/goblin/blob/b2b15a3a1a8cf6e9f9a3d3db7173cf9a18f71a2e/src/elf/dynamic.rs#L450 which are always called when loading an ELF file: https://github.com/m4b/goblin/blob/54405549abf226eafe3db62cf26757bd7af1e13b/src/elf/mod.rs#L304
* and caching afterwards does not help because each overlapping byte creates a new entry which again scans half the `Strtab` on average. So still O(n^2) runtime complexity in total.

### What does work
The proposed fix in this PR instead moves the parsing from the `get` request to the constructor of the `Strtab`, which in turn also moves the error handling with it. The trick is to use a binary search (in each `get` request) over the cached positions of the separators (found by one linear scan in the constructor). This is only done when dynamic (heap) allocations are enabled via `#[cfg(feature = "alloc")]`, otherwise we default to the original behavior.

### Missing
I did not address or test UTF-8 support yet, so it is probably broken. I will invest more time into it if this PR looks otherwise promising.

### DIFF
--- a/src/strtab.rs
+++ b/src/strtab.rs
@@ -84,8 +84,19 @@ impl<'a> Strtab<'a> {
     /// Converts the string table to a vector of parsed strings.
     ///
     /// Requires `feature = "alloc"`
-    pub fn to_vec(&self) -> Vec<&'a str> {
-        self.strings.iter().map(|&(_key, value)| value).collect()
+    pub fn to_vec(&self) -> error::Result<Vec<&'a str>> {
+        // Fallback in case `Strtab` was created using `from_slice_unparsed()`.
+        if self.strings.is_empty() {
+            let mut result = Vec::new();
+            let mut i = 0;
+            while i < self.bytes.len() {
+                let string = get_str(i, self.bytes, self.delim)?;
+                result.push(string);
+                i += string.len() + 1;
+            }
+            return Ok(result);
+        }
+        Ok(self.strings.iter().map(|&(_key, value)| value).collect())
     }
     #[cfg(feature = "alloc")]
     /// Safely gets a str reference from the parsed table starting at byte `offset`.
--- a/src/strtab.rs
+++ b/src/strtab.rs
@@ -27,7 +27,9 @@ fn get_str(offset: usize, bytes: &[u8], delim: ctx::StrCtx) -> scroll::Result<&s
 
 impl<'a> Strtab<'a> {
     /// Creates a `Strtab` directly without bounds check and without parsing it.
-    pub fn from_slice_unsafe(bytes: &'a [u8], offset: usize, len: usize, delim: u8) -> Self {
+    ///
+    /// This is potentially unsafe and should only be used if `feature = "alloc"` is disabled.
+    pub fn from_slice_unparsed(bytes: &'a [u8], offset: usize, len: usize, delim: u8) -> Self {
         Self {
             delim: ctx::StrCtx::Delimiter(delim),
             bytes: &bytes[offset..offset + len],
@@ -38,7 +40,7 @@ impl<'a> Strtab<'a> {
     /// Gets a str reference from the backing bytes starting at byte `offset`.
     ///
     /// If the index is out of bounds, `None` is returned. Panics if bytes are invalid UTF-8.
-    /// Use this method if the `Strtab` was created using `from_slice_unsafe()`.
+    /// Use this method if the `Strtab` was created using `from_slice_unparsed()`.
     pub fn get_unsafe(&self, offset: usize) -> Option<&'a str> {
         if offset >= self.bytes.len() {
             None
@@ -50,6 +52,7 @@ impl<'a> Strtab<'a> {
     /// Parses a `Strtab` from `bytes` at `offset` with `len` size as the backing string table, using `delim` as the delimiter.
     ///
     /// Errors if bytes are invalid UTF-8.
+    /// Requires `feature = "alloc"`
     pub fn parse(bytes: &'a [u8], offset: usize, len: usize, delim: u8) -> error::Result<Self> {
         let (end, overflow) = offset.overflowing_add(len);
         if overflow || end > bytes.len() {
@@ -61,9 +64,9 @@ impl<'a> Strtab<'a> {
                 overflow
             )));
         }
-        let mut result = Self::from_slice_unsafe(bytes, offset, len, delim);
+        let mut result = Self::from_slice_unparsed(bytes, offset, len, delim);
         let mut i = 0;
-        while i < len {
+        while i < result.bytes.len() {
             let string = get_str(i, result.bytes, result.delim)?;
             result.strings.push((i, string));
             i += string.len() + 1;
@@ -72,11 +75,15 @@ impl<'a> Strtab<'a> {
     }
     #[cfg(feature = "alloc")]
     /// Parses a `Strtab` with `bytes` as the backing string table, using `delim` as the delimiter between entries.
+    ///
+    /// Requires `feature = "alloc"`
     pub fn new(bytes: &'a [u8], delim: u8) -> error::Result<Self> {
         Self::parse(bytes, 0, bytes.len(), delim)
     }
     #[cfg(feature = "alloc")]
     /// Converts the string table to a vector of parsed strings.
+    ///
+    /// Requires `feature = "alloc"`
     pub fn to_vec(&self) -> Vec<&'a str> {
         self.strings.iter().map(|&(_key, value)| value).collect()
     }
@@ -100,12 +107,13 @@ impl<'a> Strtab<'a> {
             }
         }
     }
-    #[deprecated(since = "0.4.2", note = "Use from_slice_unsafe() instead")]
+    #[deprecated(since = "0.4.2", note = "Use from_slice_unparsed() instead")]
     /// Construct a strtab from a `ptr`, and a `size`, using `delim` as the delimiter
     ///
+    /// # Safety
     /// This function creates a `Strtab` directly from a raw pointer and size
     pub unsafe fn from_raw(ptr: *const u8, len: usize, delim: u8) -> Strtab<'a> {
-        Self::from_slice_unsafe(core::slice::from_raw_parts(ptr, len), 0, len, delim)
+        Self::from_slice_unparsed(core::slice::from_raw_parts(ptr, len), 0, len, delim)
     }
     #[deprecated(since = "0.4.2", note = "Bad performance, use get_at() instead")]
     #[cfg(feature = "alloc")]
@@ -158,7 +166,7 @@ impl<'a> Index<usize> for Strtab<'a> {
 #[test]
 fn as_vec_no_final_null() {
     let strtab = Strtab::new(b"\0printf\0memmove\0busta", 0x0).unwrap();
-    let vec = strtab.to_vec();
+    let vec = strtab.to_vec().unwrap();
     assert_eq!(vec.len(), 4);
     assert_eq!(vec, vec!["", "printf", "memmove", "busta"]);
 }
@@ -166,7 +174,7 @@ fn as_vec_no_final_null() {
 #[test]
 fn as_vec_no_first_null_no_final_null() {
     let strtab = Strtab::new(b"printf\0memmove\0busta", 0x0).unwrap();
-    let vec = strtab.to_vec();
+    let vec = strtab.to_vec().unwrap();
     assert_eq!(vec.len(), 3);
     assert_eq!(vec, vec!["printf", "memmove", "busta"]);
 }
@@ -174,7 +182,7 @@ fn as_vec_no_first_null_no_final_null() {
 #[test]
 fn to_vec_final_null() {
     let strtab = Strtab::new(b"\0printf\0memmove\0busta\0", 0x0).unwrap();
-    let vec = strtab.to_vec();
+    let vec = strtab.to_vec().unwrap();
     assert_eq!(vec.len(), 4);
     assert_eq!(vec, vec!["", "printf", "memmove", "busta"]);
 }
@@ -182,7 +190,7 @@ fn to_vec_final_null() {
 #[test]
 fn to_vec_newline_delim() {
     let strtab = Strtab::new(b"\nprintf\nmemmove\nbusta\n", b'\n').unwrap();
-    let vec = strtab.to_vec();
+    let vec = strtab.to_vec().unwrap();
     assert_eq!(vec.len(), 4);
     assert_eq!(vec, vec!["", "printf", "memmove", "busta"]);
 }
--- a/src/strtab.rs
+++ b/src/strtab.rs
@@ -3,7 +3,6 @@
 
 use core::fmt;
 use core::ops::Index;
-use core::slice;
 use core::str;
 use scroll::{ctx, Pread};
 if_alloc! {
@@ -15,8 +14,10 @@ if_alloc! {
 /// member index). Constructed using [`parse`](#method.parse)
 /// with your choice of delimiter. Please be careful.
 pub struct Strtab<'a> {
-    bytes: &'a [u8],
     delim: ctx::StrCtx,
+    bytes: &'a [u8],
+    #[cfg(feature = "alloc")]
+    strings: Vec<(usize, &'a str)>,
 }
 
 #[inline(always)]
@@ -25,31 +26,31 @@ fn get_str(offset: usize, bytes: &[u8], delim: ctx::StrCtx) -> scroll::Result<&s
 }
 
 impl<'a> Strtab<'a> {
-    /// Construct a new strtab with `bytes` as the backing string table, using `delim` as the delimiter between entries
-    pub fn new(bytes: &'a [u8], delim: u8) -> Self {
-        Strtab {
+    /// Creates a `Strtab` directly without bounds check and without parsing it.
+    pub fn from_slice_unsafe(bytes: &'a [u8], offset: usize, len: usize, delim: u8) -> Self {
+        Self {
             delim: ctx::StrCtx::Delimiter(delim),
-            bytes,
+            bytes: &bytes[offset..offset + len],
+            #[cfg(feature = "alloc")]
+            strings: Vec::new(),
         }
     }
-    /// Construct a strtab from a `ptr`, and a `size`, using `delim` as the delimiter
-    /// # Safety
+    /// Gets a str reference from the backing bytes starting at byte `offset`.
     ///
-    /// This function creates a `Strtab` directly from a raw pointer and size
-    pub unsafe fn from_raw(ptr: *const u8, size: usize, delim: u8) -> Strtab<'a> {
-        Strtab {
-            delim: ctx::StrCtx::Delimiter(delim),
-            bytes: slice::from_raw_parts(ptr, size),
+    /// If the index is out of bounds, `None` is returned. Panics if bytes are invalid UTF-8.
+    /// Use this method if the `Strtab` was created using `from_slice_unsafe()`.
+    pub fn get_unsafe(&self, offset: usize) -> Option<&'a str> {
+        if offset >= self.bytes.len() {
+            None
+        } else {
+            Some(get_str(offset, self.bytes, self.delim).unwrap())
         }
     }
     #[cfg(feature = "alloc")]
-    /// Parses a strtab from `bytes` at `offset` with `len` size as the backing string table, using `delim` as the delimiter
-    pub fn parse(
-        bytes: &'a [u8],
-        offset: usize,
-        len: usize,
-        delim: u8,
-    ) -> error::Result<Strtab<'a>> {
+    /// Parses a `Strtab` from `bytes` at `offset` with `len` size as the backing string table, using `delim` as the delimiter.
+    ///
+    /// Errors if bytes are invalid UTF-8.
+    pub fn parse(bytes: &'a [u8], offset: usize, len: usize, delim: u8) -> error::Result<Self> {
         let (end, overflow) = offset.overflowing_add(len);
         if overflow || end > bytes.len() {
             return Err(error::Error::Malformed(format!(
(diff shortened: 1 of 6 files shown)


src/strtab.rs
```diff
@@ -60,37 +55,54 @@ impl<'a> Strtab<'a> {
...
+        match self
+            .strings
+            .binary_search_by_key(&offset, |&(key, _value)| key)
+        {
+            Ok(index) => Some(self.strings[index].1),
+            Err(index) => {
+                if index == 0 {
+                    return None;
+                }
+                let pair = self.strings[index - 1];
```
m4b, line 94: i think destructing into tuple better here, e.g.:
```
let (candidate_offset, candidate_str)
```
(for example)

also may want to add a comment here explaining this logic, it's more complex and subtle

src/strtab.rs
```diff
@@ -60,37 +55,54 @@ impl<'a> Strtab<'a> {
...
         }
-        Ok(Strtab {
-            bytes: &bytes[offset..end],
-            delim: ctx::StrCtx::Delimiter(delim),
-        })
+        let mut result = Self::from_slice_unsafe(bytes, offset, len, delim);
+        {
+            let mut i = 0;
+            while i < len {
+                let string = get_str(i, result.bytes, result.delim)?;
```
m4b, line 62: this is stricter now, since any invalid string will cause failure at construction, instead of at `get` time, but this is the price we pay, i suppose ! :)

src/strtab.rs
```diff
@@ -60,37 +55,54 @@ impl<'a> Strtab<'a> {
...
+    ///
     /// If the index is out of bounds, `None` is returned.
     /// Requires `feature = "alloc"`
     #[cfg(feature = "alloc")]
-    pub fn get(&self, offset: usize) -> Option<error::Result<&'a str>> {
-        if offset >= self.bytes.len() {
-            None
-        } else {
-            Some(get_str(offset, self.bytes, self.delim).map_err(core::convert::Into::into))
+    pub fn get(&self, offset: usize) -> Option<&'a str> {
```
m4b, line 84: note: this is a breaking change; i think we can avoid a breaking change by adding a new method `get_at` or whatever, which uses the precompiled vector for acceleration (and prevalidated), but this is unless that is done, this is technically a breaking change

src/strtab.rs
```diff
@@ -133,36 +148,32 @@ impl<'a> Index<usize> for Strtab<'a> {
 
 #[test]
 fn as_vec_no_final_null() {
-    let bytes = b"\0printf\0memmove\0busta";
-    let strtab = unsafe { Strtab::from_raw(bytes.as_ptr(), bytes.len(), 0x0) };
```
m4b, line 137: thank you for removing the unsafe tests here; probably should at some point add tests for the unsafe api, but here the unsafety was not super important for testing basic functionality

src/strtab.rs
```diff
@@ -60,37 +55,54 @@ impl<'a> Strtab<'a> {
...
+    #[cfg(feature = "alloc")]
+    /// Converts the string table to a vector of parsed strings.
+    pub fn to_vec(&self) -> Vec<&'a str> {
+        self.strings.iter().map(|&(_key, value)| value).collect()
+    }
+    /// Safely gets a str reference from the parsed table starting at byte `offset`.
+    ///
     /// If the index is out of bounds, `None` is returned.
     /// Requires `feature = "alloc"`
     #[cfg(feature = "alloc")]
```
m4b, line 83: it's curious this method (`get`) required `alloc` feature before, but i suppose it's a happy accident since now you really are allocating if you use this `get`

src/strtab.rs
```diff
@@ -25,31 +26,25 @@ fn get_str(offset: usize, bytes: &[u8], delim: ctx::StrCtx) -> scroll::Result<&s
...
+        Self {
             delim: ctx::StrCtx::Delimiter(delim),
-            bytes,
-        }
-    }
-    /// Construct a strtab from a `ptr`, and a `size`, using `delim` as the delimiter
-    /// # Safety
-    ///
-    /// This function creates a `Strtab` directly from a raw pointer and size
-    pub unsafe fn from_raw(ptr: *const u8, size: usize, delim: u8) -> Strtab<'a> {
```
m4b, line 39: hmm why did this get removed? (also breaking change)

src/strtab.rs
```diff
@@ -60,37 +55,54 @@ impl<'a> Strtab<'a> {
...
-        Ok(strings)
+    /// Parses a `Strtab` with `bytes` as the backing string table, using `delim` as the delimiter between entries.
+    pub fn new(bytes: &'a [u8], delim: u8) -> error::Result<Self> {
+        Self::from_slice(bytes, 0, bytes.len(), delim)
     }
-    /// Safely parses and gets a str reference from the backing bytes starting at byte `offset`.
+    #[cfg(feature = "alloc")]
+    /// Converts the string table to a vector of parsed strings.
+    pub fn to_vec(&self) -> Vec<&'a str> {
+        self.strings.iter().map(|&(_key, value)| value).collect()
```
m4b, line 77: this is a bit more elegant now :)

src/strtab.rs
```diff
@@ -25,31 +26,25 @@ fn get_str(offset: usize, bytes: &[u8], delim: ctx::StrCtx) -> scroll::Result<&s
...
+            strings: Vec::new(),
         }
     }
     #[cfg(feature = "alloc")]
-    /// Parses a strtab from `bytes` at `offset` with `len` size as the backing string table, using `delim` as the delimiter
-    pub fn parse(
+    /// Parses a `Strtab` from `bytes` at `offset` with `len` size as the backing string table, using `delim` as the delimiter.
+    ///
+    /// Errors if bytes are invalid UTF-8.
+    pub fn from_slice(
```
m4b, line 42: also breaking change; this seems to actually parse it now which seems even more a correct name?

m4b (changes requested):
this is super interesting! so i think we need to figure out the breaking changes situation, whether it's worth it (i'm sort of ok with it, in particular the `get` method return signature (from `Ok(Some` to `Some` seems probably ok, but the various methods removed/renamed doesn't seem necessary?, but maybe you can change my mind :)

otherwise this is looking good; only other concern i have is for perf, parsing entire table at once could have some perf consequences? But maybe not, majority of symbols might be looked up intiially when we construct the dyn sym table + the symbol table, i'm not sure; do you have any insight into this aspect?

lasty, i'll have to reread your explanation why doing an initial parse avoids the o(n^2) runtime when doing gets, because it's not clear to me how that would happen; do you happen to have an elf binary you can share that exhibits this quadratic behavior?

otherwise this is super cool, thank you for making the PR and writing up your findings, this is very interesting :) (and in case it's not clear, yes i think this PR is welcome, and modulo concerns above, i'm ok with merging something like this)

src/strtab.rs
Lichtso: Probably because `map_err(core::convert::Into::into)` allocated an error message `String`.

src/strtab.rs
```diff
@@ -60,44 +61,70 @@ impl<'a> Strtab<'a> {
...
+        {
+            Ok(index) => Some(self.strings[index].1),
+            Err(index) => {
+                if index == 0 {
+                    return None;
+                }
+                let (string_begin_offset, entire_string) = self.strings[index - 1];
+                let offset_in_string = offset - string_begin_offset;
+                if offset_in_string < entire_string.len() {
+                    Some(&entire_string[offset_in_string..])
```
philipc, line 101: This will panic if it isn't a char boundary. I'm not sure if the return value of this function needs to distinguish between out of bounds and invalid UTF-8 like `get` does.

Lichtso: Thanks, I changed it to `str::get()`

src/strtab.rs
Lichtso: reverted to original

src/strtab.rs
Lichtso: reverted to original, and marked deprecated.
One should use `from_slice_unsafe` directly as that makes clear that no parsing happens.

src/strtab.rs
Lichtso: > which in turn also moves the error handling with it

As I already remarked in the opening message.

src/strtab.rs
Lichtso: Done: `(string_begin_offset, entire_string)`

philipc:
Do tools like `readelf`, `objdump`, `llvm-readobj` and the libraries they use also have this problem?

willglynn:
This reminds me of [`pdb::ItemFinder`](https://docs.rs/pdb/0.7.0/pdb/struct.ItemFinder.html). The situation there is similar: there's a data structure containing variable length items, which might be ill-formed, to which the user needs rapid access by index.

`pdb::ItemFinder` is to be updated by the user from a (fallible) iterator which parses one item at a time. The associated data structures form a directed acyclic graph, so the most common situation is to scan and then to refer to items which have already been seen, but separating the iterator and the finder allows (forces) the user to choose whichever index-building and error handling strategy is best for their application.

Additionally, given `N` records, `pdb::ItemFinder` stores `N >> x` offsets for some small value of `x`. This reduces memory requirements by a factor of `2^x` in exchange for iterating over up to `2^x - 1` records to retrieve a particular item. Iteration is so cheap – the data's usually in the same cache line anyway – that real code using this structure is actually faster due to the reduced recordkeeping overhead. Neither `pdb` nor `goblin` yet depend on a version of Rust with const generics so it's not easy to let the user parameterize this tradeoff, but my suspicion is that a similar tradeoff for some small value of `x` would be worthwhile. Note also that this doesn't have to make retrieval fallible: if parsing a `&[u8]` worked before, it would be safe for `get()` to `.expect()` that parsing it will work again.

Lichtso:
> Do tools like `readelf`, `objdump`, `llvm-readobj` and the libraries they use also have this problem?

@philipc That is a good question, but I will have to craft different ELF files because the current ones are not completely valid in unrelated areas and it makes these tools return early with a warning.

@m4b I added two unit tests for UTF-8 parsing and the CI is passing, can you review it again?

m4b:
> We have multiple and I wouldn't want to publish them before this is resolved.
Could you clarify what you mean here? I.e., why would resolution of this issue enable you to publish / share the problematic binaries? It seems to me if you can't share them now (which is fine), you wouldn't be able to share them after, regardless of whether this issue is resolved.

Lichtso:
> > We have multiple and I wouldn't want to publish them before this is resolved.
> 
> Could you clarify what you mean here? I.e., why would resolution of this issue enable you to publish / share the problematic binaries? It seems to me if you can't share them now (which is fine), you wouldn't be able to share them after, regardless of whether this issue is resolved.

We would simply like to keep using the upstream repo, not a fork with our security patches applied. That is all. Once this is merged we can deploy a new version of our software packages and release the exploits as they would be disarmed by the time. By release I mean first to you specifically and later to the general public, as we would want other projects depending on this one to have a change to adopt as well.

src/strtab.rs
```diff
@@ -60,44 +61,65 @@ impl<'a> Strtab<'a> {
...
+        }
+    }
+    #[deprecated(since = "0.4.2", note = "Use from_slice_unsafe() instead")]
+    /// Construct a strtab from a `ptr`, and a `size`, using `delim` as the delimiter
+    ///
+    /// This function creates a `Strtab` directly from a raw pointer and size
+    pub unsafe fn from_raw(ptr: *const u8, len: usize, delim: u8) -> Strtab<'a> {
+        Self::from_slice_unsafe(core::slice::from_raw_parts(ptr, len), 0, len, delim)
+    }
+    #[deprecated(since = "0.4.2", note = "Bad performance, use get_at() instead")]
```
m4b, line 110: awesome

src/strtab.rs
```diff
@@ -15,8 +14,10 @@ if_alloc! {
 /// member index). Constructed using [`parse`](#method.parse)
 /// with your choice of delimiter. Please be careful.
 pub struct Strtab<'a> {
-    bytes: &'a [u8],
     delim: ctx::StrCtx,
+    bytes: &'a [u8],
```
m4b, line 18: nit: layout here didn't need to change (bytes after delim in field declarations), this and some of the functions moving around did make this PR a bit hard to review, just fyi

src/strtab.rs
```diff
@@ -25,31 +26,31 @@ fn get_str(offset: usize, bytes: &[u8], delim: ctx::StrCtx) -> scroll::Result<&s
 }
 
 impl<'a> Strtab<'a> {
-    /// Construct a new strtab with `bytes` as the backing string table, using `delim` as the delimiter between entries
-    pub fn new(bytes: &'a [u8], delim: u8) -> Self {
-        Strtab {
+    /// Creates a `Strtab` directly without bounds check and without parsing it.
+    pub fn from_slice_unsafe(bytes: &'a [u8], offset: usize, len: usize, delim: u8) -> Self {
```
m4b, line 30: e.g., this function here makes a lot of diff noise; also does this function need to be `pub`? one can construct a strtable now, but the vec isn't filled, which means any `get_at` calls would fail, which feels like the strtab is "half constructed"; also it is better to not introduce any new pub apis unless necessary, since once they're pub, we can't take them back

src/strtab.rs
```diff
@@ -60,44 +61,65 @@ impl<'a> Strtab<'a> {
...
+    }
+    #[cfg(feature = "alloc")]
+    /// Parses a `Strtab` with `bytes` as the backing string table, using `delim` as the delimiter between entries.
+    pub fn new(bytes: &'a [u8], delim: u8) -> error::Result<Self> {
+        Self::parse(bytes, 0, bytes.len(), delim)
     }
-    /// Safely parses and gets a str reference from the backing bytes starting at byte `offset`.
+    #[cfg(feature = "alloc")]
+    /// Converts the string table to a vector of parsed strings.
+    pub fn to_vec(&self) -> Vec<&'a str> {
```
m4b, line 80: unfortunately this is still a breaking change, since return signature changed. i'm 50/50 on fence whether it's worth it (can just Ok wrap at end and it's non-breaking, but maybe it's fine to break?)

src/strtab.rs
```diff
@@ -60,44 +61,65 @@ impl<'a> Strtab<'a> {
...
+                let (string_begin_offset, entire_string) = self.strings[index - 1];
+                entire_string.get(offset - string_begin_offset..)
+            }
+        }
+    }
+    #[deprecated(since = "0.4.2", note = "Use from_slice_unsafe() instead")]
+    /// Construct a strtab from a `ptr`, and a `size`, using `delim` as the delimiter
+    ///
+    /// This function creates a `Strtab` directly from a raw pointer and size
+    pub unsafe fn from_raw(ptr: *const u8, len: usize, delim: u8) -> Strtab<'a> {
```
m4b, line 107: ah i see now why `from_slice_unsafe` is pub; so i'm not sure this should be deprecated? i used this in the past, e.g., to construct strtables for a dynamic linker, which doesn't want to allocate (since it's already mapped into memory). i'm not sure what others are doing, but it would force someone to first call from `raw_parts` on their pointer and len as you do here.

alternatively, it would be feasible to have from_slice_unsafe just get the ptr + len and pass it to this. however, it's probably better to get rid of / deprecate these unsafe apis, since it doesn't add much (the user can just call the unsafe `from_raw_parts` themselves in their own unsafe context, so we push it onto them, which i think is fine).

so i guess what i'm saying is maybe it might be good to either:
1. delete this
2. deprecate as you do with warning it will be removed, then delete next release

m4b (commented):
ok so other than `to_vec` being a breaking change, which if you feel strongly about, i guess i'm ok with it, the only other major concern i have is whether we're pessimizing 99% of binary's strtables in what seems like a very niche problem case.
as far as i understand it, the only security implications I see are a kind of DoS on accessing strtables for these malformed binaries, but it's not clear to me how that is impactful, other than a tool taking a long time to run? it would be good to get some further motivations here.

I'm not sure why, but I almost wish we had a variant constructor for Strtab that uses the pre-allocated form via a config we'd pass into parsing, and defaults to a non-preallocated form otherwise, but this is a much bigger PR. However, it might be worth doing, because a config for parsing options is something i've thought would be nice for a while.

with that in mind, i don't see anything in this PR that would prevent such a feature from landing in the future?

example:
```rust
enum StrtabKind {
  Preallocated(Vec<&str>),
  ZeroCopy
}

pub struct Strtab {
  bytes: &[u8],
  inner: StrtabKind
}
```

where get's dispatch depending on the enum variant accordingly. Anyway, just some food for thought, as this would allow users to opt into whether they want to pay for the preallocated penalty or not, at a programmatic, library level choice.

ok i've blabbed enough :D

that being said, i'm in favor of merging, unless anyone else has any other concerns?

src/strtab.rs
Lichtso: The deprecated marker is a few lines above (the comment) in 103.

src/strtab.rs
Lichtso: Was for consistency with the constructors (see line 30 of the prev version).

src/strtab.rs
Lichtso: > ok so other than `to_vec` being a breaking change, which if you feel strongly about, i guess i'm ok with it, the only other major concern i have is whether we're pessimizing 99% of binary's strtables in what seems like a very niche problem case.

Well, we have no data what the "usual" user of this crate is doing on average, so it is hard to optimize for.
But, as I said, this is only penalizing use cases where only a few symbols are requested once. All others such as:
* Symbols being requested multiple times
* Symbols having suffix overlap
* Most symbols being requested

Would not be hit at all or even profit from the pre-parsing.

> as far as i understand it, the only security implications I see are a kind of DoS on accessing strtables for these malformed binaries, but it's not clear to me how that is impactful, other than a tool taking a long time to run? it would be good to get some further motivations here.

I agree that many CLI tools some developer runs and cancels, if they take to long, are not affected by this.
However, think about the automated processes like CI services, etc. Furthermore, we run a decentralized network which validates ELF files on the fly, so it would easily grind to halt if somebody wanted it to, which is not acceptable.

> I'm not sure why, but I almost wish we had a variant constructor for Strtab that uses the pre-allocated form via a config we'd pass into parsing, and defaults to a non-preallocated form otherwise, but this is a much bigger PR. However, it might be worth doing, because a config for parsing options is something i've thought would be nice for a while.
> 
> with that in mind, i don't see anything in this PR that would prevent such a feature from landing in the future?
> 
> example:
> 
> ```rust
> enum StrtabKind {
>   Preallocated(Vec<&str>),
>   ZeroCopy
> }

Might not be so complicated, why not just default back to the original implementation of `get` in `get_at` if the `strings` `Vec` is empty, because it was constructed by `from_slice_unsafe` not by `parse`?

Lichtso:
> ok so other than `to_vec` being a breaking change, which if you feel strongly about, i guess i'm ok with it, the only other major concern i have is whether we're pessimizing 99% of binary's strtables in what seems like a very niche problem case.

Well, we have no data what the "usual" user of this crate is doing on average, so it is hard to optimize for.
But, as I said, this is only penalizing use cases where only a few symbols are requested once. All others such as:
* Symbols being requested multiple times
* Symbols having suffix overlap
* Most symbols being requested

Would not be hit at all or even profit from the pre-parsing.

> as far as i understand it, the only security implications I see are a kind of DoS on accessing strtables for these malformed binaries, but it's not clear to me how that is impactful, other than a tool taking a long time to run? it would be good to get some further motivations here.

I agree that many CLI tools some developer runs and cancels, if they take to long, are not affected by this.
However, think about the automated processes like CI services, etc. Furthermore, we run a decentralized network which validates ELF files on the fly, so it would easily grind to halt if somebody wanted it to, which is not acceptable.

> I'm not sure why, but I almost wish we had a variant constructor for Strtab that uses the pre-allocated form via a config we'd pass into parsing, and defaults to a non-preallocated form otherwise, but this is a much bigger PR. However, it might be worth doing, because a config for parsing options is something i've thought would be nice for a while.
> 
> with that in mind, i don't see anything in this PR that would prevent such a feature from landing in the future?
> 
> example:
> 
> ```rust
> enum StrtabKind {
>   Preallocated(Vec<&str>),
>   ZeroCopy
> }

Might not be so complicated, why not just default back to the original implementation of `get` in `get_at` if the `strings` `Vec` is empty, because it was constructed by `from_slice_unsafe` not by `parse`?

That being said, maybe `from_slice_unparsed` would be a better name than `from_slice_unsafe` ...

src/strtab.rs
Lichtso: Ironically, I was on the fence of breaking it even further by including the starting offset of each entry in the result.

m4b: Ok so I've thought about this, and I think we should return `Result` here, so that we don't have a breaking change, and if we switch to using the enum, it will work without changes; if we get that done i'm ok with merging this and making a minor release asap.

As a bonus, we can perhaps construct the vector on the fly if it's `empty` (i.e., it was constructed via the unsafe method), but that is probably a niche usecase.

Lichtso: Both done, reverted to use the `Result` return type and implemented a fallback.

Lichtso:
I made the changes you requested, could you rerun the CI tests once more?

m4b (approved):
this is great, thanks for doing this :)

m4b:
> I made the changes you requested, could you rerun the CI tests once more?

I believe the CI has run on your last change `Reverts breaking to_vec() result and adds a fallback mode. `, there is a green check mark at least next to it :)

Anyway, thank you for your patience! I think this is ready to merge; perhaps we can do the enum variant I talked about earlier, but it's fine for now I believe; we can do a minor crates release as well, since we don't (i believe!) have any breaking changes.

Thanks again for everyone's input here!

Lichtso:
Thank you as well.

Once you release the minor crate version I will send you the exploits via email to m4b.github.io a gmail.com

m4b:
ok i've published 0.4.2; do note that i caught a breaking change in the `new` function whose return signature changed; i restored the original new, and added a `new_preparsed` that has the Result return type so that a minor release should be non-breaking now. So if you had a branch using these changes, you'll have to rename your `Strtab::new` -> `Strtab::new_preparsed` is all.

Thanks for the PR!

Issue #171, opened 2019-06-24
Missing import (IAT) information for Windows PE


elast0ny:
In the struct 
```Rust
Struct goblin::pe::import::SyntheticImportDirectoryEntry { ... }
```
The field `import_address_table` does not point to the IAT but instead points to the Import Name Table.

elast0ny:
The `import_directory_entry.import_address_table_rva` points to the correct offset of the first entry in the IAT but it somehow doesnt make it to the computer `import_address_table`.

Will dig into the code to see why this is hapenning

m4b:
Was this ever resolved ? Is it still an issue ?

elast0ny:
Hi, it appears that this is still an issue unless im computing these values wrong...

Here's an example program, it's output and a screenshot of the values parsed by Ghidra :
```rust
use goblin::pe::import::SyntheticImportLookupTableEntry;
use goblin::*;

use std::fs::File;
use std::io::Read;
use std::mem::size_of_val;

fn main() {
    let mut fd = File::open("target/debug/goblin_test.exe").unwrap();
    let mut buffer = Vec::new();
    fd.read_to_end(&mut buffer).unwrap();

    let win_pe = match Object::parse(&buffer).unwrap() {
        Object::PE(pe) => pe,
        _ => {
            println!("Unhandle bin type");
            return;
        }
    };

    let import_data = win_pe.import_data.unwrap().import_data;

    for cur_import in &import_data {
        if cur_import.name != "KERNEL32.dll" {
            continue;
        }

        let abs_ilt_base: usize =
            win_pe.image_base + cur_import.import_directory_entry.import_lookup_table_rva as usize;
        let abs_iat_base: usize =
            win_pe.image_base + cur_import.import_directory_entry.import_address_table_rva as usize;

        println!(
            "{:?} ILT : 0x{:X} IAT : 0x{:X}",
            cur_import.name, abs_ilt_base, abs_iat_base
        );

        for (i, lookup_table) in cur_import
            .import_lookup_table
            .as_ref()
            .unwrap()
            .iter()
            .enumerate()
        {
            match lookup_table {
                SyntheticImportLookupTableEntry::OrdinalNumber(ordinal) => {
                    print!("\tOrdinal {} ", ordinal);
                    continue;
                }
                SyntheticImportLookupTableEntry::HintNameTableRVA((_, v)) => {
                    print!("\t{} ", v.name)
                }
            };

            // This prints values that are in the import_lookup_table_rva
[...]

PR #278, opened 2021-08-02
pe.import: Add get_imports_rva to SynthethicImportDirectoryEntry

Maimonator:
I've added an extra function called `get_imports_rva`
To be honest I'm not sure about the correct interface so I'm leaving it like this and opening it to the discussion.

I've separated `get_imports_rva` according to the discussion in #171 , As saving this information was considered `wasteful` since it's easily calculated.
The problem with this function is that it needs to know the bitness of the PE (whether it's 32bit or 64bit) to properly calculate the RVAs.

Also, I started working with rust a few days ago, so any tips or comments are more than welcome.
Thanks!

### DIFF
--- a/src/pe/import.rs
+++ b/src/pe/import.rs
@@ -90,6 +90,11 @@ pub enum SyntheticImportLookupTableEntry<'a> {
     HintNameTableRVA((u32, HintNameTableEntry<'a>)), // [u8; 31] bitfield :/
 }
 
+pub struct SyntheticImportAddressTableEntry<'a> {
+    pub rva: u64,
+    pub name: &'a str,
+}
+
 pub type ImportLookupTable<'a> = Vec<SyntheticImportLookupTableEntry<'a>>;
 
 impl<'a> SyntheticImportLookupTableEntry<'a> {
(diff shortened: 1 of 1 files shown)


src/pe/import.rs
```diff
@@ -293,6 +298,32 @@ impl<'a> SyntheticImportDirectoryEntry<'a> {
             import_address_table,
         })
     }
+
+    pub fn get_imports_rva<T: Bitfield<'a>>(
+        &self,
+    ) -> error::Result<Vec<SyntheticImportAddressTableEntry<'a>>> {
+        let mut out = Vec::<SyntheticImportAddressTableEntry>::new();
```
m4b, line 305: I don’t see any fallibility here, so I’d drop the error return. Also I would attempt to get rid of allocation via returning a vec and try `impl Iterator<Item = SyntheticImportAddressTableEntry<‘a>> + ‘a`

(you may have to play with the lifetimes). Since it’s optional too you may have to do `self.import_table.unwrap_or(&[]).iter().enumerate().flatmap`

src/pe/import.rs
```diff
@@ -293,6 +298,32 @@ impl<'a> SyntheticImportDirectoryEntry<'a> {
...
+        if let Some(import_lookup_table) = &(self.import_lookup_table) {
+            for (i, lookup_entry) in import_lookup_table.iter().enumerate() {
+                match lookup_entry {
+                    SyntheticImportLookupTableEntry::OrdinalNumber(_) => {
+                        continue;
+                    }
+                    SyntheticImportLookupTableEntry::HintNameTableRVA((_, v)) => {
+                        let import_rva = (self.import_directory_entry.import_address_table_rva
+                            as u64)
+                            + ((i * std::mem::size_of::<T>()) as u64);
```
m4b, line 315: You mentioned needing to know the size of the PE; this information is usually in the parsing context but I’m not sure if you have access to it here, haven’t looked at this code in a while. Or is the mem::size_of sufficient for this ?

m4b (commented):
I haven’t had a chance to read the entirely of the thread prompting this, I’m afk intermittently for a bit, but this PR seems reasonable. Does it solve your major usecase ?

src/pe/import.rs
Maimonator: Got it will implement it as an iterator

src/pe/import.rs
Maimonator: `mem::size_of` is sufficient for this, but honestly passing a boolean called `is_64bit`  would be a cleaner option IMO.

Maimonator:
@m4b Thanks for reviewing, I'll do the fixes and make sure the PR passes the tests.

Issue #263, opened 2021-03-01
elf: Implement parsing of symbol versioning


messense:
See https://github.com/PyO3/maturin/pull/436/files#r581158722

johannst:
Hey :]
I'd be interested to take a look at this issue. If that's fine you can just assign me.
I planned to prototype on the weekend to get a starting point for some discussions.

m4b:
Go for it!

messense:
Thank you @johannst

PR #280, opened 2021-08-13
Elf gnu symbol versioning

johannst:
GNU symbol version extension for ELF files (#263).

WIP PR to discuss overall design.

Tasks:
- [x] Parse `.gnu.version_r` section
- [x] Parse `.gnu.version_d` section
- [x] Parse `.gnu.version` section
- [x] Test parsing `.gnu.version_r` section
- [x] Test parsing `.gnu.version_d` section
- [x] Test parsing `.gnu.version` section


johannst:
First implementation done, any feedback is welcome :]

src/elf/symver.rs
```diff
@@ -0,0 +1,902 @@
...
+
+        fn next(&mut self) -> Option<Self::Item> {
+            if self.offset >= self.bytes.len() {
+                None
+            } else {
+                // Safe to unwrap as the length of the byte slice was validated in VersymSection::parse.
+                let ElfVersym { vs_val } = self
+                    .bytes
+                    .gread_with(&mut self.offset, self.ctx.le)
+                    .unwrap();
```
philipc, line 276: What if `bytes.len()` is not a multiple of `core::mem::size_of::<ElfVersym>()`?

src/elf/symver.rs
```diff
@@ -0,0 +1,902 @@
...
+
+                Some(Versym {
+                    vs_val: vs_val as usize,
+                })
+            }
+        }
+
+        #[inline]
+        fn size_hint(&self) -> (usize, Option<usize>) {
+            let len = self.bytes.len() - self.offset;
```
philipc, line 286: Divide by `core::mem::size_of::<ElfVersym>()`?

src/elf/symver.rs
```diff
@@ -0,0 +1,902 @@
...
+                    vd_ndx,
+                    vd_cnt,
+                    vd_hash,
+                    vd_aux,
+                    vd_next,
+                } = self.bytes.pread_with(self.offset, self.ctx.le).unwrap();
+
+                // Get a slice of bytes of the `Verdaux` entries.
+                //
+                // | Verdef | .. | Verdef | Verdaux | Verdaux | .. | Verdef | ..
```
philipc, line 448: I don't think this layout is required. e.g. LSB says:
"The section shall contain an array of `Elfxx_Verdef` structures, as described in Figure 10-1, optionally followed by an array of `Elfxx_Verdaux` structures, as defined in Figure 10-2."
So all of the `Verdaux` may come after all of the `Verdef`. It's best to not make any assumptions about ordering (other than that `vd_next` and `vd_aux` won't point backwards), and I don't think other parsers do.

Same for the `Verneed` structures.

src/elf/symver.rs
```diff
@@ -0,0 +1,902 @@
...
+                    (vd_next - vd_aux) as usize
+                } else {
+                    // For the last entry, ElfVerdef->vd_next == 0.
+                    // Therefore we compute the remaining length of the bytes buffer.
+                    self.bytes.len() - self.offset - vd_aux as usize
+                };
+                let bytes: &'a [u8] = self
+                    .bytes
+                    .pread_with(self.offset + vd_aux as usize, len)
+                    .unwrap();
```
philipc, line 466: For this `unwrap`, and the one above: `VerdefSection::parse` only validated that the section offset and length are within the file size, not whether the section is long enough to contain the entries, and not whether `vd_next` and `vd_aux` are valid. You need to handle invalid values for those.

Similar comments for other unwraps.

src/elf/symver.rs
```diff
@@ -0,0 +1,902 @@
...
+    /// [lsb-verdef]: https://refspecs.linuxbase.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/symversion.html#VERDEFENTRIES
+    #[derive(Debug)]
+    pub struct Verdef<'a> {
+        pub vd_version: usize,
+        pub vd_flags: usize,
+        pub vd_ndx: usize,
+        pub vd_cnt: usize,
+        pub vd_hash: usize,
+        pub vd_aux: usize,
+        pub vd_next: usize,
```
philipc, line 507: I don't think `usize` is appropriate for all of these, since they don't vary between 32-bit and 64-bit ELF. e.g. `vd_hash` is only ever going to be used as a `u32`. I suggest only using `usize` for fields that the user will commonly use as an index (I'm not sure which fields qualify for that, but I doubt it is many).

philipc (commented):
This is looking quite good! I've only reviewed for correctness, not style (which is probably ok but I'll leave that to @m4b ).

johannst:
Thanks for the quick review, I'll take a look at the details after work.

src/elf/symver.rs
johannst: Good point, I'll validate the length of the byte slice in `VersymSection::parse`.

src/elf/symver.rs
johannst: Correct

src/elf/symver.rs
johannst: Interesting, I did interpret the optionally different here, but you are right. This code should not rely on that layout.

src/elf/symver.rs
johannst: Fair point, and that counts for all the other structures exposed.

src/elf/symver.rs
johannst: Removed all calls to unwrap, but this needs another round of review.

src/elf/symver.rs
```diff
@@ -0,0 +1,907 @@
...
+                    self.index = self.count;
+                    None
+                })
+            }
+        }
+
+        #[inline]
+        fn size_hint(&self) -> (usize, Option<usize>) {
+            let len = self.count - self.index;
+            (len, Some(len))
```
philipc, line 483:
```suggestion
            (0, Some(len))
```
We can return less if `sh_info` was wrong, or if the data is corrupt.

src/elf/symver.rs
```diff
@@ -0,0 +1,907 @@
...
+                    // Validate if offset is valid index into bytes slice.
+                    if offset >= iter.bytes.len() {
+                        return None;
+                    }
+
+                    // Get a slice of bytes starting with the first ElfVerdaux entry.
+                    let bytes: &'a [u8] = &iter.bytes[offset..];
+
+                    // Bump the offset to the next ElfVerdef entry.
+                    iter.offset = iter.offset.checked_add(vd_next as usize)?;
```
philipc, line 457:
```suggestion
                    iter.offset = iter.offset.checked_add(vd_next as usize)?;
                    if vd_next == 0 {
                        self.index = self.count;
                    }
```

philipc (commented):
A couple more things for better handling of corrupt data. These apply to all four of Verdef/Verdaux/Verneed/Vernaux.

src/elf/symver.rs
johannst: Good idea.

src/elf/symver.rs
johannst: Good catch, thanks!

johannst:
I was thinking, with the primitives available for reading, to also add some utility to lookup version strings.
For example something like `Versym -> Option<&str>`.
@philipc what do you think about that? Or should we target higher-level utilities in a new issue/PR?

philipc:
It's certainly useful to have something like that, but it's a bit harder to be sure what the correct API is, so maybe leave it for a new PR, but doing a prototype (and using it for a real task) wouldn't hurt.

johannst:
> It's certainly useful to have something like that, but it's a bit harder to be sure what the correct API is, so maybe leave it for a new PR, but doing a prototype (and using it for a real task) wouldn't hurt.

In that sense the PR is done from my side.

The only question is if there is a better way to handle the test ELF binaries? I optimized them for size but if you have an idea I would prefer to not commit them here?
Building during test execution doesn't seem to be a good idea due to the windows runners.
Maybe we could maintain another repository with all the binaries and include it submodule here? But that would also be a task for another issue/PR.

johannst:
@m4b any further feedback on style or completeness, please let me know.

m4b:
@johannst sorry for delay, and thank you for your patience; going to give this one last review, but it looks good to me; if i don't merge by tomorrow, feel free to ping.

There is not breaking changes yes? (I can release this in a minor release once it's merged if you like.

johannst:
Don't  worry, that's fine, you can take your time to review it.

Actually I have one concrete question about the `cfg` guard.
Other modules wrap the implementation in a macro, but as far as I can tell, rustfmt doesn't format this code.
But to align on style, would you prefer to use a macro in the `symver` module as well?

> There is not breaking changes yes? (I can release this in a minor release once it's merged if you like.

Right, it is compatible on the API level.

m4b:
Yea I believe we've had issues in past with rustfmt and the macro stuff, it's quite annoying.

So I believe the original reason for the macro was because it put the cfg on every item, which too burdensome manually (and error prone). If the module has some entire feature applicable to it, you can just put it on the module; if there is some subset of features though, like `alloc` then probably macro is necessary.

I'm not sure what your cfg needs are, need to do a closer review to see what you mean (or you could summarize what you're thinking if you like too :) )

johannst:
> I'm not sure what your cfg needs are, need to do a closer review to see what you mean (or you could summarize what you're thinking if you like too :) )

As is today, to really use the features of the `symver` module we require `(elf32 || elf64) && alloc`.
Due to the rustfmt issues with the macros I implemented the cfg guard with a nested impl module (see below).
But you are right, in this case (where the cfg applies globally to the module) it would have been better to directly put it on the module in `src/elf/mod.rs`.
```rust
#[cfg(all(any(feature = "elf32", feature = "elf64"), feature = "alloc"))]
pub use symver_impl::*;

#[cfg(all(any(feature = "elf32", feature = "elf64"), feature = "alloc"))]
mod symver_impl {
     ...
}
```

However if we decide that is it valuable to offer the ELF type definitions in `non-alloc` environments then I need to put the cfg guards on the items in the module itself.

> So I believe the original reason for the macro was because it put the cfg on every item, which too burdensome manually (and error prone).

Definitely agree, putting it by hand on every item is quiet error prone.
If I have some time I can check why rustftmt has a problem with the macros as we use them.

src/elf/symver.rs
```diff
@@ -59,876 +59,866 @@
...
-                    })
-            }
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.offset >= self.bytes.len() {
+            None
+        } else {
+            self.bytes
+                .gread_with::<ElfVersym>(&mut self.offset, self.ctx.le)
+                .ok()
+                .map(|ElfVersym { vs_val }| Versym { vs_val })
```
m4b, line 251: nit: It would be nicer imho to implement `From<ElfVersym> for Versym` then you could do `.map(Versym::from)` I believe.

src/elf/symver.rs
```diff
@@ -59,876 +59,866 @@
...
-                        None
-                    })
-            }
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.offset >= self.bytes.len() {
+            None
+        } else {
+            self.bytes
+                .gread_with::<ElfVersym>(&mut self.offset, self.ctx.le)
+                .ok()
```
m4b, line 250: this will stop iteration on first bad parse, which I assume is intended

src/elf/symver.rs
```diff
@@ -59,876 +59,866 @@
...
+/********************
+ *  ELF Structures  *
+ ********************/
+
+/// An ELF `Symbol Version` entry.
+///
+/// https://refspecs.linuxbase.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/symversion.html#SYMVERTBL
+#[repr(C)]
+#[derive(Debug, Pread)]
+struct ElfVersym {
```
m4b, line 102: Alternatively, since this appears identical to Versym, and I don't think the size is unified over 32/64 bit forms, it might be fine to just have repr(C) and Pread on Versym? But this is also fine for consistency with the other types , so it's also fine :) (also i don't know much about Versym, but I don't think we'd extend it in the future for some reason with fields other than vs_val: u16, so maybe it is better future proof to have this internal and Versym public as you have it now. :shrug:

src/elf/symver.rs
```diff
@@ -59,876 +59,866 @@
...
+
+                // Validate if offset is valid index into bytes slice.
+                if offset >= iter.bytes.len() {
+                    return None;
+                }
+
+                // Get a slice of bytes starting with the first ElfVerdaux entry.
+                let bytes: &'a [u8] = &iter.bytes[offset..];
+
+                // Bump the offset to the next ElfVerdef entry.
```
m4b, line 433: thank you for the comments :)

src/elf/symver.rs
```diff
@@ -59,876 +59,866 @@
...
-    #[derive(Debug)]
-    pub struct VerneedSection<'a> {
-        /// String table used to resolve version strings.
-        pub verstr: Strtab<'a>,
+impl<'a> VerneedSection<'a> {
+    /// Try to parse the optional [`SHT_GNU_VERNEED`] section.
+    pub fn parse(
         bytes: &'a [u8],
-        count: usize,
+        shdrs: &'_ [SectionHeader],
```
m4b, line 605: I'm a bit surprised you have to add the `'_` annotation; does default `&[SectionHeader]` lifetime elision not compile ?

src/elf/mod.rs
```diff
@@ -53,6 +53,8 @@ pub mod dynamic;
 #[macro_use]
 pub mod reloc;
 pub mod note;
+#[cfg(all(any(feature = "elf32", feature = "elf64"), feature = "alloc"))]
```
m4b, line 56: it's a little sad this only needs the `alloc` feature because (iiuc) it creates a `Strtab`, which I believe until only recently did not require the `alloc` feature either (maybe it still did, I can't remember).

Anyway, something we could possibly investigate fixing in the future, but it's not related to this PR.

src/elf/symver.rs
```diff
@@ -0,0 +1,924 @@
...
+/************************
+ *  Version Definition  *
+ ************************/
+
+/// Helper struct to iterate over [Version Definition][Verdef] and [Version Definition
+/// Auxiliary][Verdaux] entries.
+#[derive(Debug)]
+pub struct VerdefSection<'a> {
+    /// String table used to resolve version strings.
+    pub verstr: Strtab<'a>,
```
m4b, line 317: so this is a `pub` struct, and this field is `pub`; I don't see it used anywhere, so I presume it's for clients to resolve symbols; but I'm wondering if we can expose an api to resolve it for them? Since once this is a `pub` it's `pub` forever basically, so just making sure it's something want to commit to.

m4b (approved):
This looks great, thanks so much for your patience !

I just had some minor nits, but nothing even necessary to change.

Only real question was about the pub strtab field and "future proofing" to best of our ability, but otherwise this looks great, and looks ready to merge to me. Thanks again for your patience and for committing this awesome work!

src/elf/symver.rs
johannst: Good idea, will do that

src/elf/symver.rs
johannst: You are right, this should not be required.
I'll remove those.

src/elf/symver.rs
johannst: Right, these `Strtab`s are used to resolve any symbol version related strings.
Only the rustdoc examples at the top of the module show the usage of those.

> but I'm wondering if we can expose an api to resolve it for them

Out of my head I see the following two approaches:
1. Pass a reference to the corresponding `Strtab` down to `Verneed | Vernaux | Vernaux`, and provide APIs on those structs to get the strings.
2. Offer APIs on the sections to resolve the strings, eg `VerneedSection::file_name(_:&Verneed) -> Option<&str>`

Not really decided which one I prefer, maybe after a night of sleep :)

philipc: This strtab must be the same as `Elf::dynstrtab` (because the dynamic loader can't use section headers to find the strings). So it might be better to pass a reference to `Elf::dynstrtab` when needed instead of having this field.

johannst: @philipc good point, I was mainly following the LSB here.
> The sh_link member of the section header shall point to the section that contains the strings referenced by this section.

Out of interest I also look at the common static linkers, how they craft the version related sections ([gold](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=gold/layout.cc;hb=250d07de5cf6efc81ed934c25292beb63c7e3129#l4976), [ld](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=bfd/elf.c;hb=64dbf74d428e97de737c13f42767dab88493f187#l3979) and [lld](https://github.com/llvm/llvm-project/blob/b06426da764a8d0254521b33d667db8f26ae5e2f/lld/ELF/SyntheticSections.cpp#L3100)).

In that sense we could drop the `verstr` fields completely for now and only offer the minimal APIs. Users could then  resolve the strings against `Elf::dynstrtab` directly.

As discussed previously, we could then look into providing higher-level APIs in a second PR later (which I started to toy around with).

johannst: Committed that change for now and updated the doc examples.
Let's see your opinions.

src/elf/symver.rs
```diff
@@ -597,8 +572,6 @@ pub struct Verdaux {
 /// Auxiliary][Vernaux] entries.
 #[derive(Debug)]
 pub struct VerneedSection<'a> {
-    /// String table used to resolve version strings.
-    pub verstr: Strtab<'a>,
```
m4b: i feel much better about this now since it doesn't commit us here to anything pub

m4b (approved):
Ok this PR has been going for a while, and I don't want you to lose motivation; with the `pub` field removed, I think this is in a great state to merge, and we could add nicer apis at a later date if necessary (though it seems somewhat ok as it is, at least in example)

tests/bins/elf/symver/Makefile
```diff
@@ -0,0 +1,22 @@
+# Build shared libraries (32/64) with versioned symbols and applications (32/64).
```
m4b, line 1: this is unrelated to your patch but I realized goblin is using an `include` section, and we include `tests` wholesale; since I believe `include` is only for crates downloads, we should only include what this crate needs to build it's source as a dependency, not it's tests; so all this stuff is adding a lot of unnecessary downloads to people.

johannst: Honesty I don't like to commit binaries, but I didn't see a good way at the time, and I saw other binaries. I compiled them with `-Os` but it doesn't change the fact :)

Interesting, I didn't know about the `include` key, but I agree, for people only building we should remove the `tests/` folder.
So dependencies to `goblin` become slimmer and people wanting to hack on `goblin` will need to clone it from gh anyway.

johannst:
Thanks for your review.
I agree, I think it's in a good state to merge now.

m4b:
Thank you for your patience and this awesome PR!

m4b:
So I squashed your commits, and because some of your commit messages were quite good and looks like you took time to write them, I chose the best ones and included that in the main body; thanks again! I'll have a minor release soon, since this is non-breaking, though I may want to roll up other changes if possible, not sure; is this urgently needed for you?

johannst:
Thanks for the effort to extract some of the commit messages.

> I'll have a minor release soon, since this is non-breaking, though I may want to roll up other changes if possible, not sure; is this urgently needed for you?

That's fine for me, I mainly used this features in some experiments for now where I can use a version from gh.

messense:
I'd love to have a new release to land https://github.com/PyO3/maturin/pull/626, thanks!

m4b:
ok 0.4.3 is out; i trimmed the `include` field too to remove tests, which includes binary files, etc., so the crates.io download should be much slimmer too!

Issue #291, opened 2021-12-01
PE: missing exported symbols in PEs


dureuill:
Hello, and thank you for making goblin!

I've been trying to use goblin (through [symbolic](https://docs.rs/symbolic/)), but I noticed that there is a difference in the number of export symbols reported by goblin vs other tools (eg rabin2).

For instance, taking the binaries uploaded in attachment (from open source projects like VirtualBox), I notice the following differences in reported symbols:

binary | goblin | rabin2
[libxml2.dll](https://github.com/dureuill/goblin/raw/assets/assets/libxml2.dll) | 1596 | 1597
[VBoxMRXNP.dll](https://github.com/dureuill/goblin/raw/assets/assets/VBoxMRXNP.dll) | 498 | 504

The missing symbols in goblin look like "legit" symbols, e.g. `xmlXPathNAN` in `libxml2.dll` (right between `xmlXPathNamespaceURIFunction` and `xmlXPathNewBoolean`), and `g_pszRTAssertExpr` to `g_u32RTAssertLine` (right between `g_aRTUniUpperRanges` and `NPAddConnection`).

Looking for the root cause, I believe I identified it as being in the function `section_read_size` in `src/pe/utils.rs`.

It appears that the computed size is sometimes too small.

Reading from #101, the function was introduced in this PR so that goblin correctly loads [quine.exe](https://github.com/corkami/pocs/blob/master/PE/bin/quine.exe), so simply reverting is probably not an option.

~~By introducing the changes in #292, I am able to both find the same number of symbols as `rabin2`, and still load `quine.exe`.~~

~~However, I'm unsure of the change, since it deviates from the linked reference Stack Overflow post. The rationale for the change is that it makes sense to not take anything smaller than the virtual size, I guess?~~

EDIT: the changes of #293 appear to fix the problem without having to perform changes I'm unsure of on the computed size.

What do you think?

dureuill:
Fixed by #293, so I'm closing this!

PR #292, opened 2021-12-01
Never return a section size smaller than virtual size

dureuill:
See the associated issue (#291) for details.

This PR does the following:

1. Introduce a subtraction that was present in the [stackexchange post](https://reverseengineering.stackexchange.com/questions/4324/reliable-algorithm-to-extract-overlay-of-a-pe) linked in the source
2. Replace the `min` by a `max` so that the section size is at least as big as the virtual size. Note that this deviates from the linked stackexchange post. However, it does fix the associated issue for the tested binaries.

## Tests

Manual tests with the following test driver (using symbolic, that uses goblin):

```toml
# Cargo.toml
[package]
name = "symbolic_symbols"
version = "0.1.0"
edition = "2018"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
symbolic = "*"
anyhow = "1.0"
env_logger = "*"


[patch.crates-io]
goblin = { path = "goblin" }
```

```rust
// src/main.rs
use std::ops::Deref;

use anyhow::Context;
use symbolic::debuginfo::Object;

fn main() -> anyhow::Result<()> {
    env_logger::init();
    let data = std::fs::read(std::env::args_os().nth(1).context("Expected path to binary")?)?;
    let object = Object::parse(data.deref())?;
    for symbol in object.symbols() {
        println!("{}", symbol.name().unwrap());
    }
    Ok(())
}
```

On the two tests binaries: [libxml2.dll](https://github.com/dureuill/goblin/raw/assets/assets/libxml2.dll) and
[VBoxMRXNP.dll](https://github.com/dureuill/goblin/raw/assets/assets/VBoxMRXNP.dll) + [`quine.exe`](https://github.com/corkami/pocs/blob/master/PE/bin/quine.exe) from #101.

Compared with `rabin2 -E $PE`.

1. Before this PR, not the same number of symbols, quine.exe loads successfully (with 0 symbols)
2. After this PR, same number of symbols as rabin2, quine.exe still loads successfully (still with 0 symbols)

Thank you for your time and consideration!

### DIFF
--- a/src/pe/utils.rs
+++ b/src/pe/utils.rs
@@ -32,9 +32,9 @@ fn section_read_size(section: &section_table::SectionTable, file_alignment: u32)
     let size_of_raw_data = section.size_of_raw_data as usize;
     let virtual_size = section.virtual_size as usize;
     let read_size = {
-        let read_size = (section.pointer_to_raw_data as usize + size_of_raw_data + file_alignment
+        let read_size = ((section.pointer_to_raw_data as usize + size_of_raw_data + file_alignment
             - 1)
-            & !(file_alignment - 1);
+            & !(file_alignment - 1)) - aligned_pointer_to_raw_data(section.pointer_to_raw_data as usize);
         cmp::min(read_size, round_size(size_of_raw_data))
     };
 
--- a/src/pe/utils.rs
+++ b/src/pe/utils.rs
@@ -38,11 +38,7 @@ fn section_read_size(section: &section_table::SectionTable, file_alignment: u32)
         cmp::min(read_size, round_size(size_of_raw_data))
     };
 
-    if virtual_size == 0 {
-        read_size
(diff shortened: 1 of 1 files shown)


dureuill:
I see. Thank you for the answer. I'm not sure how to move forward. Would such a breaking change be possible for goblin?

Meanwhile, I will close this PR, as it is wrong. At least, it prompted your answer :sweat_smile:. The associated issue remains open  #291.

philipc:
I think a breaking change is okay if there is no alternative, and I don't think there is, so feel free to open a PR with that change (or reopen this PR after a force push).

dureuill:
Thank you. As I had already force pushed my branch, GitHub does not let me continue in this PR, so I opened #293.

PR #293, opened 2021-12-02
Support exports without offset

dureuill:
This PR is a replacement for #292, which applies the change discussed there: 

- make offset an `Option<_>` in `Export`, and support when an `Export` has no offset.

Moreover, the correct commit of #292 (add missing subtraction) is retained.

## Test

The same procedure as in #292 was followed.
This change does fix the issue #291  and the previously missing symbols are now returned for the tested binaries. 

Thanks again for your time and consideration!

### DIFF
--- a/src/pe/utils.rs
+++ b/src/pe/utils.rs
@@ -28,13 +28,36 @@ fn section_read_size(section: &section_table::SectionTable, file_alignment: u32)
         (size + PAGE_MASK) & !PAGE_MASK
     }
 
+    // Paraphrased from https://reverseengineering.stackexchange.com/a/4326 (by Peter Ferrie).
+    //
+    // Handles the corner cases such as mis-aligned pointers (round down) and sizes (round up)
+    // Further rounding corner cases:
+    // - the physical pointer should be rounded down to a multiple of 512, regardless of the value in the header
+    // - the read size is rounded up by using a combination of the file alignment and 4kb
+    // - the virtual size is always rounded up to a multiple of 4kb, regardless of the value in the header.
+    //
+    // Reference C implementation:
+    //
+    // long pointerToRaw = section.get(POINTER_TO_RAW_DATA);
+    // long alignedpointerToRaw = pointerToRaw & ~0x1ff;
+    // long sizeOfRaw = section.get(SIZE_OF_RAW_DATA);
+    // long readsize = ((pointerToRaw + sizeOfRaw) + filealign - 1) & ~(filealign - 1)) - alignedpointerToRaw;
+    // readsize = min(readsize, (sizeOfRaw + 0xfff) & ~0xfff);
+    // long virtsize = section.get(VIRTUAL_SIZE);
+    //
+    // if (virtsize)
+    // {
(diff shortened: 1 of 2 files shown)


dureuill:
Ran `cargo fmt` and pushed again

src/pe/utils.rs
```diff
@@ -32,9 +32,10 @@ fn section_read_size(section: &section_table::SectionTable, file_alignment: u32)
     let size_of_raw_data = section.size_of_raw_data as usize;
     let virtual_size = section.virtual_size as usize;
     let read_size = {
-        let read_size = (section.pointer_to_raw_data as usize + size_of_raw_data + file_alignment
-            - 1)
-            & !(file_alignment - 1);
+        let read_size =
+            ((section.pointer_to_raw_data as usize + size_of_raw_data + file_alignment - 1)
```
m4b, line 36: this is a bit difficult to read without some context a comment might be nice what's going on with some of the bitmasking and additions and subtractions, etc. (though I know you didn't introduce most of it)

Might be nice to add a comment at least abut the final subtraction introduced?

m4b (commented):
LGTM minus a comment might be nice for any future readers what's going on in that `read_size` computation; since this is a breaking change it won't go in until goblin 5.0, which I may wait for more breaking changes to be introduced

src/pe/utils.rs
dureuill: Hi! Thank you for the review.

I modified this line while looking for the cause to the missing symbols of #291.
I followed the [stackoverflow post](https://reverseengineering.stackexchange.com/a/4326) that is referenced line 17 of `src/pe/utils.rs`, a few lines above the `section_read_size` function, and noticed that, compared to the C code in that referenced post, we were missing the subtraction. So I added it, but so far it had no visible effect on #291. 

I don't have much more context to provide than the content of the referenced stack overflow answer.

Do you need me to move the existing comment closer to the calculation on this line, or to inline some of the content of the stackoverflow post?

m4b: So this PR does not fix the underlying issue of missing symbols in PE?

dureuill: Yes it does. The [first commit](https://github.com/m4b/goblin/pull/293/commits/f984749ae48c2fdf68ac493ea17dd2fa2509cc01) fixes the issue of missing symbols in PE by making the offset field optional in the `Export` struct.

m4b: Ok great I thought you were saying this PR did not fix anything. So to summarize:

it would be nice to know/have a corner case where this aligned subtraction catches something we miss, but for now it’s fine to put your link as comment and small blurb about what the content is (in case link dies for whatever reason)

dureuill: Yes, I can see how it could have been confusing :slightly_smiling_face:

I just updated the PR to add the link to the reference post close to the calculation, as well as the relevant details as (paraphrased) comment.

Please tell me what you think

m4b: wonderful!

m4b (approved):
LGTM, since this is a breaking change I'd like to roll it up potentially with other breaks

PR #295, opened 2021-12-05
Complete ELFOSABI constants

messense:
https://github.com/bminor/glibc/blob/bada2e312a8b94c5fc2f5571b249c71cb466a640/elf/elf.h#L135-L150


m4b (approved):
great thank you!

PR #296, opened 2021-12-28
complete some constants

n01e0:


src/elf/section_header.rs
```diff
@@ -188,7 +188,7 @@ pub const SHF_ORDERED: u32 = 1 << 30;
 /// Number of "regular" section header flags
 pub const SHF_NUM_REGULAR_FLAGS: usize = 12;
 // /// Section is excluded unless referenced or allocated (Solaris).
```
m4b, line 190: Can you un comment this doc comment now ?

m4b (changes requested):
Just the one thing I mentioned and ready to go, thank you!

m4b (approved):

PR #297, opened 2021-12-31
Only parse the last `SHT_SYMTAB` in `Elf::parse()`

Lichtso:
Only parse the last `SHT_SYMTAB`, because all earlier ones are not used anyway.
This improves the runtime complexity from O(n^2) to O(n),
and makes the parsing more robust against crazy inputs.

### DIFF
--- a/src/elf/mod.rs
+++ b/src/elf/mod.rs
@@ -293,13 +293,11 @@ if_sylvan! {
 
             let mut syms = Symtab::default();
             let mut strtab = Strtab::default();
-            for shdr in &section_headers {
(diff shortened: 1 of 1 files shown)


m4b (approved):
This was O(n^2) before because it would parse every symtab and every symtab could have referenced the entire string table, is that correct?

m4b:
Also I never received that binary which was the original impetus for changing string table to avoid DoS attacks :) would still be interested to see such a thing, can send to my email m4b dot github dot io atsign gmail.com

PR #298, opened 2022-01-04
Guard every with_capacity call with a bounds check

Swatinem:
The crate does a lot of pre-allocation based on untrusted input, which
can lead to unbounded allocation.

This came up in https://github.com/getsentry/symbolic/issues/479, and I see the crate already had some bounds checks for certain parts, though not for Mach-O parsing.

I went a bit overboard and added explicit bounds checks for every place where a `with_capacity` is called.

### DIFF
--- a/src/pe/export.rs
+++ b/src/pe/export.rs
@@ -108,6 +108,21 @@ impl<'a> ExportData<'a> {
         let number_of_name_pointers = export_directory_table.number_of_name_pointers as usize;
         let address_table_entries = export_directory_table.address_table_entries as usize;
 
+        if number_of_name_pointers > bytes.len() {
+            let message = format!(
+                "Buffer is too short for {} name pointers",
+                number_of_name_pointers
+            );
+            return Err(error::Error::Malformed(message));
+        }
+        if address_table_entries > bytes.len() {
+            let message = format!(
+                "Buffer is too short for {} address table entries",
+                address_table_entries
+            );
+            return Err(error::Error::Malformed(message));
+        }
+
         let export_name_pointer_table = utils::find_offset(
             export_directory_table.name_pointer_rva as usize,
             sections,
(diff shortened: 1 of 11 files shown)


m4b (approved):
this looks great, thank you!

Swatinem:
I’m sorry for being "that guy", but may I ask for this to be merged and eventually released? ;-)

src/pe/export.rs
```diff
@@ -108,6 +108,21 @@ impl<'a> ExportData<'a> {
...
+                number_of_name_pointers
+            );
+            return Err(error::Error::Malformed(message));
+        }
+        if address_table_entries > bytes.len() {
+            let message = format!(
+                "Buffer is too short for {} address table entries",
+                address_table_entries
+            );
+            return Err(error::Error::Malformed(message));
```
m4b, line 123: I’m wondering if we should introduce a TooShort(count, &’static str) error variant which would do:

“Buffer is too short for {0} {1}” since you write the same boilerplate message every time  (and technically allocate)

Swatinem: Good idea! However I noticed that `goblin::Error` is not flagged as `#[non_exhaustive]`, so adding a new variant would be technically a breaking change :-(

So how about doing that as a followup at a later time so it does not block releasing this fix.

m4b: There is a breaking change pending anyway, so it’s perfect timing.

Swatinem: That is good to know, I will do that tomorrow then

m4b:
Ok merged, @Swatinem thanks for your patience! Hopefully can release this tonight or tomorrow

m4b:
0k goblin 0.5 is out, thanks for your patience!

m4b:
oof, i forgot scroll went to 2021 edition, that's a bit annoying

m4b:
ok i had to yank it, because it's technically broken on the msrv, 1.40; so i have to bump that to edition 2021, which is, unfortunately, 1.56.0 Hopefully we don't have a change of MSRV until the next edition :)

PR #299, opened 2022-01-08
Strtab to raw vec

n01e0:
Implemented a function to get the `bytes` of `StrTab` in `Vec` without considering `delim`.

### DIFF
--- a/src/elf/constants_header.rs
+++ b/src/elf/constants_header.rs
@@ -333,7 +333,9 @@ pub const EM_BA2: u16 = 202;
 pub const EM_XCORE: u16 = 203;
 /// Microchip 8-bit PIC(r)
 pub const EM_MCHP_PIC: u16 = 204;
-// reserved 205-209
+/// Intel Graphics Technology
+pub const EM_INTELGT: u16 = 205;
+// reserved 206-209
 /// KM211 KM32
 pub const EM_KM32: u16 = 210;
 /// KM211 KMX32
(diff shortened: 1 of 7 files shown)


n01e0:
https://github.com/m4b/goblin/issues/300

```rust
$ cat src/main.rs
use std::fs::File;
use std::io::Read;
use std::env;
use goblin::elf::*;

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let args = env::args().collect::<Vec<String>>();
    if args.len() < 2 {
        eprintln!("Usage: {} binary", args[0])
    } else {
        let mut f = File::open(&args[1])?;
        let mut buf = Vec::new();
        f.read_to_end(&mut buf)?;
        let elf = Elf::parse(&buf)?;
        let shdrs = elf.section_headers;
        let raw_strtab = elf.shdr_strtab.to_raw_vec();
        for shdr in shdrs {
            let idx = shdr.sh_name;
            let mut name = String::new();
            for c in &raw_strtab[idx..] {
                if *c == 0 {
                    break;
                } else {
                    name.push(*c as char);
                }
            }
            println!("{}", name);
        }
  }

    Ok(())
}
$ ./target/debug/readstrtab `which ls`

.interp
.note.gnu.property
.note.gnu.build-id
.note.ABI-tag
.gnu.hash
.dynsym
.dynstr
.gnu.version
.gnu.version_r
.rela.dyn
.rela.plt
.init
.plt
.plt.got
.plt.sec
.text
.fini
.rodata
.eh_frame_hdr
.eh_frame
.init_array
.fini_array
.data.rel.ro
.dynamic
.got
.data
.bss
.gnu_debugaltlink
.gnu_debuglink
.shstrtab
```

n01e0:
I would like to change the way Elf.shdr_strtab is parsed in the future.

n01e0:
No need

Issue #300, opened 2022-01-08
elf: The strtab in the section header is wrong.


n01e0:
example

```rust
$ cat src/main.rs
use std::fs::File;
use std::io::Read;
use std::env;
use goblin::elf::*;

fn main() -> Result<(), Box<dyn std::error::Error>> {
  let args = env::args().collect::<Vec<String>>();
  if args.len() < 2 {
      eprintln!("Usage: {} binary", args[0])
  } else {
      let mut f = File::open(&args[1])?;
      let mut buf = Vec::new();
      f.read_to_end(&mut buf)?;
      let elf = Elf::parse(&buf)?;
      let strtab = elf.shdr_strtab.to_vec().unwrap();
      for s in strtab {
          println!("{}", s)
      }
  }

  Ok(())
}
$ cargo r `which ls`
~ snip ~
    Finished dev [unoptimized + debuginfo] target(s) in 0.16s
     Running `target/debug/readstrtab /usr/bin/ls`

.shstrtab
.interp
.note.gnu.property
.note.gnu.build-id
.note.ABI-tag
.gnu.hash
.dynsym
.dynstr
.gnu.version
.gnu.version_r
.rela.dyn
.rela.plt
.init
.plt.got
.plt.sec
.text
.fini
.rodata
.eh_frame_hdr
.eh_frame
.init_array
.fini_array
.data.rel.ro
.dynamic
.data
.bss
.gnu_debugaltlink
.gnu_debuglink
$ readelf -S `which ls`|grep "\["
  [Nr] Name              Type             Address           Offset
  [ 0]                   NULL             0000000000000000  00000000
  [ 1] .interp           PROGBITS         0000000000000318  00000318
  [ 2] .note.gnu.pr[...] NOTE             0000000000000338  00000338
  [ 3] .note.gnu.bu[...] NOTE             0000000000000358  00000358
  [ 4] .note.ABI-tag     NOTE             000000000000037c  0000037c
  [ 5] .gnu.hash         GNU_HASH         00000000000003a0  000003a0
  [ 6] .dynsym           DYNSYM           0000000000000450  00000450
  [ 7] .dynstr           STRTAB           0000000000001050  00001050
  [ 8] .gnu.version      VERSYM           0000000000001616  00001616
  [ 9] .gnu.version_r    VERNEED          0000000000001718  00001718
  [10] .rela.dyn         RELA             00000000000017b8  000017b8
  [11] .rela.plt         RELA             0000000000002bf8  00002bf8
[...]

PR #301, opened 2022-01-09
Documented Note

n01e0:
https://github.com/m4b/goblin/issues/300

In this PR we will work on documentation

### DIFF
--- a/src/strtab.rs
+++ b/src/strtab.rs
@@ -91,7 +91,8 @@ impl<'a> Strtab<'a> {
     #[cfg(feature = "alloc")]
     /// Converts the string table to a vector of parsed strings.
     ///
-    /// Note: If you want to get the correct `Vec` of shdr_strtab, don't use this method but use `get_at` to get it sequentially.
+    /// Note: This method is used to check the parsed contents of `strtab`.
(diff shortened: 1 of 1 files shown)


src/strtab.rs
```diff
@@ -91,6 +91,23 @@ impl<'a> Strtab<'a> {
     #[cfg(feature = "alloc")]
     /// Converts the string table to a vector of parsed strings.
     ///
+    /// Note: If you want to get the correct `Vec` of shdr_strtab, don't use this method but use `get_at` to get it sequentially.
```
m4b, line 94: Can you remove the comment about the section header strtab, as philipc said this applies to any strtab and primary use of this method is just to see the contents it parsed

src/strtab.rs
```diff
@@ -91,6 +91,23 @@ impl<'a> Strtab<'a> {
...
+    ///
+    /// # Examples
+    /// ```rust
```
m4b, line 97: This is awesome, thank you!

m4b (changes requested):
Just the bit about section header string table mentioned and otherwise ready to go

m4b (approved):
whoops, forgot about this, sorry for the delay!

PR #302, opened 2022-02-04
fix: Avoid arithmetic overflow and panics

Swatinem:
The first comes from a real and valid fat file that is ~6G, the other overflow was found while fuzzing.


src/mach/exports.rs
```diff
@@ -274,10 +274,16 @@ impl<'a> ExportTrie<'a> {
     /// Create a new, lazy, zero-copy export trie from the `DyldInfo` `command`
     pub fn new(bytes: &'a [u8], command: &load_command::DyldInfoCommand) -> Self {
         let start = command.export_off as usize;
-        let end = (command.export_size + command.export_off) as usize;
+        let end = start.saturating_add(command.export_size as usize);
```
m4b, line 277: Since this silently truncates maybe should do checked add and warn when it’s too large ? I assume this would only ever happen on a 32-bit system reading the 6GB binary ?

Swatinem: I added some `log::warn` for this and the `FatArch` case.
This here I found via fuzzing, though the `FatArch` was indeed a valid file.

src/mach/load_command.rs
```diff
@@ -507,25 +507,27 @@ impl<'a> ctx::TryFromCtx<'a, Endian> for ThreadCommand {
...
+            )));
+        }
+
         // get a byte slice of the thread state
         let thread_state_byte_length = count as usize * 4;
-        let thread_state_bytes = &bytes[16..16 + thread_state_byte_length];
 
         // check the length
-        if thread_state_bytes.len() < thread_state_byte_length {
+        if bytes.len() < 16 + thread_state_byte_length {
```
m4b, line 521: i know you didn't introduce 16 but it might be nice to turn this into a `const` since it's used again below and will be puzzling to future reviewers

m4b (approved):
lgtm, the 16 const nit is not super important

m4b:
thank you for adding the tests!

PR #303, opened 2022-02-27
mach.parse: Handle DyldExportsTrie

apalm:
Fixes missing exports.


src/mach/exports.rs
```diff
@@ -294,6 +294,30 @@ impl<'a> ExportTrie<'a> {
             location,
         }
     }
+
+    /// Create a new, lazy, zero-copy export trie from the `LinkeditDataCommand` `command`
+    pub fn new_from_linkedit_data_command(
```
apalm, line 299: Presumably, we don't want a variation of the `new` function, but was simple to implement. What do you think?

src/mach/exports.rs
```diff
@@ -294,6 +294,30 @@ impl<'a> ExportTrie<'a> {
...
+        bytes: &'a [u8],
+        command: &load_command::LinkeditDataCommand,
```
m4b, line 301: Unrelated to your PR< but i just noticed the capitalization for this is strange, surprised it's not LinkEditDataCommand; is this how it occurs in mach sources?

src/mach/exports.rs
m4b: The name is kind of long, but fine; however, can you dispatch both of these implemenations to a new private constructor, say `new_impl` which takes the bytes + the offset + size, since that is how they only differ, e.g.:

```rust
fn new_impl(bytes: &'a [u8], offset: usize, size: usize) -> Self {
 // etc.
}

pub fn new(bytes: &'a [u8], command: &load_command::DyldInfoCommand) -> Self {
  Self::new_impl(bytes, command.export_off as usize, ommand.export_size as usize)
}

// one more using the LinkeditDataComand
```

apalm: @m4b Updated.

src/mach/exports.rs
apalm: Given it's `LINKEDIT`, not `LINK_EDIT`, I think the current capitalization makes sense?

m4b (approved):
Thanks for this!

m4b:
@apalm do you need a release urgently? I like to roll up a few of these non-breaking changes before another patch release

apalm:
@m4b Nope, not urgent :)

lygstate:
What's [src/.DS_Store](https://github.com/m4b/goblin/pull/303/files#diff-c426beb14886224177ebe8240bcbdc4b304e37b31a77ff1b4d85134bca0ea00e) for ?

m4b:
@lygstate good catch, sorry that i missed this :/ i've deleted it, though it's a bit sad it'll be in the git history forever; i could force push it out, but i don't think it's worth breaking anyone with an upstream clone

m4b:
actually, i've changed my mind, i'm going to force push it out

lygstate:
> actually, i've changed my mind, i'm going to force push it out

better add it into gitignore

m4b:

and thanks for your patience waiting for this!

PR #305, opened 2022-05-16
elf: Non-alloc `ProgramHeader` exposes impl.

dancrossnyc:
* Expose more of the `ProgramHeader` implementation regardless
  of whether the `alloc` feature is enabled.  For example,
  `ProgramHeader::from_raw_parts` is a wrapper around
  `core::slice::from_raw_parts` that has `alloc` dependency.

Based on an internal patch by Keith Wesolowski (@wesolows).

Signed-off-by: Dan Cross <cross@oxidecomputer.com>


src/elf/program_header.rs
```diff
@@ -308,26 +310,35 @@ macro_rules! elf_program_header_std_impl { ($size:ty) => {
...
                 let mut program_headers = vec![ProgramHeader::default(); count];
                 let offset = &mut offset;
                 bytes.gread_inout_with(offset, &mut program_headers, ctx)?;
                 Ok(program_headers)
             }
 
+            #[cfg(feature = "alloc")]
             pub fn from_bytes(bytes: &[u8], phnum: usize) -> Vec<ProgramHeader> {
                 let mut phdrs = vec![ProgramHeader::default(); phnum];
-                phdrs.copy_from_bytes(bytes).expect("buffer is too short for given number of entries");
```
m4b, line 321: it's strange this format change happened

src/elf/program_header.rs
```diff
@@ -308,26 +310,35 @@ macro_rules! elf_program_header_std_impl { ($size:ty) => {
...
                 phdrs
             }
 
             /// # Safety
             ///
             /// This function creates a `ProgramHeader` directly from a raw pointer
-            pub unsafe fn from_raw_parts<'a>(phdrp: *const ProgramHeader,
-                                             phnum: usize)
-                                             -> &'a [ProgramHeader] {
+            pub unsafe fn from_raw_parts<'a>(
```
m4b, line 338: ditto, same; i wonder did newer versions of formatter alter it's line breaking/indentation rules

m4b (approved):
neato! lgtm, and thanks for the patch :)

m4b:
Just being clear, @dancrossnyc  this doesn't break the public api (it's only promoting things previous in alloc feature flag out to not requiring it, yes?) so specifically, if someone updates and had some feature matrix set they won't get broken with missing function, is what i'm wondering about

m4b:
oof; github appears to have gone bonkers due to my force push of cherry-picks (removing an erroneous .DS_STORE in a previous patch, which is now for some reason included in this PR), sorry about that.
I'm not sure why the .DS_STORE file is being re-added in this patch, but best bet is to `git pull --rebase` on master goblin, then cherry-pick your two patches here (or squash them, since I already reviewed) and then force push.

Sorry about that :/ i think it's better to have removed the 6KB .DS_STORE from the history though.

src/elf/program_header.rs
dancrossnyc: ...or change with respect to how formatting worked inside of macro expansions, perhaps?

dancrossnyc:
> oof; github appears to have gone bonkers due to my force push of cherry-picks (removing an erroneous .DS_STORE in a previous patch, which is now for some reason included in this PR), sorry about that. I'm not sure why the .DS_STORE file is being re-added in this patch, but best bet is to `git pull --rebase` on master goblin, then cherry-pick your two patches here (or squash them, since I already reviewed) and then force push.
> 
> Sorry about that :/ i think it's better to have removed the 6KB .DS_STORE from the history though.

No problem; done now. PTAL (I, uh, also accidentally closed the PR).

dancrossnyc:
Argh. Reopened in #308 because now I've gone and made github angry. :-/

Issue #309, opened 2022-05-27
coff: String table parsing returns more strings than exist in the table


track-5:
Platform: windows 19044.1706
goblin version: 0.5.1

## Expected behavior
When I parse the string table in a coff file, I expect it to return me just the strings that exist in the symbol table.

## Actual behavior
The string table parsing code instead returns three extra strings in the beginning of the list of strings.
The problem seems to be due to the fact we don't update the offset before reading the strings.

## Repro

You'll need to have the nasm (https://nasm.us/) assembler installed to assemble the program below: 

```asm
bits 64
default rel

segment .text
global main

extern ExitProcess
main:
	xor rax, rax
	call ExitProcess
```` 

Assemble the program using this command: `nasm -f win64 repro.asm -o repro.obj`

And this is an example of a Rust program that reads a Coff file and dumps all the strings found in its string table:

```rust
use goblin::pe::Coff;

use std::fs::File;
use std::io::BufReader;
use std::io::Read;

fn dump_strtab(coff: &Coff) {
	let strtab = coff.strings.to_vec().unwrap();
	println!("{:?}", strtab);
}

fn main() {
	if let filepath = std::env::args().skip(1).next() {
		let f = File::open(filepath.unwrap()).unwrap();
		let mut reader = BufReader::new(f);
		let mut buffer = Vec::new();
		reader.read_to_end(&mut buffer);

		match Coff::parse(buffer.as_slice()) {
			Ok(coff) => dump_strtab(&coff),
			Err(e) => println!("{:?}", e)
		}
	}
}
```

Run the program with `cargo run repro.obj`.

### Actual output

```
["\u{10}", "", "", "ExitProcess"]
```

### Expected output
```
["ExitProcess"]
```

The bug seems to be in the header.rs file (https://docs.rs/goblin/0.5.1/src/goblin/pe/header.rs.html#177). The code doesn't advance the offset by 4 (number of bytes for the string table size field). It ends up reading the 4 bytes as 3 different strings, which makes sense. The size should be 16 (0x10), so the first two bytes it reads make 0x10 and end at the third byte (0x0). The two other strings come from reading the subsequent 0x0.

m4b:
@track-5 i think this is closed now? please re-open if not fixed, and thanks for the pr!

PR #310, opened 2022-05-27
 Fix issue #309 - Advance offset by string table field size

track-5:
Fixes issue #309 

There's a bug in master where the code ends up misreading the symbol table because we don't advance the offset prior to reading the strings.

This change fixes the issue by adding the correct value to the offset and also includes a unit test that covers this case.


m4b (approved):
this is great thank you! I think this is a longstanding bug/irritation you've fixed, much appreciated!

m4b:
@track-5 looks like it needs a rustfmt; you can also add `#[rustfmt::skip]` to that test block of bytes, since it doesn't need to be formatted.

track-5:
@m4b my bad! Just ran `cargo fmt` and pushed the new commit :)

m4b:
Looks like you need to use core instead of size for sizeof

track-5:
Done. Ran the cargo build command locally to be sure it works now:

```
P:\goblin>cargo build --no-default-features --features="pe32"
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
```

m4b:

PR #313, opened 2022-06-03
Remove check on debug directory size

lzybkr:
The well intentioned check on debug directory size unfortunately results
in an error when debug information is embedded in the image.

Tools like Digital Mars embed CV debug info in the image instead of
creating a PDB, and some older tools embed COFF debug info in the image.

### DIFF
--- a/src/pe/debug.rs
+++ b/src/pe/debug.rs
@@ -150,8 +150,8 @@ impl<'a> CodeviewPDB70DebugInfo<'a> {
 
         // calculate how long the eventual filename will be, which doubles as a check of the record size
         let filename_length = idd.size_of_data as isize - 24;
-        if filename_length < 0 || filename_length > 1024 {
-            // the record is too short or too long to be plausible
+        if filename_length < 0 {
+            // the record is too short to be plausible
             return Err(error::Error::Malformed(format!(
                 "ImageDebugDirectory size of data seems wrong: {:?}",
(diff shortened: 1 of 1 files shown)


src/pe/debug.rs
```diff
@@ -150,8 +150,8 @@ impl<'a> CodeviewPDB70DebugInfo<'a> {
 
         // calculate how long the eventual filename will be, which doubles as a check of the record size
         let filename_length = idd.size_of_data as isize - 24;
-        if filename_length < 0 || filename_length > 1024 {
-            // the record is too short or too long to be plausible
+        if filename_length < 0 {
```
m4b, line 153: Is there no maximum bound to check against ? (Worried about regressions if removing this)

also curious the number 1024 was used, I wonder why that was chosen ? (Haven’t looked at this code in a while)

lzybkr: 1024 is an odd choice - MAX_PATH + some small constant factor (to include the hash and other data in the section) seems reasonable, but 1024 is a bit bigger than I'd expect for the typical size of the directory.

That said, I don't think there is a maximum bound - and as this crate is typically used, the memory is already mapped/allocated, so it's mostly harmless to remove the check.

m4b (approved):

m4b:
@lzybkr thank you!

m4b:

PR #316, opened 2022-07-07
elf: Handle SHN_XINDEX for strtab section index (#304)

### LINKED ISSUE #304: goblin >= 0.4.2 fails to parse valid ELF Strtab with "bad input invalid utf8"
Starting with goblin 0.4.2, `Strtab::parse` does the following:

```rust
let mut i = 0;
while i < result.bytes.len() {
    let string = get_str(i, result.bytes, result.delim)?;
    result.strings.push((i, string));
    i += string.len() + 1;
}
```

However, it appears that the contents of the strtab in a valid ELF files are NOT always valid UTF-8.  
Some of the strtab entries in my ELF object look like `[F4, 65, 02, 00]`, or `[82, 66, 02, 00]`. This causes `get_str` to fail and the ELF fail cannot be parsed.

The ELF file in question was created by GNU ld, it's a relocatable object. 
<details>
  <summary>Here is the `readelf -h` output for the object. Note the number of section headers.</summary>

```
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 03 00 00 00 00 00 00 00 00
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - GNU
  ABI Version:                       0
  Type:                              REL (Relocatable file)
  Machine:                           Advanced Micro Devices X86-64
  Version:                           0x1
  Entry point address:               0x0
  Start of program headers:          0 (bytes into file)
  Start of section headers:          600695752 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
  Size of program headers:           0 (bytes)
  Number of program headers:         0
  Size of section headers:           64 (bytes)
  Number of section headers:         0 (274069)
  Section header string table index: 65535 (274068)
```

</details>

The first strtab entry (F4, 65, 02) that goblin fails to parse is at this offset into the file:
`000bdc20: f465 0200 81d4 0300 54f5 0200 db28 0400  .e......T....(..`

This corresponds to the following offset starting at number 65535 in the `readelf --sections` output:

```
There are 274069 section headers, starting at offset 0x23cde3c8:

Section Headers:
  [Nr] Name              Type             Address           Offset
       Size              EntSize          Flags  Link  Info  Align
  [ 0]                   NULL             0000000000000000  00000000
       0000000000042e95  0000000000000000          274068     0     0
  [ 1] .group            GROUP            0000000000000000  00000040
       0000000000000010  0000000000000004          197859   266747     4
<... 65k sections snipped ... >
  [65534] .group            GROUP            0000000000000000  000bdc10
       000000000000000c  0000000000000004          197859   258306     4
  [65535] .group            GROUP            0000000000000000  000bdc1c
       0000000000000014  0000000000000004          197859   332560     4
  [65536] .group            GROUP            0000000000000000  000bdc30
       000000000000000c  0000000000000004          197859   315046     4
<...snip...>
```

`readelf -s` shows the following: 

```
<... 65k sections snipped ...>
 65534: 0000000000000000     0 SECTION LOCAL  DEFAULT 65534 .group
 65535: 0000000000000000     0 SECTION LOCAL  DEFAULT 65535 .group
 65536: 0000000000000000     0 SECTION LOCAL  DEFAULT 65536 .group
<...snip...>
```

I don't know this particular corner of the ELF spec very well at all, but I believe something special must be happening when there are more than 2^15 section headers, and the strtab contents may not always be valid UTF-8 strings.

Sometimes this bug fails to reproduce, even though the number of section headers is far above 65k. I believe this is because the 3 bytes of binary data in the strtab entries can *accidentally* happens to be valid UTF-8. This may have been why I didn't run into this bug before today.

I can upload the object file if that would help (though it is from a large statically linked binary compiled in debug and 590M big).

tux3:
This attempts to fix #304.  
With this patch, I no longer reproduce the issue using the same large object file that triggered it before.  

I've also updated the test code to handle XINDEX for the strtab, however I don't know a good way to generate a test case for an ELF file with >=2^16 sections without embedding a very large blob, so I've left it for future work.  
(If I've missed a simple way to do this, happy to do it now)

### DIFF
--- a/src/elf/mod.rs
+++ b/src/elf/mod.rs
@@ -277,7 +277,14 @@ if_sylvan! {
 
             let section_headers = SectionHeader::parse(bytes, header.e_shoff as usize, header.e_shnum as usize, ctx)?;
 
-            let get_strtab = |section_headers: &[SectionHeader], section_idx: usize| {
+            let get_strtab = |section_headers: &[SectionHeader], mut section_idx: usize| {
+                if section_idx == section_header::SHN_XINDEX as usize {
+                    if section_headers.is_empty() {
+                        return Ok(Strtab::default())
+                    }
+                    section_idx = section_headers[0].sh_link as usize;
+                }
+
                 if section_idx >= section_headers.len() {
                     // FIXME: warn! here
                     Ok(Strtab::default())
--- a/tests/elf.rs
+++ b/tests/elf.rs
@@ -1,4 +1,4 @@
-use goblin::elf::section_header::SHT_GNU_HASH;
+use goblin::elf::section_header::{SHN_XINDEX, SHT_GNU_HASH};
 use goblin::elf::sym::{Sym, Symtab};
 use goblin::elf::symver::{VerdefSection, VerneedSection, VersymSection};
 use goblin::elf::Elf;
@@ -51,7 +51,12 @@ fn parse_text_section_size_lazy(base: &[u8]) -> Result<u64, &'static str> {
         SectionHeader::parse(base, header.e_shoff as usize, header.e_shnum as usize, ctx)
             .map_err(|_| "parse section headers error")?;
 
-    let strtab_idx = header.e_shstrndx as usize;
+    let strtab_idx = if header.e_shstrndx as u32 == SHN_XINDEX {
+        obj.section_headers[0].sh_link as usize
+    } else {
+        header.e_shstrndx as usize
+    };
+
     let strtab_shdr = &obj.section_headers[strtab_idx];
     let strtab = Strtab::parse(
         base,


tests/elf.rs
```diff
@@ -51,7 +51,12 @@ fn parse_text_section_size_lazy(base: &[u8]) -> Result<u64, &'static str> {
         SectionHeader::parse(base, header.e_shoff as usize, header.e_shnum as usize, ctx)
             .map_err(|_| "parse section headers error")?;
 
-    let strtab_idx = header.e_shstrndx as usize;
+    let strtab_idx = if header.e_shstrndx as u32 == SHN_XINDEX {
```
m4b, line 54: weird i feel like this was added already, long time ago, surprised to see it now

src/elf/mod.rs
```diff
@@ -277,7 +277,15 @@ if_sylvan! {
 
             let section_headers = SectionHeader::parse(bytes, header.e_shoff as usize, header.e_shnum as usize, ctx)?;
 
-            let get_strtab = |section_headers: &[SectionHeader], section_idx: usize| {
+            let get_strtab = |section_headers: &[SectionHeader], mut section_idx: usize| {
+                if section_idx == section_header::SHN_XINDEX as usize {
+                    if section_headers.is_empty() {
+                        // FIXME: warn! here
```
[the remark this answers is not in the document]
m4b: we could warn, but i'm not sure if elf has any logging; if it doesn't, let's not add it. I'd just remove the fixme

src/elf/mod.rs
```diff
@@ -277,7 +277,15 @@ if_sylvan! {
...
+                        return Ok(Strtab::default())
+                    }
+                    section_idx = section_headers[0].sh_link as usize;
+                }
+
                 if section_idx >= section_headers.len() {
```
m4b, line 289: ok cool we check the bound here

src/elf/mod.rs
```diff
@@ -277,7 +277,15 @@ if_sylvan! {
...
             let section_headers = SectionHeader::parse(bytes, header.e_shoff as usize, header.e_shnum as usize, ctx)?;
 
-            let get_strtab = |section_headers: &[SectionHeader], section_idx: usize| {
+            let get_strtab = |section_headers: &[SectionHeader], mut section_idx: usize| {
+                if section_idx == section_header::SHN_XINDEX as usize {
+                    if section_headers.is_empty() {
+                        // FIXME: warn! here
+                        return Ok(Strtab::default())
+                    }
+                    section_idx = section_headers[0].sh_link as usize;
```
m4b, line 286: just curious (can't remember why) it's always the first section header's sh_link that is used?

m4b (approved):
this LGTM, thanks for adding this, quite surprised wasn't done before! feel free to check about logging, otherwise we're good.

m4b:
And thank you for stepping up and fixing that issue!

src/elf/mod.rs
tux3: It seems to. I actually looked what Gimli was doing, but [the ELF docs at Oracle](https://docs.oracle.com/cd/E23824_01/html/819-0690/chapter6-43405.html#scrolltoc) says:  

>If the section name string table section index is greater than or equal to SHN_LORESERVE (0xff00), this member has the value SHN_XINDEX (0xffff) and the actual index of the section name string table section is contained in the sh_link field of the section header at index 0. Otherwise, the sh_link member of the initial section header entry contains the value zero. See [Table 12-6](https://docs.oracle.com/cd/E23824_01/html/819-0690/chapter6-94076.html#chapter6-12754) and [Table 12-7](https://docs.oracle.com/cd/E23824_01/html/819-0690/chapter6-94076.html#gavyo).

src/elf/mod.rs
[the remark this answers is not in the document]
tux3: Ack, removed the FIXME

m4b:
released in 0.5.3 on crates; thanks for fixing this @tux3!

PR #319, opened 2022-07-29
fix: Iterate all ImageDebugDirectory entries

Swatinem:
Turns out, a PE file can have more than one ImageDebugDirectory.
Thus far, goblin only looks at the very first one, trying to interpret
it as a CV record.

This code changes that logic to rather iterate over all the entries to
find the one that is a CV record.
We still only capture a single entry in the `DebugData` for backwards
compatibility reasons. In the future we could as well capture all of
them.

### DIFF
--- a/src/pe/debug.rs
+++ b/src/pe/debug.rs
@@ -1,4 +1,5 @@
 use crate::error;
+use scroll::ctx::SizeWith;
 use scroll::{Pread, Pwrite, SizeWith};
 
 use crate::pe::data_directories;
@@ -8,6 +9,9 @@ use crate::pe::utils;
 
 #[derive(Debug, PartialEq, Copy, Clone, Default)]
 pub struct DebugData<'a> {
+    // TODO: There can be more than one ImageDebugDirectory.
+    // To avoid breaking the public API here, we just return the ImageDebugDirectory that has the
+    // `IMAGE_DEBUG_TYPE_CODEVIEW` type, or the last one if none has a codeview record.
     pub image_debug_directory: ImageDebugDirectory,
     pub codeview_pdb70_debug_info: Option<CodeviewPDB70DebugInfo<'a>>,
 }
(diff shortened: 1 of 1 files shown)


Swatinem:
fun fact: I wrote a detailed blog post about how I made this discovery: https://swatinem.de/blog/format-ossification/

m4b:
looks like a CI failure?

src/pe/debug.rs
```diff
@@ -35,14 +39,36 @@ impl<'a> DebugData<'a> {
...
+                    image_debug_directory: entries[idx],
+                    codeview_pdb70_debug_info: Some(cv_record),
+                });
+            }
+        }
+        // if we don't have a codeview record, just return the last debug directory
         Ok(DebugData {
-            image_debug_directory,
-            codeview_pdb70_debug_info,
+            image_debug_directory: entries.pop().unwrap(),
```
m4b, line 70: can this unwrap ever fail?

m4b (commented):
Let's fix the CI failure then I'll do another pass at review, and thanks for this!

src/pe/debug.rs
```diff
@@ -35,14 +39,36 @@ impl<'a> DebugData<'a> {
...
+        let len = dd.size as usize;
+        let sizeof_directory = ImageDebugDirectory::size_with(&scroll::LE);
+        let num_entries = len / sizeof_directory;
+
+        let mut entries = (0..num_entries)
+            .map(|_| bytes.gread_with(&mut offset, scroll::LE))
+            .collect::<Result<Vec<ImageDebugDirectory>, _>>()?;
+
+        // find the debug directory that references the codeview record
+        for (idx, idd) in entries.iter().enumerate() {
```
luser, line 60: I know the original code just tries parsing `CodeviewPDB70DebugInfo` directly, but you should additionally be able to look at the `data_type` field to see if it is `IMAGE_DEBUG_TYPE_CODEVIEW` per [the MSDN docs](https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-image_debug_directory).

m4b:
@kkent030315 another PR, let me know if you think this is reasonable, it is a breaking change though, iirc

PR #327, opened 2022-08-30
elf.dynamic: only consider loadable segments for VM translation

lumag:
There is little point in using non-loadable segments for the VM address
translation. Ignore non-loadable segments.


m4b (approved):
I assume/hope this won't cause any semantic breakage (expecting Some(usize)) from someone depending on the behavior before

PR #329, opened 2022-09-07
Fix unbound access to member_offsets in archive/mod.rs:342

anfedotoff:
Hi!
We were doing some fuzzing with libFuzzer and our tool Sydr and found an unbound access to `member_offsets`:
`thread 'main' panicked at 'index out of bounds: the len is 2 but the index is 3', /goblin/src/archive/mod.rs:342:23`

So, I propose this fix for that.


m4b (approved):
Thanks!

m4b:
released in 0.6

PR #330, opened 2022-09-07
Fix unbound access to bytes slice at  pe/debug.rs:172

anfedotoff:
Hi!
We were doing some fuzzing with Sydr+libFuzzer and found this unbound access to bytes slice at  pe/debug.rs:172. Here is the panic message: `thread '<unnamed>' panicked at 'range end index 16256248 out of range for slice of length 4257', library/core/src/slice/index.rs:73`

I propose this fix.


m4b (approved):
Thanks also for this!

m4b:
released in 0.6

Issue #331, opened 2022-09-09
Subtruct with overflow (maybe not dangerous)


anfedotoff:
Hi!
Using our Sydr checkers we found an overflow [here](https://github.com/m4b/goblin/blob/master/src/pe/utils.rs#L58).

`thread '<unnamed>' panicked at 'attempt to subtract with overflow', /goblin/src/pe/utils.rs:58:14`

I had a brief look, it seems not dangerous, what do you think?

m4b:
I'm not sure, since that returns some number someone might use, it could end up not being great? However I'm not sure what it would help by adding a saturating_sub in those additions / subtractions, since the bug would still be there, and isn't clear to me if one number converges to 0 if the other additions will just be as equally bad?

That could would also likely suffer readability with saturating forms, though that is fixed easy enough. it's probably fine i guess? I'd be more curious what a good default "zero" value would be there if we're getting bogus numbers that contribute to it as input.

anfedotoff:
Yes, you are right, readability will suffer if we add some checks. My idea was to stop parsing and return  an error if overflow occured. But if we couldn't  say now that this overflow could lead to some error, I propose to do nothing with it.
As, I said before, goblin is used in our tools (Sydr & Casr). We plan to do some continuous fuzzing for goblin on our side. So, if this overflow will trigger something, we will know and will handle this then. What do you think?

philipc:
The only way I can see that overflowing is if `file_alignment` is not a power of 2. Is that what your Sydr checker found?

Secondly, if it does overflow, it will be a large value, and thus ignored by the following compare. I think this would be better than use `saturating_sub`, because `saturating_sub` would return a size of 0 which doesn't seem right to me.

What this code needs to do is match whatever the Windows loader does, but that might be hard to figure out.

anfedotoff:
> The only way I can see that overflowing is if `file_alignment` is not a power of 2. Is that what your Sydr checker found?

Sydr, somehow could overflow `file_alignment`. Maybe `file_alignment` is 0, it has usize type so there is a big value when we subtract -1. Here is the input for parse fuzz target: 
[crash-sydr_b28e31470d668a648e83ba8d7f76ee43e245e770_int_overflow_17_unsigned.txt](https://github.com/m4b/goblin/files/9558781/crash-sydr_b28e31470d668a648e83ba8d7f76ee43e245e770_int_overflow_17_unsigned.txt)

[...]

PR #332, opened 2022-09-16
Fix corrupted file_alignment

anfedotoff:
Fix for #331


m4b:
@koutheir does this look ok to you?

src/pe/utils.rs
```diff
@@ -97,6 +97,16 @@ pub fn find_offset(
                 section.virtual_address,
                 section.virtual_address + section.virtual_size
             );
+            if file_alignment < 512
```
koutheir, line 100: 1. The usage of `file_alignment` in the implementation of `section_read_size()` only requires it to be a strictly positive power of two. The additional restrictions *promised* by the Windows documentation do not need to be fulfilled in order for `section_read_size()` implementation to be correct.
2. The restrictions regarding `file_alignment` is **in** the implementation of `section_read_size()` so it should be checked there. This would make `section_read_size()` a fallible function, so its prototype and usage will need to account for this.

koutheir (changes requested):

src/pe/utils.rs
m4b: so to clarify, are you suggesting some changes to the patch or it's ok?

anfedotoff: How about that:
```rust
fn section_read_size(section: &section_table::SectionTable, file_alignment: u32) -> Option<usize> {
....
    if file_alignment == 0 || file_alignment & (file_alignment - 1) != 0 {
        return None;
    }
....
}
```

anfedotoff: Another question is, if we check `file_aligment` in section_read_size returning Results/Option we have to change prototypes of these functions:

- is_in_range
- is_in_section
- find_offset

They will return result now, is it ok?

koutheir: That's what I meant by:
> This would make `section_read_size()` a fallible function, so its prototype and usage will need to account for this.

anfedotoff: To be honest, I'm starting to doubt that it is a good idea to check `file_alignment` so deep in code and return an error. There are lot's of places in code (20+) that use custom error messages via ok_or_else.  I don't know how to deal with it properly. Maybe we have to use anyhow crate, or somehow to combine errors from find_offset and custom error.

Maybe we have to check file_offset in place where it constructed or used first time?

koutheir: > To be honest, I'm starting to doubt that it is a good idea to check `file_alignment` so deep in code and return an error.

This is good for local reasoning, and for future maintenance of `section_read_size()`.

> There are lot's of places in code (20+) that use custom error messages via ok_or_else. I don't know how to deal with it properly. Maybe we have to use anyhow crate, or somehow to combine errors from find_offset and custom error.

That might make sense, but it seems excessive for what this PR is trying to fix. This might warrant another PR, though.

> Maybe we have to check file_alignment in place where it constructed or used first time?

This is bad for local reasoning, and for future maintenance of `section_read_size()`.

src/pe/export.rs
```diff
@@ -195,6 +189,7 @@ impl<'a> ExportData<'a> {
             file_alignment,
             opts,
         )
+        .ok()
```
koutheir, line 192: The call to `Ok()` drops the error. Returning the error is a better idea. You can use `Result::and_then()` instead of `Option::and_then()`, or instead use the `?` operator.

src/pe/utils.rs
```diff
@@ -51,6 +53,12 @@ fn section_read_size(section: &section_table::SectionTable, file_alignment: u32)
     // }
 
     let file_alignment = file_alignment as usize;
+    if file_alignment == 0 || file_alignment & (file_alignment - 1) != 0 {
+        return Err(error::Error::Malformed(format!(
+            "file_alignment should be a power of 2 {}",
```
koutheir, line 58: I suggest a more accurate message, such as `"file_alignment={} is not a strictly positive power of 2"`.

src/pe/utils.rs
```diff
@@ -97,48 +106,36 @@ pub fn find_offset(
...
 ) -> error::Result<&'a str> {
-    match find_offset(rva, sections, file_alignment, opts) {
-        Some(offset) => Ok(bytes.pread::<&str>(offset)?),
-        None => Err(error::Error::Malformed(format!(
-            "Cannot find name from rva {:#x} in sections: {:?}",
-            rva, sections
-        ))),
-    }
+    let offset = find_offset(rva, sections, file_alignment, opts)?;
+    Ok(bytes.pread::<&str>(offset)?)
```
koutheir, line 138: A less convoluted form would be:
```rust
bytes.pread::<&str>(offset).map_err(Into::into)
```

koutheir (changes requested):

src/pe/debug.rs
```diff
@@ -100,7 +100,7 @@ impl ImageDebugDirectory {
         opts: &options::ParseOptions,
     ) -> error::Result<Self> {
         let rva = dd.virtual_address as usize;
-        let offset = utils::find_offset(rva, sections, file_alignment, opts).ok_or_else(|| {
+        let offset = utils::find_offset(rva, sections, file_alignment, opts).map_err(|_| {
```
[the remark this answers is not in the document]
anfedotoff: Yes it's true. I did it because I thought than this error message is more representative. Maybe I could combine these two messages using `map_err`? Or just leave what find_offset returns?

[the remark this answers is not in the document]
koutheir: I would simply propagate the error `find_offset()` potentially returns.

It is probably more informative to nest the error of `find_offset()` inside a more generic error that adds context, but that requires something like the `anyhow` crate. It is a good idea to do that, but it's better to fix things one at a time.

m4b:
Some notes:
1. We won't be adding anyhow/thiserror/error library to goblin, probably ever. So if the assumption of these result returning changes is they'll be more usable in a follow up PR that uses more context via anyhow/whatever, want to be clear this is not a good assumption to make :) I want to keep the dependencies as thin as possible.
2. i think the signature of `find_offset` returning Option was fine, and more flexible (allowing users to convert that into a more specific error if need be), for a utils library function (that maybe/probably shouldn't be exposed, but it is useful). I'm 50/50 on the result signature, because that forces our error type onto user (and technically a string allocation) that uses this function directly. I also don't have any data on if anyone is even using this function so also maybe doesn't matter that much. I'm not convinced though that semantically it's an error returning function (just like HashMap::get doesn't return an error, or various `find` methods in stdlib generally don't return error).

So I'm wondering why couldn't all the places find_offset was used with the new Result signature not have used the function find_offset_or or left it with `ok_or` semantics? This is probably annoying question since all the work was done to make it use the result form now, but just curious :) I'm just wondering if this is not the best breaking change, specifically:

1. It is an avoidable break as far as I can see
2. It doesn't really address some necessary change like a field going from T -> Option<T>
3. It doesn't add much in the way of future proofing (this is also similar to 1., in that the user can just convert the option to error if they want)

Anyway, I think it's probably fine if this ends up being better maintainable code, but it is a breaking change, so it'll get merged in a little later (along with the other breaking change that's been pending).

On that note, may be a good time to think of any other breaking changes for PE stuff that would be good to get in (and it's fine if there are none!)

m4b:
@anfedotoff i'd be ok with merging this but i think we have to roll back the breaking change to the `find_offset` function; it seems unnecessary to me. If you want a result in the callsites, I'd suggest using `find_offset_or` and restore the error messages. Then we can merge and it won't be breaking, otherwise I think this PR will just languish.

Thanks for your patience!

m4b:
@anfedotoff Ok that works for me!

anfedotoff:
> @anfedotoff Ok that works for me!

I've done something strange, playing with git history and doing rebase master. I'll better reopen new PR from fresh master. Fix will be very short, sorry)

PR #339, opened 2022-10-29
elf.reloc: add workaround for 0-sized relas

Jhynjhiruu:
Some ELF files include a .rela section with size 0 instead of simply not including the section. This caused issues with the parser, since bytes.pread_with(offset, filesz) would try and read 0 bytes from a 0-length buffer, which returned an error. This workaround simply swaps out the value with the entire input when filesz is 0; this is the smallest change that appears to work, since count will then be set to 0. This could also be worked around by adding a check in elf.mod (or many other methods), but I thought that it was better to allow the parser to handle 0-sized sections.

### DIFF
--- a/src/elf/reloc.rs
+++ b/src/elf/reloc.rs
@@ -416,9 +416,9 @@ if_alloc! {
         pub fn parse(bytes: &'a [u8], offset: usize, filesz: usize, is_rela: bool, ctx: Ctx) -> crate::error::Result<RelocSection<'a>> {
             // TODO: better error message when too large (see symtab implementation)
             let bytes = if filesz != 0 {
-                bytes.pread_with(offset, filesz)?
+                bytes.pread_with::<&'a [u8]>(offset, filesz)?
             } else {
-                bytes
+                &[]
             };
 
             Ok(RelocSection {
--- a/src/elf/reloc.rs
+++ b/src/elf/reloc.rs
@@ -415,7 +415,11 @@ if_alloc! {
         /// Parse a REL or RELA section of size `filesz` from `offset`.
         pub fn parse(bytes: &'a [u8], offset: usize, filesz: usize, is_rela: bool, ctx: Ctx) -> crate::error::Result<RelocSection<'a>> {
             // TODO: better error message when too large (see symtab implementation)
-            let bytes = bytes.pread_with(offset, filesz)?;
+            let bytes = if filesz != 0 {
+                bytes.pread_with(offset, filesz)?
+            } else {
+                bytes
+            };
 
             Ok(RelocSection {
                 bytes: bytes,


src/elf/reloc.rs
```diff
@@ -415,7 +415,11 @@ if_alloc! {
         /// Parse a REL or RELA section of size `filesz` from `offset`.
         pub fn parse(bytes: &'a [u8], offset: usize, filesz: usize, is_rela: bool, ctx: Ctx) -> crate::error::Result<RelocSection<'a>> {
             // TODO: better error message when too large (see symtab implementation)
-            let bytes = bytes.pread_with(offset, filesz)?;
+            let bytes = if filesz != 0 {
+                bytes.pread_with(offset, filesz)?
+            } else {
+                bytes
```
m4b, line 421: wouldn't it be better to return a `&[]` ? (or will that not unify in borrow checker?)

m4b (commented):
just had the one comment, waiting for clarification, otherwise this lgtm, thanks for this fix!

src/elf/reloc.rs
Jhynjhiruu: No; see: https://github.com/m4b/goblin/blob/master/src/elf/mod.rs#L371
Unless I'm misunderstanding how `push` works in Rust (very possible tbh), returning &[] wouldn't work since it would then be pushing an invalid relocation object into the array here. My original fix was adding a 0-size check and continuing the loop.

m4b: so the bytes are returned so that we can parse out relocs lazily, via this method or iterator: https://github.com//m4b/goblin/blob/d035559b991e631eae7aa91d89cde5e5e03e9cc6/src/elf/reloc.rs#L435

in the patch, you return the entire bytes passed in (which is basically the elf binary); probably this won't cause any issues, since get method checks index >= count (and index will always be >= 0), but i think it makes more sense to return an empty byte slice, since the reloc section is empty, and this shows reader that we have an empty reloc section.

Jhynjhiruu: Oh, right, yep, didn't read what you wrote properly. Assuming that works, that would definitely be better.

m4b: :) i played around with it, should work; i had to manually annotate the the empty slice though, something like:

```
let reloc_bytes: &'a [u8] = &[];
bytes
```

if the `filesz == 0` branch; or do a mutable empty slice with that annotation and assign the pread result if `filesz` is not 0, your choice :)

semi unrelated: i noticed rustfmt is totally broken for me when i add an if (i assume because it's inside `if_alloc!` macro, which we've had trouble with in past.

Jhynjhiruu: Looks good to me. Do you want to edit the PR? I'm very new to Rust so I'm not entirely confident about contributing code to a project like this.

m4b: You should go for it :) i believe in you! Your fix was already good, i believe if you just modify what you have and write:

```rust
} else {
  let bytes: &'a [u8] = &{];
  bytes
}
```
it should work.

This is a very rust borrowchecker thing, and i think scroll library (provides the `pread` call) is interacting badly here.

So pread requires a generic param typically, unless the variable it gets bound is clearly type inferrable.

Specifically, if you write:

```rust
let x = bytes.pread::<u16>(0)?
```
it's same as;
```rust
let x: u16 = bytes.pread(0)?;
```

so there's a weird error (for me at least) when i attempted to just pass an empty slice, i think it was trying to type infer the bytes as an array or something. (remember, in an if the two different arms types must match/be identical)

it may also work telling pread explicitly to do:

```rust
if filesiz != 0 {
  bytes.pread::<&'a [u8]>(offfset, filesz)?
} else {
  &[]
}
```
haven't tried that :)

anyway, rambling a bit!

Jhynjhiruu: I implemented it with the turbofish (turbo-fish? turbo_fish? turbo::<fish>?) and `bytes.pread_with`, since I prefer that. It works my end, so we'll see what CI has to say about it.

m4b (approved):
LGTM, thank you!!

PR #342, opened 2022-11-07
mach: Implement `LC_NOTE`

messense:


src/mach/load_command.rs
```diff
@@ -1431,6 +1447,7 @@ pub enum CommandVariant {
     VersionMinWatchos(VersionMinCommand),
     DyldExportsTrie(LinkeditDataCommand),
     DyldChainedFixups(LinkeditDataCommand),
+    Note(NoteCommand),
```
m4b, line 1450: I'm realizing this and previous PR of yours are technically breaking changes, glad we got these in now; perhaps would be prudent to add the tag on the enum that it's incomplete (in case a new load command is added and we don't want to add a breaking change again?)

m4b (commented):
thank you, this is good stuff! Could you add the enum attribute that makes additions not a breaking change, and then since this is breaking, if it's ok I'd like to wait a bit and roll up other breaking changes.
on that note, it might be prudent to survey other enums in this library to add this attribute to so that we can start thinking about 1.0 release

messense:
Thanks for the review, added `#[non_exhaustive]` to `CommandVariant`.

m4b (approved):
this is great thanks so much!

m4b:
NB for m4b: breaking change

m4b:

PR #347, opened 2022-11-15
elf.section_header: additional workaround for 0-length sections

Jhynjhiruu:
One more change necessary for the funky ELFs I have that I forgot to add previously is allowing check_size to immediately return Ok(()) on a 0-sized section, even if that section isn't marked as SHT_NOBITS.


m4b (approved):
LGTM, thanks!

PR #351, opened 2023-01-01
Add AFL++ fuzzing support

anfedotoff:
Hi! 
I think adding AFL++ fuzzing support  it's a good idea for finding new bugs. What do you think?

### DIFF
--- a/fuzz-afl/fuzz_targets/afl_parse_elf.rs
+++ b/fuzz-afl/fuzz_targets/afl_parse_elf.rs
@@ -1,0 +1,26 @@
+#[macro_use]
+extern crate afl;
+
+fn main() {
+    fuzz!(|data: &[u8]| {
+        if let Ok(elf) = goblin::elf::Elf::parse(data) {
+            for section_header in &elf.section_headers {
+                let _ = elf.shdr_strtab.get_at(section_header.sh_name);
+            }
+
+            for _relocation in &elf.dynrels {}
+
+            if let Some(mut it) = elf.iter_note_headers(data) {
+                while let Some(Ok(_a)) = it.next() {}
+            }
+
+            if let Some(mut it) = elf.iter_note_sections(data, None) {
+                while let Some(Ok(_a)) = it.next() {}
+            }
+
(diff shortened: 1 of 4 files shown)


m4b:
well this is annoying looks like our macos CI test is failing (unrelated to your PR) :/

fuzz-afl/fuzz_targets/afl_parse_elf.rs
```diff
@@ -0,0 +1,26 @@
+#[macro_use]
+extern crate afl;
+
+fn main() {
+    fuzz!(|data: &[u8]| {
+        if let Ok(elf) = goblin::elf::Elf::parse(data) {
+            for section_header in &elf.section_headers {
+                let _ = elf.shdr_strtab.get(section_header.sh_name);
```
cgzones, line 8: Maybe use [`black_box()`](https://doc.rust-lang.org/std/hint/fn.black_box.html) to avoid the compiler optimizing everything away?

anfedotoff: goblin
I build with `cargo afl build --release` this line is still here, according to IDA. You helped me to notice this:

    warning: use of deprecated associated function `goblin::strtab::Strtab::<'a>::get`: Bad performance, use get_at() instead

 I think we should change `get` method to `get_at`.

m4b:
this seems great! (sorry for the delay)

this doesn't add any dependencies to goblin for users right? This is just a crate that is in the project directory, that we can incidentally run fuzzing on, yes?

anfedotoff:
> this seems great! (sorry for the delay)
> 
> this doesn't add any dependencies to goblin for users right? This is just a crate that is in the project directory, that we can incidentally run fuzzing on, yes?

Yes it doesn't add any new dependencies. It's like cargo fuzz, but cargo afl with AFL++ inside. You could look at [image-rs](https://github.com/image-rs/image) which have both libFuzzer and AFL++ support.

anfedotoff:
@m4b, so what do you think, do we need AFL++ fuzzing support?

m4b (approved):

m4b:

PR #352, opened 2023-01-26
elf: Add basic GNU PROPERTY note support

x64k:
Signed-off-by: Alexandru Lazar <alazar@startmail.com>


x64k:
Probably unrelated to this PR -- I wanted to look at the macos tests and see why they're failing (I figured it was unrelated to this change but you can never be too sure). It seems that at some point `dyldinfo` got a facelift and was reborn as `dyld_info`, with an incompatible CLI interface, so tests that rely on `dyldinfo` output now bork.

Sadly all I know about dyld is that it's completely unrelated to Bob Dylan so I'm hardly qualified to fix this, but I figured I'd mention it to save braver (and more Mac-minded) souls thirty seconds worth' of stack tracing.

m4b (approved):
this is great, thank you!

m4b:
would be really great to get the macos tests fixed; i will have to disable them seems like for moment

m4b:
@x64k thank you for the PR! do you need a release soon or can you wait a bit? (i like to roll up these changes to reduce churn)

m4b:
also thanks for pointing out the dyldinfo failure; i just disabled the test until can figure out or someone comes along to match the new dyld_info api to what we output

x64k:
@m4b hey, thanks! There are absolutely no constraints on my end as far as releases are concerned, please don't worry about it.

PR #359, opened 2023-03-10
pe: fix certificate table parsing

baloo:
There appear to be an off-by-one error in the parsing of the certificate table. This error would occur when the cert table is at the end of the binary.

cc @RaitoBezarius


RaitoBezarius (approved):
Yes, I was definitely a bit overzealous about getting out of index. :>

m4b (approved):
LGTM thank you

m4b:

PR #362, opened 2023-03-13
pe: add authenticode support / v2

baloo:
Alternative to #358

This removes the `digest` dependency and allows to implement hashing like:
```rust
trait AuthenticodeDigest {
    /// [`authenticode_digest`] returns the result of the provided hash algorithm.
    fn authenticode_digest<D: Digest>(&self) -> Output<D>;
}

impl AuthenticodeDigest for PE<'_> {
    fn authenticode_digest<D: Digest>(&self) -> Output<D> {
        let mut digest = D::new();

        for chunk in self.authenticode_ranges() {
            digest.update(chunk);
        }

        digest.finalize()
    }
}


fn main() {
    let mut buf = Vec::new();
    let mut f =
        File::open("/nix/store/bhsxra1hc7yhja2kzw5rdds90i3w3a54-linux-5.10.147/bzImage").unwrap();
    f.read_to_end(&mut buf).unwrap();

    let pe = PE::parse(&buf).unwrap();

    let hash = pe.authenticode_digest::<Sha256>();
    println!("hash: {:x?}", hash);
}
```

### DIFF
--- a/src/pe/authenticode.rs
+++ b/src/pe/authenticode.rs
@@ -1,0 +1,116 @@
+// Reference:
+//   https://learn.microsoft.com/en-us/windows-hardware/drivers/install/authenticode
+//   https://download.microsoft.com/download/9/c/5/9c5b2167-8017-4bae-9fde-d599bac8184a/Authenticode_PE.docx
+
+// Authenticode works by omiting sections of the PE binary from the digest
+// those sections are:
+//   - checksum
+//   - data directory entry for certtable
+//   - certtable
+
+use core::ops::Range;
+
+use super::PE;
+
+impl PE<'_> {
+    /// [`authenticode_ranges`] returns the various ranges of the binary that are relevant for
+    /// signature.
+    pub fn authenticode_ranges(&self) -> ExcludedSectionsIter<'_> {
+        ExcludedSectionsIter {
+            pe: self,
+            state: IterState::default(),
+        }
+    }
+}
+
+/// [`ExcludedSections`] holds the various ranges of the binary that are expected to be
+/// excluded from the authenticode computation.
+#[derive(Debug, Clone, Default)]
+pub(super) struct ExcludedSections {
+    checksum: Range<usize>,
+    datadir_entry_certtable: Range<usize>,
+    certtable: Option<Range<usize>>,
+}
+
+impl ExcludedSections {
+    pub(super) fn new(
+        checksum: Range<usize>,
+        datadir_entry_certtable: Range<usize>,
+        certtable: Option<Range<usize>>,
+    ) -> Self {
+        Self {
+            checksum,
+            datadir_entry_certtable,
+            certtable,
+        }
+    }
+}
+
+pub struct ExcludedSectionsIter<'s> {
+    pe: &'s PE<'s>,
+    state: IterState,
+}
+
+#[derive(Debug, PartialEq)]
+enum IterState {
+    Initial,
+    DatadirEntry(usize),
+    CertTable(usize),
+    Final(usize),
+    Done,
+}
+
+impl Default for IterState {
+    fn default() -> Self {
+        Self::Initial
+    }
+}
+
+impl<'s> Iterator for ExcludedSectionsIter<'s> {
+    type Item = &'s [u8];
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let bytes = &self.pe.bytes;
+
+        if let Some(sections) = self.pe.authenticode_excluded_sections.as_ref() {
+            loop {
+                match self.state {
+                    IterState::Initial => {
+                        self.state = IterState::DatadirEntry(sections.checksum.end);
+                        return Some(&bytes[..sections.checksum.start]);
+                    }
+                    IterState::DatadirEntry(start) => {
+                        self.state = IterState::CertTable(sections.datadir_entry_certtable.end);
+                        return Some(&bytes[start..sections.datadir_entry_certtable.start]);
+                    }
+                    IterState::CertTable(start) => {
+                        if let Some(certtable) = sections.certtable.as_ref() {
+                            self.state = IterState::Final(certtable.end);
+                            return Some(&bytes[start..certtable.start]);
+                        } else {
+                            self.state = IterState::Final(start)
+                        }
+                    }
+                    IterState::Final(start) => {
+                        self.state = IterState::Done;
+                        return Some(&bytes[start..]);
+                    }
+                    IterState::Done => return None,
+                }
+            }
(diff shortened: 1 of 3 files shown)


src/pe/mod.rs
```diff
@@ -190,9 +243,25 @@ impl<'a> PE<'a> {
...
+                }
+            }
+
+            #[cfg(feature = "pe_authenticode")]
+            {
+                excluded_authenticode_sections = Some(excluded_sections);
             }
         }
         Ok(PE {
+            #[cfg(feature = "pe_source")]
```
m4b, line 261: this does not need to be a cfg; PE already has lifetime tied to the bytes, so storing it in pe is fine (we do the same for mach-o).

src/pe/mod.rs
```diff
@@ -190,9 +243,25 @@ impl<'a> PE<'a> {
...
+            bytes,
+            #[cfg(feature = "pe_authenticode")]
```
m4b, line 263: in general i do not like this pattern in rust code, where the struct size changes according to cfg flags; if we merge something like this would very much prefer/require no cfg flag here (and we already have too many cfg flags)

src/pe/authenticode.rs
```diff
@@ -0,0 +1,117 @@
...
+
+        out.into()
+    }
+}
+
+/// [`ExcludedAuthenticodeSections`] holds the various ranges of the binary that are expected to be
+/// excluded from the authenticode computation.
+#[derive(Debug, Clone, Default)]
+pub(super) struct ExcludedAuthenticodeSections {
+    pub checksum: Range<usize>,
```
m4b, line 50: i think you can remove these pubs if it is pub(super) (only this mod needs to know about these fields afaics?)
You may actually even be able to remove the `pub(super)` if you move the constructor/fn new here

src/pe/mod.rs
```diff
@@ -29,6 +31,11 @@ use log::debug;
 #[derive(Debug)]
 /// An analyzed PE32/PE32+ binary
 pub struct PE<'a> {
+    #[cfg(feature = "pe_source")]
```
m4b, line 34: remove

src/pe/mod.rs
```diff
@@ -92,6 +104,47 @@ impl<'a> PE<'a> {
         let mut certificates = Default::default();
         let mut is_64 = false;
         if let Some(optional_header) = header.optional_header {
+            #[cfg(feature = "pe_authenticode")]
```
m4b, line 107: remove

src/pe/mod.rs
```diff
@@ -92,6 +104,47 @@ impl<'a> PE<'a> {
...
+                            )
+                        }
+                        magic => {
+                            return Err(error::Error::Malformed(format!(
+                                "Unsupported header magic ({magic:#x})"
+                            )))
+                        }
+                    };
+
+                authenticode::ExcludedAuthenticodeSections {
```
m4b, line 141: i think you should extract this into section into a `pub(super) fn new()` for `authenticode::ExcludedAuthenticodeSections`

src/pe/mod.rs
```diff
@@ -190,9 +243,25 @@ impl<'a> PE<'a> {
...
+
+                #[cfg(feature = "pe_authenticode")]
+                {
+                    let start = certificate_table.virtual_address as usize;
+                    let end = start + certificate_table.size as usize;
+                    excluded_sections.certtable = Some(start..end);
+                }
+            }
+
+            #[cfg(feature = "pe_authenticode")]
```
m4b, line 255: remove

src/pe/mod.rs
```diff
@@ -5,6 +5,8 @@
 
 use alloc::vec::Vec;
 
+#[cfg(feature = "pe_authenticode")]
```
m4b, line 8: remove

src/pe/authenticode.rs
```diff
@@ -0,0 +1,117 @@
...
+        ExcludedAuthenticodeSectionsIter {
+            pe: self,
+            state: IterState::default(),
+        }
+    }
+
+    /// [`authenticode_slice`] is intended for convenience when signing a binary with a PKCS#11
+    /// interface (HSM interface).
+    /// Some algorithms (RSA-PKCS at least) for signature require the non-prehashed slice to be provided.
+    pub fn authenticode_slice(&self) -> Box<[u8]> {
```
m4b, line 29: is this really useful to have here? I hesitate to add `pub` functions since they become maintenance burden/have to be watched for breaking changes, etc.

src/pe/authenticode.rs
```diff
@@ -0,0 +1,117 @@
...
+
+use alloc::{boxed::Box, vec::Vec};
+use core::ops::Range;
+
+use super::PE;
+
+impl PE<'_> {
+    /// [`authenticode_ranges`] returns the various ranges of the binary that are relevant for
+    /// signature.
+    pub fn authenticode_ranges(&self) -> ExcludedAuthenticodeSectionsIter<'_> {
```
m4b, line 19: better to move this to `mod.rs` i think.

Cargo.toml
```diff
@@ -50,6 +50,8 @@ mach64 = ["alloc", "endian_fd", "archive"]
 pe32 = ["alloc", "endian_fd"]
 pe64 = ["alloc", "endian_fd"]
 archive = ["alloc"]
+pe_source = []
+pe_authenticode = ["pe_source"]
```
m4b, line 54: delete :)

m4b (changes requested):
this is much more palatable, thank you for making these changes so quickly!
We don't need these feature flags, as I mentioned we also tuck the mach-o bytes inside the struct, so this isn't an unusual pattern already. Otherwise just some minor things/changes I noted, but this looks ok to me :)

src/pe/authenticode.rs
baloo: If I do so, then I need to provide a constructor for `ExcludedAuthenticodeSectionsIter`. I kind of like this ability to split `impl` in multiple submodules, but it's up to you.

src/pe/authenticode.rs
baloo: correct

src/pe/authenticode.rs
m4b: ok yes i agree

baloo:
anything missing or can we merge as this?

m4b (approved):
Thank you!

m4b:
This is great @baloo sorry for delay and thanks for making these changes!

m4b:
non-breaking

m4b:

PR #366, opened 2023-04-12
Fix `is_lib` detection for PIE and static PIE Elf objects.

m-hilgendorf:
PIE executables (either static or dynamically linked) erroneously report `is_lib == true`, due to the fact that they have `e_type == ET_DYN` in their elf headers. 

To distinguish you have to check if the DT_FLAGS_1 was set in the dynamics section. 

The part that I'm not sure about is `lazy_parse`, which can't detect if something is a library or not.

### DIFF
--- a/src/elf/mod.rs
+++ b/src/elf/mod.rs
@@ -383,7 +383,7 @@ if_sylvan! {
             let verdef = symver::VerdefSection::parse(bytes, &section_headers, ctx)?;
             let verneed = symver::VerneedSection::parse(bytes, &section_headers, ctx)?;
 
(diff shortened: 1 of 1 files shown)


m4b (approved):
LGTM, thanks! lazy parsing is up to user to fill in rest of fields, i'm not sure what we do right now for is_lib?

m4b:
non-breaking

m-hilgendorf:
I think right now what you do is check the `e_type` of the Elf header for `ET_DYN`. However, `ET_DYN` only implies the object is dynamic, not if it is a library. The field name is a bit misleading in that sense, it might make more to expose the type directly through an enum (including the other variants of e_type).

m4b:

PR #377, opened 2023-08-24
Add `ParseOptions` field to toggle parsing Attribute Certificates for PEs loaded into memory

### LINKED ISSUE #376: Cannot parse loaded image due to rigidity around parsing Attribute Certificates
Currently, there is no option to skip parsing of attribute certificates, which prevents parsing loaded images because [the loader does not include these certificates](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#other-contents-of-the-file) (just before "Section Data"). Per the linked docs:

> Another exception is that attribute certificate and debug information must be placed at the very end of an image file, with the attribute certificate table immediately preceding the debug section, because the loader does not map these into memory. 

Related: #312. I thought that issue was too general for this concern, so I made a new one. I haven't run into issues with the address resolution -- I'm using the `resolve_rva` field of `ParseOptions` to deal with that; however, forced parsing of the attribute certificate table is blocking for the emulation-based fuzzing project I work on, where we need to parse PEs straight from their in-memory representation.

suttonbradley:
Per [this documentation](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#other-contents-of-the-file) (just above "Section Data"), the attribute certificate table is not loaded into memory along with the rest of the PE. This PR fixes #376 by adding a `ParseOptions` field that allows a user to skip parsing the attribute certificate table altogether, making parsing of loaded PEs possible.

### DIFF
--- a/src/pe/options.rs
+++ b/src/pe/options.rs
@@ -3,11 +3,22 @@
 pub struct ParseOptions {
     /// Wether the parser should resolve rvas or not. Default: true
     pub resolve_rva: bool,
+    /// Whether or not to parse attribute certificates.
+    /// Set to false for in-memory representation, as the [loader does not map this info into
+    /// memory](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#other-contents-of-the-file).
+    /// For on-disk representations, leave as true.
+    /// Default: true
+    #[cfg(feature = "in_memory")]
+    pub parse_attribute_certificates: bool,
 }
 
 impl ParseOptions {
     /// Returns a parse options structure with default values
     pub fn default() -> Self {
-        ParseOptions { resolve_rva: true }
+        ParseOptions {
+            resolve_rva: true,
+            #[cfg(feature = "in_memory")]
+            parse_attribute_certificates: true,
+        }
     }
 }
--- a/src/pe/options.rs
+++ b/src/pe/options.rs
@@ -6,9 +6,7 @@ pub struct ParseOptions {
     /// Whether or not to parse attribute certificates.
     /// Set to false for in-memory representation, as the [loader does not map this info into
     /// memory](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#other-contents-of-the-file).
-    /// For on-disk representations, leave as true.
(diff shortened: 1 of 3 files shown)


suttonbradley:
I'm open to opinions here design-wise -- in order to avoid a major version bump, I feature-gated the new `ParseOptions` field. Also open to opinions on naming, as I'll prioritize speed here over anything else.

m4b:
Since there are breaking changes coming up I think the design of this should ignore that as a downside; I don't think a feature flag for this is the best approach, and would prefer it to work programatically instead, if you're still up for fixing it :)

suttonbradley:
Makes sense, changed it accordingly. Let me know what else I can do here to get it merged!

src/pe/options.rs
```diff
@@ -3,11 +3,19 @@
 pub struct ParseOptions {
     /// Wether the parser should resolve rvas or not. Default: true
     pub resolve_rva: bool,
+    /// Whether or not to parse attribute certificates.
+    /// Set to false for in-memory representation, as the [loader does not map this info into
+    /// memory](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#other-contents-of-the-file).
+    /// For on-disk representations, leave as true. Default: true
+    pub parse_attribute_certificates: bool,
```
nightmared, line 10: Nit: it would perhaps be slightly better to have the fields in ascending alphanumeric order (that is, `parse_attribute_certificates` before `resolve_rva`).

nightmared:
Hello, just wanted to drop by and say that this proved useful to me, for parsing a binary loaded in memory inside an UEFI environment.

If I were very fussy, I would say that perhaps you could add a comment to the description of the field `parse_attribute_certificates` to warn users that as a consequence of disabling that flag, `authenticode_excluded_sections` will be incomplete.

Except for the 6 commits that could probably be squashed into a single one, this looks good to me.

m4b (approved):
If you think adding the comment as @nightmared suggested is ok, then go ahead otherwise I’ll merge this in a few hours

m4b:
released in 0.8.0, happy new year!

PR #379, opened 2023-09-15
pe: add COFF object file support

### LINKED ISSUE #324: Support for COFF object files
COFF object files are not supported. The function [goblin::peek()](https://docs.rs/goblin/latest/goblin/fn.peek.html) doesn't seem to support this COFF format.

This lack of support causes the [issue #36 in `bingrep`](https://github.com/m4b/bingrep/issues/36), reproducible as follows:
```bash
$ rustup target add x86_64-pc-windows-gnu
$ echo "" | rustc --target x86_64-pc-windows-gnu --emit obj - --crate-type lib
$ bingrep rust_out.o
Unknown magic: 0x38664
```

Whereas:
```bash
$ file rust_out.o
rust_out.o: Intel amd64 COFF object file, not stripped, 3 sections, symbol offset=0x8c, 10 symbols, 1st section name ".text"
```

vadimcn:
Fixes  #324

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -42,6 +42,9 @@
 //!                 Object::PE(pe) => {
 //!                     println!("pe: {:#?}", &pe);
 //!                 },
+//!                 Object::COFF(coff) => {
+//!                     println!("coff: {:#?}", &coff);
+//!                 },
 //!                 Object::Mach(mach) => {
 //!                     println!("mach: {:#?}", &mach);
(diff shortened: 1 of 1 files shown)


m4b (approved):
Thank you! I'm wondering if we should add non-exhaustive to the Hint enum, but this is fine for now (and breaking change either way)

m4b:
NB breaking change

Issue #274, opened 2021-05-28
MIPS64 parse error: "type is too big (1236271128) for 137416"


Anniywell:
[Error file](https://github.com/m4b/goblin/files/6559897/mips64.txt)
```
target_mips64: ELF 64-bit LSB pie executable, MIPS, MIPS64 rel2 version 1 (SYSV), dynamically linked, interpreter /lib64/ld.so.1, BuildID[sha1]=bacc6abf4a4687c897f7b49b78a9fcbb710a107c, for GNU/Linux 3.2.0, with debug_info, not stripped
```
This is my code:
```rust
use goblin::{error, Object};
use std::path::Path;
use std::fs;
// use std::io;

fn run () -> error::Result<()> {
    let path = Path::new("./mips64");
    let buffer = fs::read(path)?;
    match Object::parse(&buffer)? {
        Object::Elf(elf) => {
            println!("elf: {:#?}", &elf);
        },
        Object::PE(pe) => {
            println!("pe: {:#?}", &pe);
        },
        Object::Mach(mach) => {
            println!("mach: {:#?}", &mach);
        },
        Object::Archive(archive) => {
            println!("archive: {:#?}", &archive);
        },
        Object::Unknown(magic) => { println!("unknown magic: {:#x}", magic) }
    }
    Ok(())
}

fn main() {
    let r = run();
    let _ = match r {
        Ok(a) => a,
        Err(error) => {
            panic!("{:?}", error.to_string());
        }
    };
   }
```

m4b:
confirmed, thanks for uploading the file! likely something in section headers or program headers says something has a larger size (1.2GB) than the binary itself. It could be zero space issue, though I thought we'd fixed that, or something else is going on. @Anniywell would you like to investigate? A simple way to test, would be to e.g., add some prints into the elf backend then run this:
```
cargo run --example rdr mips64.txt
```

So i suspect something is wrong with this calculation:
```rust
                let max_reloc_sym = dynrelas.iter()
                    .chain(dynrels.iter())
                    .chain(pltrelocs.iter())
                    .fold(0, |num, reloc| cmp::max(num, reloc.r_sym));
                if max_reloc_sym != 0 {
                    num_syms = cmp::max(num_syms, max_reloc_sym + 1);
                }
```
which reports there to be `51511296`  51 million symbols, which doesn't seem right :)

philipc:
[...]

PR #382, opened 2023-10-22
elf.reloc: fix mips64 parse error

wynemo:
https://github.com/m4b/goblin/issues/274

### DIFF
--- a/src/elf/reloc.rs
+++ b/src/elf/reloc.rs
@@ -253,14 +253,56 @@ pub mod reloc64 {
     pub const SIZEOF_RELA: usize = 8 + 8 + 8;
     pub const SIZEOF_REL: usize = 8 + 8;
 
+    #[cfg(not(all(
+        target_os = "linux",
+        any(target_arch = "mips64le", target_arch = "mips64")
+    )))]
     #[inline(always)]
     pub fn r_sym(info: u64) -> u32 {
         (info >> 32) as u32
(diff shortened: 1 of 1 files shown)


src/elf/reloc.rs
```diff
@@ -253,16 +253,58 @@ pub mod reloc64 {
     pub const SIZEOF_RELA: usize = 8 + 8 + 8;
     pub const SIZEOF_REL: usize = 8 + 8;
 
+    #[cfg(not(all(
```
m4b, line 256: Why would these bit masking helper functions be cfg’d out on mips systems?

m4b (commented):
Need to understand why these are cfgd out

src/elf/reloc.rs
m4b: Ah I see further down is the other condition. So I have to reread the linked thread but is the issue with how mips itself does some bit masking here or if the binary is from a mips arch ? If the latter a cfg is not appropriate as we should be able to read a mips elf binary on a whatever machine.

m4b (changes requested):
Unfortunately I think the correct solution here is to deprecate the older functions and pass down whether it’s a mips binary to these functions, with a conditional inside the new functions where you have cfgs now basically

m4b:
closing, my original suggestion is probably what we should do

PR #383, opened 2023-12-08
pe: make sure authenticode is identical before/after signature

baloo:
When adding the signature, the last section of the file will be padded to 8-bytes align. We need to make sure the payload we feed to a signer is always padded to 8-bytes.

This fixes signature breakage.

cc @RaitoBezarius 
Thanks to @foxboron for the heads up!

### DIFF
--- a/src/pe/authenticode.rs
+++ b/src/pe/authenticode.rs
@@ -154,6 +154,10 @@ impl<'s> Iterator for ExcludedSectionsIter<'s> {
                         //    image. The NumberOfSections field of COFF File Header indicates how big
                         //    the table should be. Do not include any section headers in the table whose
                         //    SizeOfRawData field is zero.
+
+                        // Implementation detail:
+                        // We require allocation here because the section table has a variable size and
+                        // needs to be sorted.
                         let mut sections: VecDeque<SectionTable> = self
                             .pe
                             .sections
--- a/src/pe/authenticode.rs
+++ b/src/pe/authenticode.rs
@@ -12,6 +12,8 @@ use core::ops::Range;
 
 use super::PE;
 
+static PADDING: [u8; 7] = [0; 7];
+
 impl PE<'_> {
     /// [`authenticode_ranges`] returns the various ranges of the binary that are relevant for
     /// signature.
@@ -57,6 +59,7 @@ enum IterState {
     DatadirEntry(usize),
     CertTable(usize),
     Final(usize),
+    Padding(usize),
     Done,
 }
 
@@ -92,8 +95,17 @@ impl<'s> Iterator for ExcludedSectionsIter<'s> {
                         }
                     }
                     IterState::Final(start) => {
+                        let buf = &bytes[start..];
+                        self.state = IterState::Padding(buf.len());
+                        return Some(buf);
+                    }
+                    IterState::Padding(hash_size) => {
                         self.state = IterState::Done;
-                        return Some(&bytes[start..]);
+
+                        if hash_size % 8 != 0 {
+                            let pad_size = 8 - hash_size % 8;
+                            return Some(&PADDING[..pad_size]);
+                        }
                     }
                     IterState::Done => return None,
                 }
--- a/src/pe/authenticode.rs
+++ b/src/pe/authenticode.rs
@@ -8,9 +8,11 @@
 //   - data directory entry for certtable
 //   - certtable
 
-use core::ops::Range;
+use alloc::collections::VecDeque;
+use core::{mem, ops::Range};
+use log::debug;
 
-use super::PE;
+use super::{section_table::SectionTable, PE};
 
 static PADDING: [u8; 7] = [0; 7];
 
@@ -31,19 +33,22 @@ impl PE<'_> {
 pub(super) struct ExcludedSections {
     checksum: Range<usize>,
     datadir_entry_certtable: Range<usize>,
-    certtable: Option<Range<usize>>,
+    certificate_table_size: usize,
+    end_image_header: usize,
 }
 
 impl ExcludedSections {
     pub(super) fn new(
         checksum: Range<usize>,
         datadir_entry_certtable: Range<usize>,
-        certtable: Option<Range<usize>>,
+        certificate_table_size: usize,
+        end_image_header: usize,
     ) -> Self {
         Self {
             checksum,
             datadir_entry_certtable,
-            certtable,
+            certificate_table_size,
+            end_image_header,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


baloo:
I ended up re implementing the authenticode parser according to the specification. I think it's better that way, although a little bit more complex.

src/pe/authenticode.rs
```diff
@@ -74,28 +95,164 @@ impl<'s> Iterator for ExcludedSectionsIter<'s> {
 
         if let Some(sections) = self.pe.authenticode_excluded_sections.as_ref() {
             loop {
-                match self.state {
+                match mem::replace(&mut self.state, IterState::Pending) {
```
baloo, line 98: The only reason I need to mem::replace (and make data moves) is because I'm consuming the sections progressively and it's a member of `IterState`.

I could avoid that be making `sections: VecDeque<_>` a member of `ExcludedSectionsIter` instead and just mutate it.

src/pe/authenticode.rs
```diff
@@ -76,24 +96,155 @@ impl<'s> Iterator for ExcludedSectionsIter<'s> {
...
+                        //       (File Size) - ((Size of AttributeCertificateTable) + SUM_OF_BYTES_HASHED)
+                        //
+                        // Note: The size of Attribute Certificate Table is specified in the second
+                        //       ULONG value in the Certificate Table entry (32 bit: offset 132,
+                        //       64 bit: offset 148) in Optional Header Data Directories.
+                        let file_size = bytes.len();
+                        if file_size > sum_of_bytes_hashed {
+                            let extra_data_start = sum_of_bytes_hashed;
+                            let len =
+                                file_size - sections.certificate_table_size - sum_of_bytes_hashed;
```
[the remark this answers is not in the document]
baloo: The spec calls for:
```
(File Size) - ((Size of AttributeCertificateTable) + SUM_OF_BYTES_HASHED)
```

I can go with `file_size - (sections.certificate_table_size + sum_of_bytes_hashed)`, but you suggestion is wrong I think.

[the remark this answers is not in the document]
Foxboron: It's probably wrong :) I thought it was more important to point out the error.

src/pe/authenticode.rs
```diff
@@ -29,34 +34,49 @@ impl PE<'_> {
...
+            certificate_table_size,
+            end_image_header,
         }
     }
 }
 
 pub struct ExcludedSectionsIter<'s> {
     pe: &'s PE<'s>,
     state: IterState,
+    sections: VecDeque<SectionTable>,
```
m4b, line 60: doesn't this mean the iterator allocates every time it's created? this is a little unusual for iterator implementations (most people expect the allocation to occur at their own choosing, e.g., from `collect::<Vec<_>>()` or whatever

src/pe/authenticode.rs
```diff
@@ -76,24 +96,155 @@ impl<'s> Iterator for ExcludedSectionsIter<'s> {
...
+                    }
+                    IterState::HeaderEnd {
+                        end_image_header,
+                        sum_of_bytes_hashed,
+                    } => {
+                        // 9. Build a temporary table of pointers to all of the section headers in the
+                        //    image. The NumberOfSections field of COFF File Header indicates how big
+                        //    the table should be. Do not include any section headers in the table whose
+                        //    SizeOfRawData field is zero.
+                        let mut sections: VecDeque<SectionTable> = self
```
m4b, line 157: i'd not expect allocations in the iterator itself; (same comment as above); is there anyway to avoid allocating? (if there's not, or it's too difficult, that's fine, just something to maybe document?)

m4b (commented):
only question about allocation; otherwise, looks good to me, thankfully authenticode::ExcludedSections is not pub so not a breaking change

m4b:
after this change I believe we'll do a release, which will have the breaking changes we've accumulated, thanks for everyone's patience

src/pe/authenticode.rs
baloo: Yes it does. But parsing PE already requires you to allocate (for the section table specifically).
I don't really have a better option here. The spec requires me to sort the section tables, and there is a non-fixed amount of them.

baloo:
@Foxboron The file itself might be read-only (stored in usr/lib like `/usr/lib/shim/shimx64.efi`). I can't make changes to the file in a parser.
What I'm trying to do here is to make sure the data I feed to the signer is padded like it would be once we append a signature to the original file (and not break its checksum).

> The spec demands the file size itself to be aligned. The buffer you feed the signed is missing 12 bytes which is going to affect the padding you are applying.
Where is this 12 bytes alignment requirement? I'm only taking care of the 8 bytes alignment here.

The 8 bytes alignment problem is already handled when we actually *write* to the file. (this was taken care of in https://github.com/m4b/goblin/pull/361)

Foxboron:
> What I'm trying to do here is to make sure the data I feed to the signer is padded like it would be once we append a signature to the original file (and not break its checksum).

Sure, but you need to check the actual file size, not the buffer you are feeding the signer. I don't know Rust but

```
                            debug!("hashing {extra_data_start:#x} {len:#x}",);
                            let buf = &bytes[extra_data_start..extra_data_start + len];

                            self.state = IterState::Padding(buf.len());
```

Tells me you are not passing the file size itself, just a slice of the bytes you are hashing. The actual file size is defined as ` let file_size = bytes.len();` and not passed to the padding function.

Foxboron:
I would probably not hide the padding between `file_size > sum_of_bytes_hashed` either. I don't think there is any case where this isn't true (we are always omitting at least 12 bytes I think?), but the padding should be checked unconditionally.

baloo:
@Foxboron I see your point.

Padding was done by virtue of hash_size holding the data that was sent to the hasher / signer. I think the code was working because the `sum_of_bytes_hashed` and the `certificate_table_size` were 8-bytes aligned. So it will still end up being equal and I ended up with a derivative value that was equally 8-bytes aligned as `file_size` was.

I rewrote the logic in the last commit to be closer to the spec, and what you expect (? I think?).


Just to confirm, you meant 8-bytes aligned right? I can't find any requirement to align stuff to 12-bytes in the spec, nor any other implementation (not even in yours).

src/pe/authenticode.rs
```diff
@@ -76,24 +96,166 @@ impl<'s> Iterator for ExcludedSectionsIter<'s> {
...
                         self.state = IterState::Done;
-                        return Some(&bytes[start..]);
+
+                        if pad_size != 0 {
+                            debug!("hashing {pad_size:#x}");
+
+                            // NOTE (safety): pad size will be at most 7, and PADDING has a size of 7
+                            //                pad_size is computed ~10 lines above.
+                            debug_assert!(pad_size <= 7);
+                            debug_assert_eq!(PADDING.len(), 7);
```
baloo, line 255: @m4b are those `debug_assert` fine? I thought I'd might as well back up my claims?

m4b: If they’re some internal check that should only be false if someone refactors this code and doesn’t maintain some internal invariant afterwards then yea. If they’re like things that probably should be true but maybe aren’t if the binary is malformed, then no, should be an error

baloo: no, they are only meant to protect against a refactor.

m4b (approved):
let's go!

m4b:
released in 0.8.0, happy new years!

baloo:
Thanks! and happy new year to you too!

PR #386, opened 2024-01-05
Support for lossy MachO parsing

h33p:
Made it possible to parse MachO objects in less strict manner. However, default behavior is not changed. Therefore, there now is a new function `parse_2` which accepts additional argument for lossyness.

Open questions:

- Is the default behavior intended? If not, `parse_2` could be made private.
- Is `parse_2` sufficient name, or should this be called something like `parse_lossy`?
- I made all of the commands work without fail in lossy mode, however, maybe some should fail?
- Should `from_32_2` `from_64_2` on `Segment` be exposed to the public? I currently have it `pub(crate)`.

### DIFF
--- a/src/mach/mod.rs
+++ b/src/mach/mod.rs
@@ -147,7 +147,17 @@ impl<'a> MachO<'a> {
         }
     }
     /// Parses the Mach-o binary from `bytes` at `offset`
-    pub fn parse(bytes: &'a [u8], mut offset: usize) -> error::Result<MachO<'a>> {
+    pub fn parse(bytes: &'a [u8], offset: usize) -> error::Result<MachO<'a>> {
+        Self::parse_impl(bytes, offset, false)
+    }
+
+    /// Parses the Mach-o binary from `bytes` at `offset` in lossy mode
+    pub fn parse_lossy(bytes: &'a [u8], offset: usize) -> error::Result<MachO<'a>> {
+        Self::parse_impl(bytes, offset, true)
+    }
+
+    /// Parses the Mach-o binary from `bytes` at `offset` in `lossy` mode
+    fn parse_impl(bytes: &'a [u8], mut offset: usize, lossy: bool) -> error::Result<MachO<'a>> {
         let (magic, maybe_ctx) = parse_magic_and_ctx(bytes, offset)?;
         let ctx = if let Some(ctx) = maybe_ctx {
             ctx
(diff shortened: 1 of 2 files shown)


src/mach/mod.rs
```diff
@@ -147,7 +147,12 @@ impl<'a> MachO<'a> {
         }
     }
     /// Parses the Mach-o binary from `bytes` at `offset`
-    pub fn parse(bytes: &'a [u8], mut offset: usize) -> error::Result<MachO<'a>> {
+    pub fn parse(bytes: &'a [u8], offset: usize) -> error::Result<MachO<'a>> {
+        Self::parse_2(bytes, offset, false)
+    }
+
+    /// Parses the Mach-o binary from `bytes` at `offset` in `lossy` mode
+    pub fn parse_2(bytes: &'a [u8], mut offset: usize, lossy: bool) -> error::Result<MachO<'a>> {
```
m4b, line 155: definitely don't want this name in the public api; also why is lossy a bool? generally having a bool in a public function is not the best due to boolean blindness.

src/mach/segment.rs
```diff
@@ -466,6 +466,16 @@ impl<'a> Segment<'a> {
         segment: &SegmentCommand32,
         offset: usize,
         ctx: container::Ctx,
+    ) -> Result<Self, error::Error> {
+        Self::from_32_2(bytes, segment, offset, ctx, false)
+    }
+
+    pub(crate) fn from_32_2(
```
m4b, line 473: i'd rename this `from_32_impl` or something, change it to private, and have `from_32` call it with the boolean flag = false, and `from_32_lossy` call it with true

src/mach/mod.rs
```diff
@@ -502,6 +529,11 @@ pub enum Mach<'a> {
 impl<'a> Mach<'a> {
     /// Parse from `bytes` either a multi-arch binary or a regular mach-o binary
     pub fn parse(bytes: &'a [u8]) -> error::Result<Self> {
+        Self::parse_2(bytes, false)
+    }
+
+    /// Parse from `bytes` either a multi-arch binary or a regular mach-o binary
+    pub fn parse_2(bytes: &'a [u8], lossy: bool) -> error::Result<Self> {
```
m4b, line 536: ditto; i'd make this function, `fn parse_impl` or something, then have:
```rust
pub fn parse()

pub fn parse_lossy()
```

call the `impl` with the boolean value forwarded

m4b (changes requested):
interesting PR, thank you! so as per the comments, I don't want to expose a `parse_2` name publicly, needs to be `parse` and `parse_lossy` or something, with the `parse_impl` getting forwarded whether it's lossy or not.

the only other consideration is whether you want to think about future lossy parsing, where a config is passed in to the lossy form, and this determines which things get dropped/ignored, etc., though that might be overkill, i don't know.

I'm also not totally sold on the `lossy` name, but it's ok.

Some other random alternatives that are probably worse names:
1. strict/unstrict
2. infallible

all said and done lossy is probably the best name?

Elf also has a similar, but slightly orthogonal support for "lazy" parsing, which doesn't parse the whole thing, but iirc, this was ostensibly to allow infallible parsing when some parts are broken/let user decide if they want to fail or not, may want to look at that, so we have a more unified api/approach to lossy/infallible/lazy parsing?

h33p:
Thanks for review, finally got around addressing your comments. Please let me know if you need anything else.

m4b (approved):
Test would be nice for this if you have time, otherwise lgtm

m4b:
@h33p CI is failing on rustfmt

m4b:
once format fixed this is good to go

h33p:
This should be good to go now @m4b

m4b (approved):
thanks!

m4b:
note: non-breaking

m4b:
released in 0.8.1

Issue #325, opened 2022-08-25
Is it possible to modify binary with goblin?


MatthewShao:
Hi, goblin is a really cool project! My use case is to modify the symbol names of elfs according to a set of rules. I went through the doc of [goblin](https://docs.rs/goblin/latest/goblin/index.html)::[strtab](https://docs.rs/goblin/latest/goblin/strtab/index.html)::[Strtab](https://docs.rs/goblin/latest/goblin/strtab/struct.Strtab.html#), it seems the underlaying bytes are immutable. Is it possible to modify those and write back to the binary, and if so, could you give a simple example? Thanks.

msuiche:
Looking for a similar answers, can we modify (removing/adding program_headers, adding NOTES, etc.) a parsed Elf object and write the modified file header to a target file.

msuiche:
Actually, if the structures had methods to know there current offset and to convert structs into [u8] - that would be way enough since they are repr(C)

m4b:
modifying binaries is a complex question; https://github.com/m4b/faerie is an example of an object file writer using goblin; that being said, modifying an existing binary/executable/shared library is possible, though the various pointers or offsets relied upon in various places would need to be updated.

since goblin assumes someone else owns the incoming bytes, the user could in theory write all this logic to manipulate the bits at whatever offsets it needs to in order to modify the binary, but the user would have to do all the necessary research to figure out what needs to be modified where, etc.

for elf and macho the structs all implement Pwrite so you could populate them with whatever values you want, then write them back into the byte stream, for example.

Anyway, I think the answer is yes, but it will be complicated! If you end up with something like objcopy or a binary modifier using goblin, please share back here, always interested to see new usescases, or problems encountered (with goblin) when doing so, thanks for your interest!

Closing this, but feel free to re-open if you encounter some bug or issue in goblin that prevents you from doing what you want :)

PR #387, opened 2024-01-11
elf: Add convenient functions for symbol write

tiann:
Here is an example of modifying elf:

```rust
use goblin::elf::{section_header, sym::Sym, Elf};
use scroll::{ctx::SizeWith, Pwrite};

use anyhow::Result;
use std::fs;

fn main() -> Result<()> {
    let path = "my.elf";
    let mut buffer = fs::read(path)?;
    let elf = Elf::parse(&buffer)?;

    let ctx = elf.syms.ctx().clone();
    for (index, mut sym) in elf.syms.iter().enumerate() {
        let Some(name) = elf.strtab.get_at(sym.st_name) else {
            continue;
        };

        if name == "my_symbol" {
            let offset = elf.syms.offset() + index * Sym::size_with(elf.syms.ctx());
            println!("name: {}, offset: {}", name, offset);
            sym.st_shndx = section_header::SHN_ABS as usize;
            sym.st_value = 0xeaeaeaea;
            buffer.pwrite_with(sym, offset, ctx)?;
            break;
        }
    }

    Ok(())
}

```

https://github.com/m4b/goblin/issues/325#issuecomment-1288407913 shows that we can modify elf, but we can't get the underlying offset and the symbol size.

Although we can iterate the section header and find the offset, it is not so convenient, this PR exposes some useful field to make it easier.

### DIFF
--- a/src/elf/sym.rs
+++ b/src/elf/sym.rs
@@ -541,6 +541,18 @@ if_alloc! {
             self.count
         }
 
+        /// The offset of symbol table in elf
+        #[inline]
+        pub fn offset(&self) -> usize {
+            self.start
+        }
+
+        /// The ctx of symbol table
+        #[inline]
+        pub fn ctx(&self) -> &Ctx {
+            &self.ctx
+        }
+
         /// Returns true if table has no symbols.
         #[inline]
         pub fn is_empty(&self) -> bool {


m4b (approved):
thanks LGTM!

m4b:
note: non-breaking, added pub api

m4b:
released in 0.8.1

PR #391, opened 2024-01-25
elf: Allow parsing SectionHeader from raw bytes

### LINKED ISSUE #390: Unable to parse ELF SectionHeader from raw bytes
I am unsure if this is a side effect of something else or done on purpose but let's imagine the following code:
```
            let mut fs = fs::File::open(path)?;
            let section_header_bytes = copy_array(
                fs,
                elf_head.e_shoff as usize,
                (elf_head.e_shnum as usize) * (elf_head.e_shentsize as usize),
            );
            // println!("SectionHeaderBytes: {:?}", section_header_bytes);

            let mut elf = elf::Elf::lazy_parse(elf_head)?;
            let context = goblin::container::Ctx {
                container: elf.header.container()?,
                le: elf.header.endianness()?,
            };

            elf.section_headers = SectionHeader::parse(
                &section_header_bytes,
                0,
                elf_head.e_shnum as usize,
                context,
            )?;

            println!("SectionHeaders: {:#?}", elf.section_headers);
```

The `SectionHeader::parse()` will fail because of https://github.com/m4b/goblin/blob/5b67d29bee4fa31766cdb9ec04062503b836701e/src/elf/section_header.rs#L449-L451

It likely makes sense when parsing the whole complete file, but in my case I want to just parse this section, and the offset for that section is e.g., at `0xc8e912a0` so that's quite a few bytes to read from the file for nothing. Locally just avoiding this works very well.

lissyx:
Fixes #390

### DIFF
--- a/src/elf/section_header.rs
+++ b/src/elf/section_header.rs
@@ -442,18 +442,25 @@ if_alloc! {
             self.sh_addr as usize..(self.sh_addr as usize).saturating_add(self.sh_size as usize)
         }
         /// Parse `count` section headers from `bytes` at `offset`, using the given `ctx`
+        /// Assuming this is read from the whole file, it will check offset.
         #[cfg(feature = "endian_fd")]
-        pub fn parse(bytes: &[u8], mut offset: usize, mut count: usize, ctx: Ctx) -> error::Result<Vec<SectionHeader>> {
-            use scroll::Pread;
+        pub fn parse(bytes: &[u8], offset: usize, count: usize, ctx: Ctx) -> error::Result<Vec<SectionHeader>> {
             // Zero offset means no section headers, not even the null section header.
             if offset == 0 {
                 return Ok(Vec::new());
             }
+            Self::parse_from(bytes, offset, count, ctx)
+        }
+        /// Parse `count` section headers from `bytes` at `offset`, using the given `ctx`
+        /// without performing any offset checking to allow parsing relatively
+        #[cfg(feature = "endian_fd")]
+        pub fn parse_from(bytes: &[u8], mut offset: usize, mut count: usize, ctx: Ctx) -> error::Result<Vec<SectionHeader>> {
+            use scroll::Pread;
             let empty_sh = bytes.gread_with::<SectionHeader>(&mut offset, ctx)?;
             if count == 0 as usize {
-                // Zero count means either no section headers if offset is also zero (checked
-                // above), or the number of section headers overflows SHN_LORESERVE, in which
-                // case the count is stored in the sh_size field of the null section header.
+                // Zero count means either no section headers or the number of section headers
+                // overflows SHN_LORESERVE, in which case the count is stored in the sh_size field
+                // of the null section header.
                 count = empty_sh.sh_size as usize;
             }
 
--- a/tests/elf.rs
+++ b/tests/elf.rs
@@ -75,6 +75,36 @@ fn parse_text_section_size_lazy(base: &[u8]) -> Result<u64, &'static str> {
     Err("Didn't find text section")
 }
 
+fn parse_from_text_section_size_lazy(base: &[u8]) -> Result<bool, &'static str> {
+    let header = Elf::parse_header(base).map_err(|_| "parse elf header error")?;
+
+    use goblin::container::{Container, Ctx};
+    use goblin::elf::SectionHeader;
+
+    let ctx = Ctx {
+        le: scroll::Endian::Little,
+        container: Container::Big,
+    };
+
+    let sh_from_orig =
+        SectionHeader::parse(base, header.e_shoff as usize, header.e_shnum as usize, ctx)
+            .map_err(|_| "parse() section headers error")?;
+
+    let sh_from_offset = SectionHeader::parse_from(
+        &base[header.e_shoff as usize..],
+        0,
+        header.e_shnum as usize,
+        ctx,
+    )
+    .map_err(|_| "parse_from() section headers error")?;
+
+    if sh_from_orig == sh_from_offset {
+        return Ok(true);
+    }
+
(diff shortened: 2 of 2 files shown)


orhun (approved):

src/elf/section_header.rs
```diff
@@ -442,13 +442,19 @@ if_alloc! {
...
         #[cfg(feature = "endian_fd")]
         pub fn parse(bytes: &[u8], mut offset: usize, mut count: usize, ctx: Ctx) -> error::Result<Vec<SectionHeader>> {
-            use scroll::Pread;
             // Zero offset means no section headers, not even the null section header.
             if offset == 0 {
                 return Ok(Vec::new());
             }
+            Self::parse_raw(bytes, offset, count, ctx)
+        }
+        /// Parse `count` section headers from `bytes` at `offset`, using the given `ctx`
```
m4b, line 454: update document comment here to explain what this does., and why it's different; I'm also not sure "raw" is the best name here/worried we're getting an api divergence from e.g., `parse_lossy` in the mach-o PR for example

m4b (commented):
might be good to review other names for this function, perhaps `parse_lossy`, i'm not sure if it does what the mach-o PR does, but otherwise this LGTM

src/elf/section_header.rs
lissyx: I'll adopt `parse_lossy` to follow PR #386

lissyx:
@m4b The lossy Mach-O PR seems to do a few different things, whereas here we just avoid checking the offset on the `parse_lossy` version. Their PR also introduce a `parse_impl()` so that might match better with what we do in fact ?

src/elf/section_header.rs
```diff
@@ -442,13 +442,20 @@ if_alloc! {
             self.sh_addr as usize..(self.sh_addr as usize).saturating_add(self.sh_size as usize)
         }
         /// Parse `count` section headers from `bytes` at `offset`, using the given `ctx`
+        /// Assuming this is read from the whole file, it will check offset.
         #[cfg(feature = "endian_fd")]
-        pub fn parse(bytes: &[u8], mut offset: usize, mut count: usize, ctx: Ctx) -> error::Result<Vec<SectionHeader>> {
-            use scroll::Pread;
+        pub fn parse(bytes: &[u8], offset: usize, count: usize, ctx: Ctx) -> error::Result<Vec<SectionHeader>> {
```
lissyx, line 447: Build was complaining with mutability of `offset` and `count`, but I am unsure whether this is a change we want from an API point of view. Will remove if you think we should.

tests/elf.rs
```diff
@@ -75,6 +75,32 @@ fn parse_text_section_size_lazy(base: &[u8]) -> Result<u64, &'static str> {
     Err("Didn't find text section")
 }
 
+fn parse_impl_text_section_size_lazy(base: &[u8]) -> Result<bool, &'static str> {
```
lissyx, line 78: happy to use a nicer name

lissyx:
@m4b Gentle ping for re-reviewing ? Thanks !

src/elf/section_header.rs
m4b: Sorry I should have been clearer; so in general, `parse_impl` should be a private fn that is used by e.g., two other (maybe `pub`) function bodies, (really the idiom I use is any privately shared body has `_impl` e.g., `fn something_impl`).

What I meant by the api similarity was the new `pub` function should perhaps be called `parse_lossy`, but looking again at this PR, I think you're right, and my suggestion was wrong (apologies!), in that it isn't parsing in a lossy manner, but rather it's parsing the raw headers *at* or *from* a specific byte slice.

So I think in this case a better name is `parse_from` or `parse_at` (I think I'm partial to `parse_from`?)

Alternatives that I don't like but considered:
1. parse_relative
2. parse_headers_from
3. parse_headers_at

If you change the function back to `parse_from` I think we're ok to merge.

One last thing, I don't really like that the user would have to know that `count == 0` will "turn on" the `empty_sh.sh_size`  condition, but I'm not sure of a better way to factor this, and it was also technically true of the original function anyway, so it's fine i guess :)

m4b (commented):
just the name change (again, sorry!)  then ready to go, thank you for your patience!

src/elf/section_header.rs
lissyx: yes, I get the meaning of the change, I was unsure about how it could impact your API :)

src/elf/section_header.rs
lissyx: Honestly, I dont really care about the naming, so if you have a preference over `parse_from`, let's use it :)

src/elf/section_header.rs
```diff
@@ -442,13 +442,20 @@ if_alloc! {
...
+            Self::parse_impl(bytes, offset, count, ctx)
+        }
+        /// Parse `count` section headers from `bytes` at `offset`, using the given `ctx`
+        /// without performing any offset checking to allow parsing relatively
+        #[cfg(feature = "endian_fd")]
+        pub fn parse_impl(bytes: &[u8], mut offset: usize, mut count: usize, ctx: Ctx) -> error::Result<Vec<SectionHeader>> {
+            use scroll::Pread;
             let empty_sh = bytes.gread_with::<SectionHeader>(&mut offset, ctx)?;
             if count == 0 as usize {
                 // Zero count means either no section headers if offset is also zero (checked
```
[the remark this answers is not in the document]
lissyx: @m4b remove that part, but I was not sure the wording would not need a bit of refinement ?

m4b (approved):
thank you!

m4b:
release note: non-breaking

m4b:
released in 0.8.1

PR #395, opened 2024-03-11
fix doc warnings

5225225:
All of these were reported by `cargo doc`. (This should probably have a CI job set up to fail on warnings).

For the authenticode change, I figured it wasn't really that important to link to the method that the doc is *on*, so I just deleted that. And for the strtab change, the anchor hash at the end of the url didn't seem to *link* anywhere relevant?


m4b (approved):
thank you!

m4b:
if you would be interested in setting up a CI job for that, if it isn't too difficult, that would be nice; my only major concern would be if the error messages in CI aren't nice/good/helpful, this could increase difficulty for contributors perhaps

m4b:
release: non-breaking

Issue #307, opened 2022-05-23
'cannot map exception_rva (0x5e000) into offset' when parsing PE32+ library


robbert-vdh:
Hi,

I've been using winedump to parse a whole bunch of Windows library files to check whether they export certain functions. This has been working fine, but shelling out for every file is a bit inefficient and it adds an external dependency. I've been meaning to replace this with goblin, but with one of the 700-ish files I've been parsing I'm getting this error:

> Malformed entity: cannot map exception_rva (0x5e000) into offset

Problem is, it's happening with this VST2 plugin, and they don't provide a demo: https://joeysturgistones.com/products/gain-reduction-deluxe

I'm not sure if I can legally provide a small snippet from the binary , but I definitely can't upload the entire file.

How should I proceed? For now I'll probably add Winedump as a backup in case goblin can't parse the file, but I'd love get rid of the Winedump dependency entirely.

m4b:
This may be what 9359fd09 was designed for; @robbert-vdh could you try out the parse_opt version by any chance?

robbert-vdh:
With that option disabled I'm getting a `bad offset 1467245907`. No idea what would or should be at that offset though, don't have the time to really look into this at the moment. Presumably there's something funky going on with this binary, but both Wine as well as actual Windows can handle it just fine.

m4b:
anyone want to test if this is still a bug?

robbert-vdh:
I just tested the same file again with goblin 0.6.0 from crates.io, it still happens.

ideeockus:
@robbert-vdh Hi! Can you provide information about PE sections? `readpe -S` or smth

I'm faced a similar problem, and I think the problem is probably in `utils::find_offset`

PR #396, opened 2024-03-13
Fix parsing PE with zero raw_data_size of section

ideeockus:
1. Sometimes I need to parse PE header without parsing DOS header + DOS stub

2. Seems that finding offset by rva is not implemented correctly and breaks down on some PE files. See #307


src/pe/header.rs
```diff
@@ -569,6 +569,29 @@ impl Header {
             optional_header,
         })
     }
+
+    pub fn parse_without_dos(bytes: &[u8]) -> error::Result<Self> {
```
m4b, line 573: needs documentation if `pub`; i also want to understand what's different here and why this can't be the original function for parsing?

m4b (changes requested):
it seems like there are two changes, one adding a new public function, and the other fixing a bug in the rva helper function, yes? need to understand why you need the new pub function, but the other fix probably is ok, though i would prefer a test showing how it fixes a bug

src/pe/header.rs
ideeockus: @m4b Hello! This bug occurs when parsing file AmdCleanupUtility https://www.amd.com/en/resources/support-articles/faqs/GPU-601.html. Should I add `.exe` file to tests?

ideeockus: > need to understand why you need the new pub function

Do you mean error on `DosHeader::parse(&bytes)?` should be handled inside original `Header::parse`?
And try to parse without DosHeader in case of Error?

m4b: What I'm asking is why is a new function being added, when does the user call it, and why can't we handle this transparently inside of the regular `parse` function? Ideally I'd prefer not to add another `pub` function if possible; maybe it is not possible, I don't know

ideeockus: we can handle this transparently inside of the regular `parse` function, but
1. users may expect that `parse` returns error when DOS stub is missing
2. there are existing tests that check invalid stub will not parse
3. when it is known that we have not DOS stub in buffer, it makes sense to use `parse_without_dos`

ideeockus: I just added the documentation, is it ok now?

src/pe/header.rs
```diff
@@ -569,6 +570,30 @@ impl Header {
             optional_header,
         })
     }
+
+    /// Parses PE header from the given bytes that not contains DOS stub, default DosHeader will be used
```
m4b, line 574: grammar: Parses PE header from the given bytes, a default DosHeader and DosStub are generated, and any malformed header or stub is ignored.

src/pe/utils.rs
```diff
@@ -64,6 +64,8 @@ fn section_read_size(section: &section_table::SectionTable, file_alignment: u32)
 
     if virtual_size == 0 {
         read_size
+    } else if size_of_raw_data == 0 {
```
m4b, line 67: i'm a little worried about this; this function is used all over goblin, can you add a comment why this check is done, and why it returns the virtual_size in this case? also there are no tests added to verify anything here

src/pe/header.rs
m4b: ok thank you for clarifying when this happens, and why to use; i've suggested grammar for the documentation of `parse_without_dos`

src/pe/header.rs
```diff
@@ -543,6 +543,7 @@ pub struct Header {
 }
 
 impl Header {
+    /// Parses PE header from the given bytes
```
m4b, line 546: thank you for adding documentation here; can you add the part about how we will fail if dos stub or header is wrong?

e.g.:

```
Parses PE header from the given bytes; this will fail if the DosHeader or DosStub is malformed or missing in some way
```

src/pe/header.rs
```diff
@@ -569,6 +570,30 @@ impl Header {
...
+            None
+        };
+
+        Ok(Header {
+            dos_header,
+            dos_stub: DosStub::default(),
+            signature,
+            coff_header,
+            optional_header,
+        })
```
m4b, line 595: I should have said this before, but all of this is basically identical, it might be nice to make a private function:

```rust
fn parse_impl(bytes: &[u8]), header: DosHeader, stub: DosStub) -> Self {
  // highlighted section
}
```
and then the two pub functions call this respectively (your new function supply Default to the header and stub args)

m4b (changes requested):
thanks for your patience! this is the last round, to summarize:

1. documentation grammar fixes I suggested
2. making the private impl function would be nice

and then this is ready to go, thank you!

src/pe/utils.rs
ideeockus: analogically to `virtual_size == 0` -> used `read_size`, if `read_size == 0` -> used `virtual_size`. i saw [pefile](https://github.com/erocarrera/pefile/tree/master) does this and there are no errors in parsing the AMD sample, so i added this check too and it works

ideeockus: also added test + sample for this case: took file from https://www.amd.com/en/resources/support-articles/faqs/GPU-601.html + packed with upx.

is it ok if i add this file to tests?

src/pe/header.rs
ideeockus: thanks for suggestion, fixed

src/pe/utils.rs
m4b: I have no idea what the license of that binary is :sweat_smile:  so probably let's not add it for now; also, it seems to parse just fine for me without this patch? e.g.:

```
cargo run --example rdr -- ~/Downloads/amdcleanuputility.exe
```
has no issue, neither does `bingrep`

m4b (changes requested):
Ok this looks great, sorry for the delay! If you just delete the binary (and test), we can merge this; we can even put it on a minor release of 0.9 if you want, since it has no breaking changes. There is another change upcoming that will need to rebase against this, or you rebase against them, i'd prefer the first, since we can minor release it if want, and also your PR has been outstanding longer :)

src/pe/utils.rs
ideeockus: Parse `amdcleanuputility.exe` itself is fine. But if binary will packed with upx, parsing goes broken, while binary can be run properly.

Ok, test removed from PR

m4b:
@ideeockus just fyi, in the next breaking release 0.10, the pe::Header will have a lifetime parameter, just curious, will this be ok for your usecases? If people need owned versions I was thinking of some ways to allow this, but if it's not readily needed, I won't think too hard about it

m4b (approved):

m4b:
ok published in 0.9.2, thanks for your contribution!

PR #397, opened 2024-03-14
pe: add Terse Executable (TE) support

### LINKED ISSUE #398: Add Terse Executable (TE) support to goblin
I would like to request the addition of Terse Executable (TE) support to goblin's PE module. a terse executable is simply a PE/PE32+ binary with the header replaced with a terse executable header (as defined in the PI specification: [TE}(https://uefi.org/specs/PI/1.8/V1_TE_Image.html). The TE header is a subset of values from the PE/PE32+ header and are the only fields necessary for the binary to be properly executed by a PI architecture compliant loader and executor. Terse executables are most commonly used by UEFI compliant firmware to reduce the overall size of the binary in firmware.

If willing to add this support, I have  created #397 for you to review and provide feedback on. I am open to changing implementation (such as having it be it's own module rather than a subset of the PE module (like `PeCoff`). As the maintainers, I look to you for details such as these.

Thanks for your time :)

Javagedes:
Add terse executable (TE) support to the PE module. A terse executable is a PE32/PE32+ binary with a reduced header size containing only the fields necessary for the binary to be properly executed by a [Platform Initialization (PI) architecture](https://uefi.org/specs/PI/1.8/V1_Introduction.html) compliant loader and executor. Terse executables are most commonly used by UEFI compliant firmware to reduce the overall size of the binary.

Only the header is replaced, and no other data is changed in a terse executable, resulting in all address values being invalid. The TE parser must take appropriate action to fix up addresses during parsing by adjusting the existing value by the difference between the `stripped_size` and the new size of the header.

closes #398


src/pe/header.rs
```diff
@@ -587,6 +587,145 @@ impl ctx::TryIntoCtx<scroll::Endian> for Header {
     }
 }
 
+/// The TE header is a reduced PE32/PE32+ header containing only fields required for
+/// execution in the PI architecture. The TE header is described in by the PI spec:
```
m4b, line 591: nit: I'm not familiar with PI architecture, might be good to mention what this acronym is/stands for elsewhere or before use in docs?

tests/bins/te/Readme.md
```diff
@@ -0,0 +1,18 @@
+# TE binaries
```
m4b, line 1: thanks for adding this readme! Please rename the file to README.md to match other readmes though :)

tests/te.rs
```diff
@@ -0,0 +1,109 @@
+#[cfg(test)]
+mod te_tests {
```
m4b, line 2: thank you for adding tests :)

m4b (approved):
this generally looks very good, thank you! i'm 50/50 on the fence about whether it should be it's own module, just so it's slightly more encapsulated, but it's such little code it may not be worth it. only minor nits, I'll give a chance to anyone else who wants to review, then i'm ok to merge this basically :)

src/pe/header.rs
Javagedes: Oh yes. That makes sense! I've updated the PR comment, and a few areas in the documentation to specify the acronym (Platform Initialization) and added a link to the specification.

tests/bins/te/Readme.md
Javagedes: Updated README.md to match the standard of the repository. Thanks!

tests/te.rs
Javagedes: No problem :) It's important!

Javagedes:
> this generally looks very good, thank you! i'm 50/50 on the fence about whether it should be it's own module, just so it's slightly more encapsulated, but it's such little code it may not be worth it. only minor nits, I'll give a chance to anyone else who wants to review, then i'm ok to merge this basically :)

@m4b I'm not against moving it to it's own module if you prefer. At a minimum, it would clean up the `#[cfg(feature = "te")]` that I sprinkled throughout `pe/mod.rs` and `pe/header.rs`. Additionally, if more capabilities were added later, it would remove the possibility of having a bloated pe module, and the need of moving it to another module down the road.

m4b:
thanks @Javagedes this is great

note: non-breaking

m4b:
release in 0.8.1

m4b:
released in 0.9.0, thank you for your patience!

Issue #314, opened 2022-06-14
PE images with multiple debug directory entries not fully supported


lzybkr:
Goblin currently parses the first debug directory and ignores any subsequent entries.

A PE debug directory can have multiple entries, for example, a C# program compiled with the option `/debug:embedded` might have 3 entries like:

```
Debug Directories

        Time Type        Size      RVA  Pointer
    -------- ------- -------- -------- --------
    DB2C7143 cv           11C 000024C8      6C8    Format: RSDS, {2943D330-EE6C-414F-A997-0E64FA8EB29F}, 1, Program.pdb
    00000000 pdbhash       27 000025E4      7E4    SHA256: 30 D3 43 29 6C EE 4F E1 69 97 0E 64 FA 8E B2 9F 43 71 2C 5B 0F 63 56 9C 29 D7 42 47 3E 86 68 70
    00000000 mpdb         790 0000260B      80B    Embedded Portable PDB: signature = 4244504D, uncompressed size = B30
```

In the above - the pdb data is in the image, not in a separate file. This isn't the only scenario where there may be multiple debug directory entries.

I'm not sure what a fix should look like - the [debug_data](https://github.com/m4b/goblin/blob/38200916e73c87d32563be279c4528558c43469b/src/pe/mod.rs#L58) field returns a singleton [Option\<DebugData\>](https://github.com/m4b/goblin/blob/38200916e73c87d32563be279c4528558c43469b/src/pe/debug.rs#L10), and the [image_debug_directory](https://github.com/m4b/goblin/blob/38200916e73c87d32563be279c4528558c43469b/src/pe/debug.rs#L11) field is also a single instance of [ImageDebugDirectory](https://github.com/m4b/goblin/blob/38200916e73c87d32563be279c4528558c43469b/src/pe/debug.rs#L58).

I think the `debug_data` field should ideally be a `Option<Vec<DebugData>>`, but that would be an api breaking change.

I suppose there could be an additional field name `additional_debug_data` to hold the entries after the first.

Note that I thought it might be very difficult to read the additional entries, but it's not too bad. One must explicitly add a reference to the `scroll` crate to use the `Pread` trait and then you can write something like this:

```rust
    let pe = PE::parse(&bytes)?;

    if let Some(optional_header) = pe.header.optional_header {
        if let Some(debug_directory) = optional_header.data_directories.get_debug_table() {
            let entries = debug_directory.size as usize / std::mem::size_of::<debug::ImageDebugDirectory>();
            
[...]

PR #403, opened 2024-04-12
PE: Support multiple debug directories and VCFeature, Repro, ExDllCharacteristics, POGO parsers

kkent030315:
This PR addresses the issue https://github.com/m4b/goblin/issues/314.

~~- The change is straightforward. It makes the `DebugData::image_debug_directory: ImageDebugDirectory` into `Vec<ImageDebugDirectory>` as someone pointed out in the https://github.com/m4b/goblin/issues/314#issuecomment-1979849304. So this is to be an breaking change.~~
~~- There is another addition of VC feature metadata `IMAGE_DEBUG_TYPE_VC_FEATURE` (`IMAGE_DEBUG_VC_FEATURE_ENTRY`) in the debug directory.~~

If anyone have suggestion of making this semantics without breaking the backward compatibility I am open to discuss.

---

update:

- (22e1a3ac6b0197838d55e96c609936e1001cc7fd) Refactored `DebugData` and its parsers for comfortable code that makes multiple debug directories in mind.
  - **`TE::fixup_debug_data` has been merged to `DebugData::parse_*`**: As `DebugData` is now more compatible with TE (Terse Executable) by adding `DebugData::parse_with_opts_and_fixup` and fixup RVA field in `DebugData` and `ImageDebugDirectoryIterator` in order if any fixup is required for `ImageDebugDirectory::address_of_raw_data` and `ImageDebugDirectory::pointer_to_raw_data`. 
- (04dad2d6bf10b60dc6b21989852b0a5bfd8be681) Added decent documentation for structs and its fields, constants in `pe::debug`.
- (04dad2d6bf10b60dc6b21989852b0a5bfd8be681) Added parser for `IMAGE_DEBUG_TYPE_VC_FEATURE`.
- (04dad2d6bf10b60dc6b21989852b0a5bfd8be681) Added parser for `IMAGE_DEBUG_TYPE_REPRO`.
- (04dad2d6bf10b60dc6b21989852b0a5bfd8be681) Added parser for `IMAGE_DEBUG_TYPE_EX_DLLCHARACTERISTICS`.
- (24879d67fbb76746b465319e44f21e6cf10e4a4d) Added parser for `IMAGE_DEBUG_TYPE_POGO`.
- (04dad2d6bf10b60dc6b21989852b0a5bfd8be681) Added decent integration tests for `DebugData` parsers.
- (22e1a3ac6b0197838d55e96c609936e1001cc7fd) (bfa48ca6cdec1116356b4471635abcc6455828a4) Removed `impl ImageDebugDirectory` and it has been merged to `DebugData::parse_*`.


src/pe/debug.rs
```diff
@@ -98,16 +110,22 @@ impl ImageDebugDirectory {
...
             error::Error::Malformed(format!(
                 "Cannot map ImageDebugDirectory rva {:#x} into offset",
                 rva
             ))
         })?;
-        let idd: Self = bytes.pread_with(offset, scroll::LE)?;
-        Ok(idd)
+        let idd_list = (0..entries)
+            .map(|i| {
+                let entry = offset + i * core::mem::size_of::<ImageDebugDirectory>();
```
m4b, line 124: nice :)

src/pe/debug.rs
```diff
@@ -127,60 +145,119 @@ pub struct CodeviewPDB70DebugInfo<'a> {
 }
 
 impl<'a> CodeviewPDB70DebugInfo<'a> {
-    pub fn parse(bytes: &'a [u8], idd: &ImageDebugDirectory) -> error::Result<Option<Self>> {
+    pub fn parse(bytes: &'a [u8], idd: &Vec<ImageDebugDirectory>) -> error::Result<Option<Self>> {
```
m4b, line 148: this should take a `&[ImageDebugDirectory]` to be more idiomatic rust

src/pe/debug.rs
```diff
@@ -127,60 +145,119 @@ pub struct CodeviewPDB70DebugInfo<'a> {
...
         Self::parse_with_opts(bytes, idd, &options::ParseOptions::default())
     }
 
     pub fn parse_with_opts(
         bytes: &'a [u8],
-        idd: &ImageDebugDirectory,
+        idd: &Vec<ImageDebugDirectory>,
```
m4b, line 154: ditto here

src/pe/debug.rs
```diff
@@ -127,60 +145,119 @@ pub struct CodeviewPDB70DebugInfo<'a> {
...
+    /// The count of guard stack
+    pub guard_stack_count: u32,
+    /// The count of SDL
+    pub sdl_count: u32,
+    /// The count of guard
+    pub guard_count: u32,
+}
+
+impl<'a> VCFeatureInfo {
+    pub fn parse(bytes: &'a [u8], idd: &Vec<ImageDebugDirectory>) -> error::Result<Option<Self>> {
```
m4b, line 226: ditto, `&[IMageDebugDirectory]`

src/pe/debug.rs
```diff
@@ -127,60 +145,119 @@ pub struct CodeviewPDB70DebugInfo<'a> {
...
+        Self::parse_with_opts(bytes, idd, &options::ParseOptions::default())
+    }
+
+    pub fn parse_with_opts(
+        bytes: &'a [u8],
+        idd: &Vec<ImageDebugDirectory>,
```
m4b, line 232: ditto

src/pe/debug.rs
```diff
@@ -1,15 +1,17 @@
...
 use crate::pe::data_directories;
 use crate::pe::options;
 use crate::pe::section_table;
 use crate::pe::utils;
 
-#[derive(Debug, PartialEq, Copy, Clone, Default)]
+#[derive(Debug, PartialEq, Clone, Default)]
 pub struct DebugData<'a> {
-    pub image_debug_directory: ImageDebugDirectory,
+    pub image_debug_directories: Vec<ImageDebugDirectory>,
```
m4b, line 12: do you know if there is a maximum image debug directories possible? if yes, perhaps we should make this an array and put it back to `Copy`?

m4b (changes requested):
this seems fine to me, other than the fixes w.r.t. `&Vec<T>` -> `&[T]`

src/pe/debug.rs
kkent030315: Yes, the maximum entry is defined as 21.
https://github.com/winsiderss/systeminformer/blob/6f97511ad7bb0c8807d99a3151494f55d0e21fc8/phlib/include/mapimg.h#L714

m4b:
@kkent030315 thanks for this PR, I think if we fix the basic nits this is ready to go, thank you!

m4b:
I'd also like to see this one go in for sure :)

kkent030315:
@m4b Thank you for the review! I'm going to mess with this PR very soon.

src/pe/debug.rs
kkent030315: Index itself is maximum 21 but it isn't sorted that way, I mean theres some gaps between entries, and some of "gaps" aren't even defined as "reserved" entry. So I think the best way is to implement iterator for this, as well as how we do in unwind table and rich metadatas.

kkent030315:
@m4b Hi, this PR has been refactored, please take a look https://github.com/m4b/goblin/pull/403#issue-2238926501 for up-to-date changelists.
Since there's lots of things changed and a little bit challenging code I guess so I'd like to get your review again. Thank you :)

src/pe/debug.rs
```diff
@@ -131,60 +308,127 @@ pub struct CodeviewPDB70DebugInfo<'a> {
...
-        let mut offset: usize = match opts.resolve_rva {
-            true => idd.pointer_to_raw_data as usize,
-            false => idd.address_of_raw_data as usize,
-        };
+        if let Some(idd) = idd {
+            // ImageDebugDirectory.pointer_to_raw_data stores a raw offset -- not a virtual offset -- which we can use directly
+            let mut offset: usize = match opts.resolve_rva {
+                true => idd.pointer_to_raw_data as usize,
+                false => idd.address_of_raw_data as usize,
+            };
```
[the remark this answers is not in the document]
kkent030315:
```rs
    #[test]
    fn parse_debug_exdllcharacteristics_rva() {
        let opts = ParseOptions {
            resolve_rva: false,
            parse_attribute_certificates: true,
        };
        let binary = crate::pe::PE::parse_with_opts(DEBUG_DIRECTORIES_TEST_MSVC_BIN, &opts)
            .expect("Unable to parse binary");
        assert_eq!(binary.debug_data.is_some(), true);
        let debug_data = binary.debug_data.unwrap();
        assert_eq!(debug_data.ex_dll_characteristics_info.is_some(), true);
        let ex_dll_characteristics_info = debug_data.ex_dll_characteristics_info.unwrap();
        let ex_dll_characteristics_info_expect = ExDllCharacteristicsInfo {
            characteristics_ex: IMAGE_DLLCHARACTERISTICS_EX_CET_COMPAT
                | IMAGE_DLLCHARACTERISTICS_EX_CET_COMPAT_STRICT_MODE,
        };
        assert_eq!(
            ex_dll_characteristics_info,
            ex_dll_characteristics_info_expect
        );
    }
```

Malformed error at `DebugData::parse_with_opts_and_fixup`

```
Unable to parse binary: Malformed("ImageDebugDirectory offset 0x2000 and size 0xa8 exceeds the bounds of the bytes size 0x600")
```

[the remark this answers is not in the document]
kkent030315: Well digging into how `opts.resolve_rva` is used revealed that it is for mapped image, so it's expected behaviour at all. Nvm

src/pe/debug.rs
```diff
@@ -131,60 +326,127 @@ pub struct CodeviewPDB70DebugInfo<'a> {
...
+        bytes: &'a [u8],
+        idd: ImageDebugDirectoryIterator<'_>,
+    ) -> error::Result<Option<Self>> {
         Self::parse_with_opts(bytes, idd, &options::ParseOptions::default())
     }
 
     pub fn parse_with_opts(
         bytes: &'a [u8],
-        idd: &ImageDebugDirectory,
+        idd: ImageDebugDirectoryIterator<'_>,
```
m4b, line 338: breaking change; is this avoidable?

src/pe/debug.rs
```diff
@@ -131,60 +326,127 @@ pub struct CodeviewPDB70DebugInfo<'a> {
...
-        let filename_length = idd.size_of_data as isize - 24;
-        if filename_length < 0 {
-            // the record is too short to be plausible
-            return Err(error::Error::Malformed(format!(
-                "ImageDebugDirectory size of data seems wrong: {:?}",
-                idd.size_of_data
-            )));
-        }
-        let filename_length = filename_length as usize;
+            // calculate how long the eventual filename will be, which doubles as a check of the record size
```
m4b, line 353: it looks like whitespace error was introduced here? i'm surprised CI doesn't catch this via `rustfmt`?

src/pe/debug.rs
```diff
@@ -131,60 +326,127 @@ pub struct CodeviewPDB70DebugInfo<'a> {
 }
 
 impl<'a> CodeviewPDB70DebugInfo<'a> {
-    pub fn parse(bytes: &'a [u8], idd: &ImageDebugDirectory) -> error::Result<Option<Self>> {
+    pub fn parse(
+        bytes: &'a [u8],
+        idd: ImageDebugDirectoryIterator<'_>,
```
m4b, line 331: breaking change, for same reasons below

src/pe/debug.rs
```diff
@@ -131,60 +326,127 @@ pub struct CodeviewPDB70DebugInfo<'a> {
...
-                codeview_signature,
-                signature,
-                age,
-                filename,
+    pub fn parse_with_opts(
+        bytes: &'a [u8],
+        idd: ImageDebugDirectoryIterator<'_>,
+        opts: &options::ParseOptions,
+    ) -> error::Result<Option<Self>> {
+        let idd = idd.collect::<Result<Vec<_>, _>>()?;
```
m4b, line 423: do you really need to allocate the vec here? even iterating twice is probably better

src/pe/debug.rs
```diff
@@ -201,61 +463,660 @@ pub struct CodeviewPDB20DebugInfo<'a> {
 }
 
 impl<'a> CodeviewPDB20DebugInfo<'a> {
-    pub fn parse(bytes: &'a [u8], idd: &ImageDebugDirectory) -> error::Result<Option<Self>> {
+    pub fn parse(
+        bytes: &'a [u8],
+        idd: ImageDebugDirectoryIterator<'_>,
```
m4b, line 468: ditto same breaking change

src/pe/debug.rs
```diff
@@ -201,61 +463,660 @@ pub struct CodeviewPDB20DebugInfo<'a> {
...
     pub fn parse_with_opts(
         bytes: &'a [u8],
-        idd: &ImageDebugDirectory,
+        idd: ImageDebugDirectoryIterator<'_>,
         opts: &options::ParseOptions,
     ) -> error::Result<Option<Self>> {
-        if idd.data_type != IMAGE_DEBUG_TYPE_CODEVIEW {
-            // not a codeview debug directory
-            // that's not an error, but it's not a CodeviewPDB20DebugInfo either
+        let idd = idd.collect::<Result<Vec<_>, _>>()?;
```
m4b, line 478: same concern about allocation

src/pe/debug.rs
```diff
@@ -201,61 +463,660 @@ pub struct CodeviewPDB20DebugInfo<'a> {
...
+    ) -> error::Result<Option<Self>> {
+        Self::parse_with_opts(bytes, idd, &options::ParseOptions::default())
+    }
+
+    pub fn parse_with_opts(
+        bytes: &'a [u8],
+        idd: ImageDebugDirectoryIterator<'_>,
+        opts: &options::ParseOptions,
+    ) -> error::Result<Option<Self>> {
+        let idd = idd.collect::<Result<Vec<_>, _>>()?;
```
m4b, line 570: ditto allocation concern

src/pe/debug.rs
```diff
@@ -201,61 +463,660 @@ pub struct CodeviewPDB20DebugInfo<'a> {
...
```
m4b, line 764: ditto allocation; it seems like all of these allocate simply to call `find` on the type iiuc?

m4b (commented):
initial review, did not dive in deeply; main concern was initially with the allocations; I can't recall offhand, but there is probably some trait you can implement to allow `find` to work with your iterator, to prevent the allocation; you'll likely need clone/copy on the iterator since it'll likely consume it, but that should be fine.

also would like to understand why the breaking change is necessary, can you give some motivation for that?

lastly, thank you so much (as usual!) for your incredible documentation on the code you commit, truly outstanding stuff!

src/pe/debug.rs
kkent030315: Yes, it is avoidable. This PR was initially designed and assumed subject to breaking change, while I'm pretty sure we could do that w/o actually breaking it. I give it a shot.

kkent030315: @m4b Alright, everything looks improved and there are no explicit alloc's left there. On the other hand, a breaking `pub` change in `debug::DebugData::parse_with_opts_and_fixup` for TE seems inevitable at this time. We actually _can_ let them alone, but that's out of our original purpose to support _multiple_ debug data directories.

Perhaps, there's a two ideas:
1. Let the TE breaking changes alone and release the rest as minor changes,
   Then, do the TE breaking changes separately in a 0.10 rollup.
2. Or, rollup everything, including the breaking TE change, in this PR as-is.

Which one do you prefer?

m4b:
I think this is ready to go once the above suggestions are applied.

m4b (commented):
ARGH i forgot to submit these comments

src/pe/debug.rs
```diff
@@ -1,16 +1,143 @@
...
+            .find(|idd| idd.data_type == data_type)
+    }
+}
+
+/// Represents debug data extracted from a PE (Portable Executable) or TE (Terse Executable) file.
+#[derive(Debug, PartialEq, Clone, Default)]
 pub struct DebugData<'a> {
-    pub image_debug_directory: ImageDebugDirectory,
+    /// Raw data covering bytes of entire [`ImageDebugDirectory`]
+    pub data: &'a [u8],
```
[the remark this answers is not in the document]
m4b: done

kkent030315:
@m4b Thank you for the review. Everything addressed. 👍

src/pe/debug.rs
```diff
@@ -36,88 +163,176 @@ impl<'a> DebugData<'a> {
...
+                })?;
+
+        // Ensure that the offset and size do not exceed the length of the bytes slice
+        if offset + dd.size as usize > bytes.len() {
+            return Err(error::Error::Malformed(format!(
+                "ImageDebugDirectory offset {:#x} and size {:#x} exceeds the bounds of the bytes size {:#x}",
+                offset, dd.size, bytes.len()
+            )));
+        }
+        let data = &bytes[offset..offset + dd.size as usize];
```
m4b, line 193: probably in the future we should port this kind of code to something like:

```rust
let Some (data) = bytes.get(offset..offset + dd.size as usize) else {
              return Err(error::Error::Malformed(format!(
                "ImageDebugDirectory offset {:#x} and size {:#x} exceeds the bounds of the bytes size {:#x}",
                offset, dd.size, bytes.len()
            )));
};
```

this is clearer, and doesn't repeat the bounds, but we its minor nit. I should have remembered this earlier, as we've done these in several other places, but the `else` guard syntax only merged fairly recently (maybe a year or so ago iirc, but maybe even longer :sweat_smile: )

src/pe/debug.rs
```diff
@@ -140,12 +355,6 @@ impl<'a> CodeviewPDB70DebugInfo<'a> {
         idd: &ImageDebugDirectory,
         opts: &options::ParseOptions,
     ) -> error::Result<Option<Self>> {
-        if idd.data_type != IMAGE_DEBUG_TYPE_CODEVIEW {
-            // not a codeview debug directory
-            // that's not an error, but it's not a CodeviewPDB70DebugInfo either
-            return Ok(None);
-        }
-
```
m4b, line 148: i'm just curious why did this get removed? Is it because we now parse CODEVIEW?

m4b (approved):
great stuff as usual, thank you! (and thank you for your patience!)

m4b:
NB: breaking change

src/pe/debug.rs
kkent030315: The check is moved to the caller:

```rs
if let Some(idd) = &it.find_type(IMAGE_DEBUG_TYPE_CODEVIEW) {
    codeview_pdb70_debug_info = CodeviewPDB70DebugInfo::parse_with_opts(bytes, idd, opts)?;
    codeview_pdb20_debug_info = CodeviewPDB20DebugInfo::parse_with_opts(bytes, idd, opts)?;
}
```

Consumers (callers) should be aware of debug types whenever they are calling any of `IMAGE_DEBUG_TYPE_*` parsers.

PR #404, opened 2024-04-12
PE: parse thread local storage - TLS data

kkent030315:
Parses TLS (thread local storage) data directory in PE64/32.

### DIFF
--- a/src/pe/mod.rs
+++ b/src/pe/mod.rs
@@ -220,12 +220,12 @@ impl<'a> PE<'a> {
                 )?);
             }
 
-            if let Some(&tls_table) = optional_header.data_directories.get_tls_table() {
+            if let Some(tls_table) = optional_header.data_directories.get_tls_table() {
                 tls_data = if is_64 {
                     tls::TlsData::parse_with_opts::<u64>(
                         bytes,
                         image_base,
-                        &tls_table,
+                        tls_table,
                         &sections,
                         file_alignment,
                         opts,
--- a/src/pe/mod.rs
+++ b/src/pe/mod.rs
@@ -26,6 +26,7 @@ pub mod relocation;
 pub mod section_table;
 pub mod subsystem;
 pub mod symbol;
+pub mod tls;
 pub mod utils;
 
 use crate::container;
@@ -71,6 +72,8 @@ pub struct PE<'a> {
     pub libraries: Vec<&'a str>,
     /// Debug information, if any, contained in the PE header
     pub debug_data: Option<debug::DebugData<'a>>,
+    /// TLS information, if any, contained in the PE header
+    pub tls_data: Option<tls::TlsData<'a>>,
     /// Exception handling and stack unwind information, if any, contained in the PE header
     pub exception_data: Option<exception::ExceptionData<'a>>,
     /// Certificates present, if any, described by the Certificate Table
@@ -106,6 +109,7 @@ impl<'a> PE<'a> {
         let mut import_data = None;
         let mut libraries = vec![];
         let mut debug_data = None;
+        let mut tls_data = None;
         let mut exception_data = None;
         let mut certificates = Default::default();
         let mut is_64 = false;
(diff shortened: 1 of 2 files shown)


src/pe/tls.rs
```diff
@@ -0,0 +1,247 @@
...
+            let offset =
+                utils::find_offset(rva, sections, file_alignment, opts).ok_or_else(|| {
+                    error::Error::Malformed(format!(
+                        "cannot map tls address_of_callbacks rva ({:#x}) into offset",
+                        rva
+                    ))
+                })?;
+            let mut i = 0;
+            // Read the callbacks until we find a null terminator
+            loop {
```
m4b, line 226: on a malformed binary could this cause an infinite loop?

src/pe/tls.rs
```diff
@@ -0,0 +1,247 @@
...
+    // NOTE: `u32` for 32-bit binaries, `u64` for 64-bit binaries.
+    pub address_of_callbacks: u64,
+    /// The size of the zero fill.
+    pub size_of_zero_fill: u32,
+    /// The characteristics of the TLS.
+    pub characteristics: u32,
+}
+
+/// TLS information.
+#[repr(C)]
```
m4b, line 35: i don't think you need `repr(c)` here, but it's possible i'm missing something

src/pe/mod.rs
```diff
@@ -216,6 +220,29 @@ impl<'a> PE<'a> {
                 )?);
             }
 
+            if let Some(&tls_table) = optional_header.data_directories.get_tls_table() {
+                tls_data = if is_64 {
+                    tls::TlsData::parse_with_opts::<u64>(
+                        bytes,
+                        image_base,
+                        &tls_table,
```
m4b, line 228: nit: shouldn't need `&` here and above in `let Some(&tls_table)`

src/pe/tls.rs
```diff
@@ -0,0 +1,247 @@
...
+    /// Raw data of the TLS.
+    pub raw_data: Option<&'a [u8]>,
+    /// TLS index.
+    pub slot: Option<u32>,
+    /// TLS callbacks.
+    pub callbacks: Vec<u64>,
+}
+
+impl ImageTlsDirectory {
+    #[allow(unused)]
```
m4b, line 49: not needed

m4b (commented):
some nits i would like to see fixed, but otherwise this lgtm; i almost merged but would prefer to see them fixed before merging, thank you for the PR!

src/pe/tls.rs
kkent030315: Yes. I think it would make sense to have a RVA validation there and return the malformed error if invalid RVAs found.

src/pe/tls.rs
kkent030315: Done.

src/pe/mod.rs
kkent030315: Done.

src/pe/tls.rs
kkent030315: You're right. Done.

src/pe/tls.rs
kkent030315: Done.

m4b: so what i meant specifically is: is this loop guaranteed to terminate? this comes up when there are malformed binaries, and the loop only terminates in a well-formed case, but if the loop has no upper bound, it might cause the library to run forever, which is unacceptable.

m4b (commented):
This is ready to go basically, but I need more clarification on whether that loop is guaranteed to terminate or not; after that can merge :) and thank you for this!

src/pe/tls.rs
m4b: Yea I was just making sure there wasn't some obvious termination condition we could add, e.g., end of file, etc., that would guarantee we are done looping; my general reflex when i see `loop` is to ask how it can terminate, as we've had bugs in past of infinite loops without proper termination condition, but your answer is sufficient, thank you!

m4b (approved):
LGTM thank you!

m4b:
note: breaking change

m4b:
released in 0.9.0, thank you for your patience!

PR #405, opened 2024-04-12
PE: read reserved DOS header fields

kkent030315:
This PR fixes `DosHeader::parse` by reading `reserved` and `reserved2` fields from the input instead of assuming that they are always zeros (PE64/32).

In the PECOFF, the DOS header fields except `e_lfanew` (the `pe_pointer` alias in the goblin) are entiely not read by the LDR (Windows loader). So assuming that the two fields in question does not really matter.

However, these fields are sometimes (and rarely) used by the PE packers for marking/watermaking purpose. In the real world example, Riot Games uses the `e_ss = 0x4952` and `e_sp = 0x544F` fields for `MZ ... RIOT` marker.


m4b (approved):
lgtm thank you!

m4b:
note: non-breaking

Issue #400, opened 2024-03-19
feature (PE): add rich header


JohnScience:
More about rich header: <https://bytepointer.com/articles/the_microsoft_rich_header.htm>.

Algorithm for parsing: <https://0xrick.github.io/win-internals/pe3/#rich-header>

Absence of rich header parsing for me is a non-issue but may be useful for those who do cybersecurity and malware analysis.

PR #406, opened 2024-04-13
PE: parse rich header and refactor DOS stub parser

kkent030315:
This PR adds parsing of Rich headers, as someone opened issue #400.

- Added rich header `goblin::pe::header::RichHeader` `goblin::pe::header::RichMetadata` parsing if present, with success/fail/corrupted tests.
- DOS stub is now non-fixed size of slice reference, and explicitly isolated from the DOS header.
  - The DOS stub is not always guaranteed to be 64-bytes long; there are some linkers (Borland C++) and PE packers generate application-specific DOS stub; (#422) and
  - The DOS stub includes the DOS header in definition. The DOS header and DOS stub are sometimes known to be separate, but also sometimes together. This is confusing stuff, and there are no official statement. By design, separating the header and the stub would be preferred. Concating them in the higher implementations (on the user side) are pretty easiest than making it together (otherwise user needs to mess with the raw buffer in the higher implementations.. bad).

I took the constant bytes in the test code from [mthiesen/link-patcher](https://github.com/mthiesen/link-patcher) (MIT). If this can potentially be license incompliance, I am happy to make own specimens for testing.


kkent030315:
I'm a bit tired, will fix the CI at the rest of today or tomorrow.

src/pe/header.rs
```diff
@@ -775,11 +795,12 @@ impl CoffHeader {
 ///
 /// The PE header is located at the very beginning of the file and
 /// is followed by the section table and sections.
-#[derive(Debug, PartialEq, Copy, Clone, Default)]
-pub struct Header {
+#[derive(Debug, PartialEq, Clone, Default)]
+pub struct Header<'a> {
     pub dos_header: DosHeader,
     /// DOS program for legacy loaders
-    pub dos_stub: DosStub,
```
m4b, line 782: breaking change, and unclear why it went from `DosStub` to `&[u8]`

src/pe/header.rs
```diff
@@ -1071,6 +1225,69 @@ mod tests {
         0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
     ];
 
+    static NO_RICH_HEADER: [u8; 262] = [
```
m4b, line 1228: this should be `const` not `static`

src/pe/mod.rs
```diff
@@ -410,7 +410,7 @@ impl<'a> ctx::TryIntoCtx<scroll::Endian> for PE<'a> {
             }
             _ => None,
         };
-        bytes.gwrite_with(self.header, &mut offset, ctx)?;
+        bytes.gwrite_with(self.header.clone(), &mut offset, ctx)?;
```
m4b, line 413: i think keeping copy derive on self.header is fine, and also its a breaking change to remove

src/pe/header.rs
```diff
@@ -814,20 +831,21 @@ impl Header {
...
 
-impl ctx::TryIntoCtx<scroll::Endian> for Header {
+impl<'a> ctx::TryIntoCtx<scroll::Endian> for Header<'a> {
     type Error = error::Error;
 
     fn try_into_ctx(self, bytes: &mut [u8], ctx: scroll::Endian) -> Result<usize, Self::Error> {
         let offset = &mut 0;
         bytes.gwrite_with(self.dos_header, offset, ctx)?;
-        bytes.gwrite_with(self.dos_stub, offset, ctx)?;
+        bytes.gwrite_with(self.dos_stub, offset, ())?;
```
m4b, line 848: i'd expect this to use the same ctx here but maybe there's a good reason

src/pe/header.rs
```diff
@@ -393,17 +409,21 @@ impl DosHeader {
 ///
 /// * De facto, can be followed by a non-standard ["Rich header"](https://0xrick.github.io/win-internals/pe3/#rich-header).
 /// * According to the standard, is followed by the  [Header::signature] and then the [CoffHeader].
-pub struct DosStub(pub [u8; 0x40]);
-impl Default for DosStub {
+pub struct DosStub<'a> {
+    pub data: &'a [u8],
```
m4b, line 413: this is a breaking change, going from newtype struct to regular struct with named field

src/pe/header.rs
```diff
@@ -381,10 +380,27 @@ impl DosHeader {
             pe_pointer,
         })
     }
+
+    /// Parse the DOS stub.
+    ///
+    /// The DOS stub is a small program that prints the message "This program cannot be run in DOS mode" and exits; and
+    /// is not really read for the PECOFF file format. It's a relic from the MS-DOS era.
+    pub fn parse_dos_stub<'a>(bytes: &'a [u8], pe_pointer: u32) -> error::Result<&'a [u8]> {
```
m4b, line 388: I would expect this fucntion to be `pub fn new` in the impl of `DosStub`?

src/pe/header.rs
```diff
@@ -792,15 +813,11 @@ pub struct Header {
     pub optional_header: Option<optional_header::OptionalHeader>,
 }
 
-impl Header {
-    pub fn parse(bytes: &[u8]) -> error::Result<Self> {
+impl<'a> Header<'a> {
```
m4b, line 816: this is also technically a breaking change; i believe you message makes clear why this should now be borrowed/have a lifetime

m4b (changes requested):
unfortunately there are some breaking changes here, but some seem unavoidable; primarily I'd like to see DosStub have it's parse method move into an `impl DosStub` and the `&[u8]` be reverted back to the DosStub type that was there before. otherwise it's looking ok!

m4b:
@kkent030315 gentle ping; would definitely like to see this go in, but I had requested some changes :)

kkent030315:
@m4b Hey, sorry for being shadow. I had do deal with the my life, you know.. anyways, a little bit changes have been applied to fix the code indeed, and the tests `cargo test` are passed. Would you like to checking this out?

The current codebase in this PR is what I am used/using as an backend parser helper for my own binary rewriting framework that can partiolize the any PE64 binary into object-by-object (aka struct-by-struct, byte-by-byte etc) and this rich header code works perfectly with the  1000+ unique PE64 binaries over the world (regardless of compiler toolchains, including but not limited to MSVC(link), Clang(LLD), MinGW-GCC, _NOTE: Only MSVC linker inserts rich header_) as well, so it's quite stable for now I believe, while there are something to deal with the coding style you pointed though.

image

I'm going to deal with another my PRs ASAP. I'd like to contribute more on goblin, as theres too many things to get rid of, and some insufficient features.

src/pe/header.rs
kkent030315: Solved in b1b841cea67fd4fbd7d0ae948dafb6b9d66a346f.
How's it now? if it looks good for you I'd like you to make this change request solved on your end :)

src/pe/header.rs
kkent030315: Solved in 0e7c3778a552e507fe17801c3a82b9852de53542.
Since this is trivial changes, I'd like to make this change request solved on my end, however if you have anything else please reopen.

src/pe/mod.rs
kkent030315: Ah, I agreeed. there was a borrow[^1] of `self.header` after this code and I think I was too lazy and let it clone, but it shouldn't be.

[^1]: [`"Symbol tables in PE are deprecated and not supported to write"` assertion](https://github.com/m4b/goblin/pull/406/files#diff-6303a321bf22fd69f034c1622e2b63c543fb1ceb6f3828e3dd81403e8fff4942R431)

kkent030315: I guess solved in 7bcc51b0b83a44d94b051f3079c8dac5fc8c3b20.
I simply moved the assertion before the `bytes.gwrite_with(self.header, &mut offset, ctx)?;`, since nothing modifies `self.header.coff_header.pointer_to_symbol_table` betweeen the grwrite and the assertion, so it's semantically the same no no problems.
Please let me know if you have better option. Otherwise I'd like you to mark this change request as resolved in your end.

src/pe/header.rs
kkent030315: Yeah.. since it's a code I wrote a while ago, and I now realized that `Copy` impl is the whole problem why I assigned the  explicit lifetime to the structs. Simply because [`std::vec::Vec<T>`](https://doc.rust-lang.org/std/vec/struct.Vec.html) doesn't implement `Copy`. So `RichHeader`, `RichMetadata` and the `Header` are entiely not be able to impl `Copy`. Maybe I need to remove `Vec<u8>` from those structs somehow

src/pe/header.rs
kkent030315: Unfortunately, there is no choice but to make this breaking change. Since the dos stub never be fixed size. It is very specific to the toolchain and compilers.

For ex:

- MSVC(link.exe): 64-bytes
- Clang(LLD): 54-bytes[^1][^2]
- Borland: 64-bytes, with unique signatures: `MZP` and "This program must be run under Win32"

So technically, it is implemented in wrong way to make it `Sized`. I believe we cannot avoid breaking changes to fix this problem and make goblin more stabilize and reliable in that way.

[^1]: https://github.com/llvm/llvm-project/blob/54d3cf14213f18e44ef9ed2ffe3b3131e472e2f5/lld/COFF/Writer.cpp#L70C1-L76C3
[^2]: This makes the `MZx` while is `MZ.` on MSVC, so I usually identifies the binary by this `MZx` to whether or it is compiled with MSVC linker. If `MZP`, it's Borland.

src/pe/header.rs
kkent030315: Solved in b1b841cea67fd4fbd7d0ae948dafb6b9d66a346f.
Since its trivial, I am making this change request solved.

src/pe/header.rs
kkent030315: So... maybe, ways if we would make DOS stub dynamically sized, I think of:

- **Breaking Change**: To have explicit lifetimes (as well as I did ealier; reverted recently) and do not copy any (regardless of trivial or big) buffers (`&'a [u8]`s).
  - Pros: Great code, no deep copy.
  - Cons: Codebase would potentially go to the hell of "lifetimes" and coding will not be easy for beginner contributors
- **Breaking Change**: Keep implicit lifetimes. Remove `Copy` impls from appropriate structs around `Header`, and abuse `clone`s, deep copy, and deal with compatibility problems within the current codebase.
  - Pros: The effect is minimum.
  - Cons: with performance tradeoffs.

kkent030315: Yeah.. something to related, I found another issue. During implementing tests I noticed that the DOS header produced by Borland(Indeed Turbo incremental linker included in Borland C++ Compiler) is not usual in any way and we need more changes to make goblin stable.

For short: The fixed `PE_POINTER_OFFSET` should be removed and must be dynamizally calculated. PE pointer offset of 
the binary in question is indeed `0x200`.

|Borland|MSVC|
|-|-|
|borland|msvc|

kkent030315: Fixed in 870cad9ab5d81778f9845f88f5a3b9b12e14741c.

Alright, breaking change again but it's inevitable. Removed `const PE_POINTER_OFFSET`, which can lead to the bug code and codes referenced that have been adjusted with correct calculation of PE pointer offset.

I removed an one assert that asserts that DOS header is ends at `PE_POINTER_OFFSET`, which is wrong implementation, as I mentioned ealier. No alternative assertion is required, since DOS stub bytes can be anything that is a valid DOS program bytecodes in theory.

With those changes, to ensure it is still correct and improved, I added the tests for Borland headers, and is now everything is passed.

Addiotonally, I added the detailed explaination on the default DOS stub byte code (in `impl Default for DosStub`). I hope that improves docs as well.

I'd like to hear from you about how to handle those inevitable breaking changes at this time.

src/pe/header.rs
kkent030315: Reopening. Yeah, the issue is that the `DosStub` is no longer a fixed size that the compiler knows at the compile time, and it should be dynamically sized. So `Pread` and `Pwrite` derives cannot be implemented. I guess there are no way to avoid using of `Vec<u8>`, but maybe in hacky way, like below, which is crazy (really) and we mustn't do that.

```rs
pub struct DosStub {
    pub data: [u8; 0x1000],
    pub real_len: usize,
}
```

I'm not faimilar with `scroll` or `Pread`, ``Pwrite` though, do you have any options?

```rs
error[E0308]: mismatched types
   --> src\pe\header.rs:887:45
    |
887 |         bytes.gwrite_with(dos_stub, offset, ctx)?;
    |               -----------                   ^^^ expected `()`, found `Endian`
    |               |
    |               arguments to this method are incorrect
```

m4b:
I'm sorry I knew there was another PR I needed to check before releasing 9.0, I will give this a review, but probably won't be till weekend. Please feel free to ping me then if you haven't gotten any feedback. And thanks for your contributions! :)

m4b:
Also I hope everything in your life is good; don't feel need to push yourself either, your life comes first :)

kkent030315:
@m4b Thanks for taking attention on this! I think theres lots of things to discuess here. But I think breaking changes are inevitable for this PR, perhaps how about introducing this feature on 0.10?

m4b (changes requested):
Ok this is the last time, I think we're almost done here; as I mention in another comment, tl;dr I did not realize the heart of the issue is that the DosStub is variable and bounded by `pe_pointer_offset`; your changes look correct, assuming _that statement_ is correct :) that being said, we should go back to your original intuition, as I  mention, and the bytes should not be allocated/owned, but slices, so we have to pass a lifetime param to Header, etc. This is fine, this pattern is common in other parts of goblin, e.g., mach-o has a lot of that.

For bonus points, we can keep Copy, i'm pretty sure, on all the structs, if you do `impl Iterator<Item=Result<RichMetadata>>` for a method on RichHeader  to avoid having a Vec inside of it.

After that, I Think this is ready to go, though do note it is a breaking change, a pretty good one, but I think it's worthwhile.

Oh and you have the honor of deleting the comment about adding RichHeader to the Header field, since you implemented it :)

src/pe/header.rs
```diff
@@ -332,12 +331,12 @@ impl DosHeader {
...
         );
 
-        let pe_pointer = bytes
+        let pe_pointer: u32 = bytes
             .pread_with(PE_POINTER_OFFSET as usize, scroll::LE)
             .map_err(|_| {
                 error::Error::Malformed(format!(
                     "cannot parse PE header pointer (offset {:#x})",
-                    PE_POINTER_OFFSET
+                    offset
```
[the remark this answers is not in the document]
kkent030315: We have an assertion right before this block. So it does not really matter if we use `PE_POINTER_OFFSET` or `offset`, so I'm reverting this.

```rs
        debug_assert!(
            offset == PE_POINTER_OFFSET as usize,
            "expected offset ({:#x}) after reading DOS header to be at 0x3C",
            offset
        );
```

src/pe/header.rs
```diff
@@ -835,6 +902,139 @@ impl ctx::TryIntoCtx<scroll::Endian> for Header {
...
+/// The Rich marker is a XOR-decoded version of the string "Rich" and is used to identify the Rich header.
+pub const RICH_MARKER: u32 = 0x68636952;
+
+/// The Rich header is a undocumented header that is used to store information about the build environment.
+///
+/// The Rich Header first appeared in Visual Studio 6.0 and contains: a product identifier, build number, and the number of times it was used during the build process.
+#[derive(Debug, PartialEq, Clone, Default)]
+pub struct RichHeader {
+    /// The Rich header data without the padding.
+    pub data: Vec<u8>,
```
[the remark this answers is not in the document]
kkent030315: Sometimes this is useful for calculating hash of rich data, so it would be nice to expose it to consumers. Anyways I am reverting back to byte slices and impl iterator for rich metadatas, as well as how unwind table is implemented.

src/pe/header.rs
```diff
@@ -835,6 +902,139 @@ impl ctx::TryIntoCtx<scroll::Endian> for Header {
...
+
+        // The Rich header is not present in all PE files.
+        if (pe_header_start_offset - dos_header_end_offset) < 8 {
+            return Ok(None);
+        }
+
+        // The Rich header is located between the DOS header and the PE header.
+        let scan_start = dos_header_end_offset + 4;
+        let scan_end = pe_header_start_offset;
+        debug_assert!(scan_end > scan_start, "Rich header scan range is invalid");
```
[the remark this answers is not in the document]
kkent030315: Sure thing, let's make it return malformed error.

src/pe/header.rs
```diff
@@ -835,6 +902,139 @@ impl ctx::TryIntoCtx<scroll::Endian> for Header {
...
+        let rich_header = &rich_header[padding_count..];
+
+        let metadatas = rich_header
+            .chunks(8)
+            .map(|chunk| {
+                let build_and_product = u32::from_le_bytes(chunk[..4].try_into().unwrap()) ^ key;
+                let build = (build_and_product & 0xFFFF) as u16;
+                let product = (build_and_product >> 16) as u16;
+                let use_count = u32::from_le_bytes(chunk[4..8].try_into().unwrap()) ^ key;
+                RichMetadata {
```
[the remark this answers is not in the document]
kkent030315: Thank you for the suggestion! It looks very brilliant to me.
I think it is perfect now: ceeb6ac0a85ea4e64a23286c1ecdac613cedeb9e

src/pe/header.rs
```diff
@@ -835,6 +902,139 @@ impl ctx::TryIntoCtx<scroll::Endian> for Header {
...
+            }
+        };
+
+        let rich_header = &rich_header[rich_start_offset..];
+
+        // Skip padding bytes
+        let padding_count = rich_header
+            .chunks(4)
+            .take_while(|chunk| {
+                let value = u32::from_le_bytes((*chunk).try_into().unwrap());
```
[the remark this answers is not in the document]
kkent030315: Sure, it should be "explicitly" panic-free now.

```rs
        // Skip padding bytes
        let padding_size = rich_header
            .chunks(4)
            .map(|chunk| chunk.pread_with::<u32>(0, scroll::LE))
            .collect::<Result<Vec<_>, _>>()?
            .into_iter()
            .take_while(|value| value == &key)
            .count()
            * 4;
```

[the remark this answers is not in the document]
kkent030315: And also, I'm going to improve readability for `* 4` portion.

src/pe/header.rs
```diff
@@ -391,17 +390,86 @@ impl DosHeader {
...
+        let start_offset = DOS_STUB_OFFSET as usize;
+        let end_offset = pe_pointer as usize;
+        if bytes.len() < end_offset as usize {
+            return Err(error::Error::Malformed(format!(
+                "DOS stub is too short ({} bytes) to contain the PE header pointer ({:#x})",
+                bytes.len(),
+                end_offset
+            )));
+        }
+        let dos_stub_area = &bytes[start_offset..end_offset];
```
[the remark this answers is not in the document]
kkent030315: Letter looks better for me, so we'd rather check `end_offset > start_offset` and leave the slice.

src/pe/header.rs
```diff
@@ -835,6 +902,139 @@ impl ctx::TryIntoCtx<scroll::Endian> for Header {
...
+            .count()
+            * 4;
+
+        // Extract the Rich header data without the padding
+        let rich_header = &rich_header[padding_count..];
+
+        let metadatas = rich_header
+            .chunks(8)
+            .map(|chunk| {
+                let build_and_product = u32::from_le_bytes(chunk[..4].try_into().unwrap()) ^ key;
```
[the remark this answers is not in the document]
kkent030315: Well, I introduced `RichMetadataIterator<'_>` and `RichMetadata::parse`, any invalid rich headers (including not limited to the invalid alignment) will now be treated as malformed errors. It should be fine now.

src/pe/header.rs
```diff
@@ -835,6 +902,139 @@ impl ctx::TryIntoCtx<scroll::Endian> for Header {
...
+        // The Rich header is located between the DOS header and the PE header.
+        let scan_start = dos_header_end_offset + 4;
+        let scan_end = pe_header_start_offset;
+        debug_assert!(scan_end > scan_start, "Rich header scan range is invalid");
+        let scan_stub = &bytes[scan_start..scan_end];
+
+        // First locate the Rich marker and the subsequent 32-bit encryption key.
+        let (rich_end_offset, key) = match scan_stub
+            .windows(8)
+            .position(|window| u32::from_le_bytes(window[..4].try_into().unwrap()) == RICH_MARKER)
```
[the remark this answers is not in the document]
kkent030315: I agree with you that this syntax isn't well and even looks ugly😅.
The syntax looks a little bit complex now, but is now panic-free and explicit malformed errors:

```rs
        // First locate the Rich marker and the subsequent 32-bit encryption key.
        let (rich_end_offset, key) = match scan_stub
            .windows(8)
            .enumerate()
            .filter_map(
                |(index, window)| match window.pread_with::<u32>(0, scroll::LE) {
                    // Marker matches, then return its index
                    Ok(marker) if marker == RICH_MARKER => Some(Ok(index)),
                    // Error reading with scroll
                    Err(e) => Some(Err(error::Error::from(e))),
                    // Marker did not match
                    _ => None,
                },
            )
            // Next is the very first element succeeded
            .next()
        {
            Some(Ok(rich_end_offset)) => {
                let rich_key =
                    scan_stub.pread_with::<u32>(rich_end_offset + RICH_MARKER_SIZE, scroll::LE)?;
                (rich_end_offset, rich_key)
            }
            // Something went wrong, e.g., reading with scroll
            Some(Err(e)) => return Err(e),
            // Marker did not found, rich header is assumed it does not exist
            None => return Ok(None),
        };
```

src/pe/header.rs
```diff
@@ -835,6 +902,139 @@ impl ctx::TryIntoCtx<scroll::Endian> for Header {
...
+                let rich_key =
+                    u32::from_le_bytes(scan_stub[offset + 4..offset + 8].try_into().unwrap());
+                (offset, rich_key)
+            }) {
+            Some(data) => data,
+            None => return Ok(None),
+        };
+
+        // Scope the buffer
+        let rich_header = &scan_stub[..rich_end_offset];
```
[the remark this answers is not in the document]
kkent030315: Oh yes. I've added check for this.

src/pe/header.rs
```diff
@@ -835,6 +902,139 @@ impl ctx::TryIntoCtx<scroll::Endian> for Header {
...
+        {
+            Some(offset) => offset,
+            None => {
+                return Err(error::Error::Malformed(format!(
+                    "Rich header does not contain the DanS marker"
+                )));
+            }
+        };
+
+        let rich_header = &rich_header[rich_start_offset..];
```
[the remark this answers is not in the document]
kkent030315: same; I've added check for this.

kkent030315:
@m4b Thank you very much for the review! I think it's almost perfect now. Would you able to take a look into the fixes?

src/pe/header.rs
kkent030315: > i was surprised to notice the latter is a longer slice, e.g. comparing the two:

So-called rich header is within DOS stub bytes. So the `stub2` contains whole data of rich header in its tail. :)

kkent030315:
Maybe something to discuss here https://github.com/m4b/goblin/pull/406#discussion_r1817859854:

> We can explicitly isolate DOS stub and Rich header datas, with some mroe complex implementations, but it makes no sense for me to do that.

I can definitely do that work if we should.

kkent030315:
Self review is done. Looks perfect for me right now.

m4b:
@kkent030315 apologies I wanted to merge the parse_without_dos patch since it was before this, and was waiting a while; so you'll have to rebase, and add a rich_header parameter to the `_impl` function and pass a `None`  for `parse_without_dos` is my guess; in meantime I'll do a final review but this looks ready to go just cursorily glancing at it

I appreciate your patience!

m4b:
So I just tried doing this rebase, and it's very annoying/tedious due to all the individual commits; i suggest squashing your patch down to a single commit first, and then rebasing on master to make it easiest.

kkent030315:
How about now? I resolved all conflicts wtih the upstream and should be able to merge with squash without problems.

kkent030315:
> apologies I wanted to merge the parse_without_dos patch since it was before this, and was waiting a while; so you'll have to rebase, and add a rich_header parameter to the _impl function and pass a None for parse_without_dos is my guess; in meantime I'll do a final review but this looks ready to go just cursorily glancing at it

As of current implementation of rich header parser is always individual regardless of DOS stub, since it takes hardcoded offset of end of DOS stub, so rich header parser still works even if DOS stub is default (no rich header) by `Header::parse_without_dos`, as I added test for this situation:

```rs
    #[test]
    fn parse_without_dos() {
        let header = Header::parse_without_dos(&BORLAND_PE32_VALID_NO_RICH_HEADER).unwrap();
        assert_eq!(header.dos_stub, DosStub::default());
        assert_eq!(header.rich_header.is_none(), true);

        // DOS stub is default but rich parser still works
        let header = Header::parse_without_dos(&CORRECT_RICH_HEADER).unwrap();
        assert_eq!(header.dos_stub, DosStub::default());
        assert_eq!(header.rich_header.is_some(), true);
    }
```

I don't know if we should not parse rich header when `Header::parse_without_dos` or we should. Do you have any idea?

m4b (approved):
Let's get feedback from them whether they want the rich data, otherwise we can leave as is; i have to run outside right now but i'll likely merge this tonight, thanks for all your great work!

m4b:
@ideeockus do you have any opinion on the rich header question?

src/pe/header.rs
```diff
@@ -391,17 +391,96 @@ impl DosHeader {
 ///
 /// * De facto, can be followed by a non-standard ["Rich header"](https://0xrick.github.io/win-internals/pe3/#rich-header).
 /// * According to the standard, is followed by the  [Header::signature] and then the [CoffHeader].
-pub struct DosStub(pub [u8; 0x40]);
-impl Default for DosStub {
+pub struct DosStub<'a> {
+    pub data: &'a [u8],
```
m4b, line 395: I'm going to change this to private and have a method `bytes() -> &'a [u8]` in a separate patch; if we ever want to switch this to something like `Cow<'a, [u8]>` and have an owned version to allow `Header<'a>` will be easier

m4b:
this was an epic PR @kkent030315 thanks for your patience!

PR #409, opened 2024-05-07
Add support for Codeview PDB 2.0 debug directory format

joschock:
Adds support for Codeview PDB 2.0 debug directory format.


m4b (approved):
how hard would it be to add a test?

joschock:
> how hard would it be to add a test?

Is there a test for the existing codeview7 support? I poked a bit, but I didn't see one - but if so, it would be pretty straightforward to extend/modify for CV2. Otherwise, it's slightly more work. If desired, I could push up a subsequent PR with a test...

joschock:
@m4b - any objections to merging this as is? Or would you like me to push a test first?

m4b:
note: breaking change

tests for this would be nice at some point

m4b:
released in 0.9.0

Issue #412, opened 2024-06-11
Parsing fails when a PE file has a TLS callback list which contains only the terminator


gabrielesvelto:
The code [here](https://github.com/m4b/goblin/blob/617775898fdfd0b843b2e937c80e455cf6f2a637/src/pe/tls.rs#L230-L241) checks whether an entry in the TLS callback list is the terminator only after having tried to load it. For executables with an empty callback list which contains only the terminator, parsing will fail.

gabrielesvelto:
This was fixed in https://github.com/m4b/goblin/pull/425

PR #413, opened 2024-06-11
Check if an entry in the TLS callback list is the terminator before trying to load it

gabrielesvelto:
This fixes issue #412.

### DIFF
--- a/src/pe/tls.rs
+++ b/src/pe/tls.rs
@@ -227,6 +227,11 @@ impl<'a> TlsData<'a> {
                 } else {
                     bytes.pread_with::<u32>(offset + i * 4, scroll::LE)? as u64
                 };
+
+                if callback == 0 {
+                    break;
+                }
+
                 // Each callback is an VA so convert it to RVA
                 let callback_rva = callback as usize - image_base;
                 // Check if the callback is in the image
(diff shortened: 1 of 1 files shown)


m4b (commented):
is it possible to add a test or update a test to make sure this doesn't regress?

gabrielesvelto:
Sure, I'll try to add it.

m4b:
let me know if you have a test case ready, or if you can modify an existing one that shows this change won't add any regressions, etc. thanks!

gabrielesvelto:
> let me know if you have a test case ready, or if you can modify an existing one that shows this change won't add any regressions, etc. thanks!

Yeah, I'll add a test here and provide a fix for issue #412 too later this year. I've been caught up in other stuff but I'll try to come back to this ASAP.

m4b:
@gabrielesvelto no worries, just doing some triage this weekend/poking people :)

m4b:
@gabrielesvelto thank you for this patch; it turns out this was re-discovered and fixed in another patch #425  fixing some recent tls stuff breaking on rust produced binaries, but thank you for your contribution and sorry we didn't merge it earlier :)

gabrielesvelto:
np, glad it's fixed!

PR #414, opened 2024-07-02
Fix issue 370

### LINKED ISSUE #370: Failing to parse QEMU memory dump note .shstrtab
Hi,
if you generate an ELF core file containing the memory dump of VM in QEMU (```qemu-system-x86_64``` than in console ```dump-guest-memory FILENAME```) and you parse it with a simple Rust program as
```
use goblin::Object;
use std::io::Read;
use std::fs::File;

fn main() {

    let mut file = File::open("/tmp/elf").map_err(|_| "open file error").expect("Error");

    let mut head = vec![0; 1024*1024*2];
    file.read(&mut head).ok();
    println!("{:?}\n", Object::parse(&head));
}
```

you get ```Err(Malformed("Section 1 size (151127112) + offset (11) is out of bounds. Overflowed: false"))``` but the ELF core is correctly formatted. I suppose the error is a offset-by-one error.

h33p:
Closes #370

The issue gets triggered, because qemu/virsh generated elf dump contains strtab at the end of the dump file. Meanwhile, dumps are large, and thus, it is infeasible to provide the entire dump file to the parser. This PR relaxes the requirements of the parser and returns default strtab, if the initial size check fails.

### DIFF
--- a/src/elf/mod.rs
+++ b/src/elf/mod.rs
@@ -70,6 +70,7 @@ if_sylvan! {
     use crate::container::{Container, Ctx};
     use alloc::vec::Vec;
     use core::cmp;
+    use log::warn;
 
     pub use header::Header;
     pub use program_header::ProgramHeader;
@@ -286,12 +287,19 @@ if_sylvan! {
                 }
 
                 if section_idx >= section_headers.len() {
-                    // FIXME: warn! here
+                    warn!("strtab section idx {} is out of bounds ({})", section_idx, section_headers.len());
                     Ok(Strtab::default())
                 } else {
                     let shdr = &section_headers[section_idx];
-                    shdr.check_size(bytes.len())?;
-                    Strtab::parse(bytes, shdr.sh_offset as usize, shdr.sh_size as usize, 0x0)
+                    // If size check fails, that means strtab is outside user supplied buffer. We
+                    // can either hard-fail there, or return an empty strtab. The latter is less
+                    // disturbing.
+                    if shdr.check_size(bytes.len()).is_ok() {
+                        Strtab::parse(bytes, shdr.sh_offset as usize, shdr.sh_size as usize, 0x0)
+                    } else {
+                        warn!("strtab section goes outside the provided buffer");
+                        Ok(Strtab::default())
+                    }
                 }
             };
 


m4b (commented):
Hmm, I'm not sure if this is the best approach to fixing this issue.
In general, if the file is huge, I'd expect it to be memmap'd and lazy paged in as user iterates/accesses parts. This should be sufficient to not crash parsing a huge coredump file.
Secondly, unless I've misunderstood, if the issue is that the strtab is *out of bounds*, this means that the file is malformed, unless only a chunk of the file is being loaded and passed to goblin; but goblin's `parse` was not really ever designed for that usecase.
In that case, probably the user should use `lazy_parse` , since strtab will get defaulted there.
Lastly, I'm also not sure if this really fixes any issue; won't subsequent lookups all fail due to an empty strtab, and so this is really just kicking the can down the road?

h33p:
@m4b thank you for your reply!

> Secondly, unless I've misunderstood, if the issue is that the strtab is out of bounds, this means that the file is malformed, unless only a chunk of the file is being loaded and passed to goblin;

Exactly. just as in #370, only the head chunk is passed over to goblin. The file itself is not malformed.

> In that case, probably the user should use lazy_parse , since strtab will get defaulted there.

This probably is the right way to go about this. My usecase is [only reading the program headers](https://github.com/memflow/memflow-kcore/blob/main/src/lib.rs#L25), nothing more.

This means, I can just use [`ProgramHeader::parse`](https://docs.rs/goblin/latest/goblin/elf/program_header/struct.ProgramHeader.html#method.parse), however, it requires a scroll Ctx. Is there a neat way to access/build that without copying the logic in [`parse_misc`](https://docs.rs/goblin/latest/src/goblin/elf/mod.rs.html#477)?

m4b:
Ok, I see your issue. So, this is primarily because lazy_parse isn't really fleshed out in a consistent way.  Iiuc, you would do:
```rust
let mut elf = Object::lazy_parse(bytes)?;
// how to populate program_headers?
```

So I think the best path forward here will be:
1. exposing the ctx by making the field `pub` or a `pub fn ctx()` (I don't want to do this)
2. adding some helper methods, especially for e.g., `program_headers` and `section_headers`, where it takes care of passing the various fiddly bits (i think this is ideal, but i'm not sure whether a builder-esque pattern, or mutating methods on the `elf` struct?

to flesh out 2. a bit, i mean something like:

```rust
pub fn parse_program_headers(&mut self, bytes: &[u8]) {
            let program_headers = ProgramHeader::parse(bytes, self.header.e_phoff as usize, self.header.e_phnum as usize, self.ctx)?;

            let mut interpreter = None;
            for ph in &program_headers {
                if ph.p_type == program_header::PT_INTERP && ph.p_filesz != 0 {
                    let count = (ph.p_filesz - 1) as usize;
                    let offset = ph.p_offset as usize;
                    interpreter = bytes.pread_with::<&str>(offset, ::scroll::ctx::StrCtx::Length(count)).ok();
                }
            }
      self.program_headers = program_headers;
      self.interpreter = interpreter;
}
```

Obviously this would forward to either private `static` methods that return the appropriate types, and so code duplication is reduced, or we use `&mut self` methods to construct/parse whichever components we want.

so there's a number of things I would change at this point, which is somewhat unrelated to your request to just parse the program headers:

1. We should probably introduce a builder or a `&mut self` api
2. `lazy_parse` *should* have taken some bytes and parsed the header, and defaulted the rest
3. `parse` calls `lazy_parse`, and then uses the above builder api to construct the remaining structs

-------------

I know this isn't probably what you wanted to hear, and while I understand your patch may "fix" the issue for you, it isn't unfortunately the right solution to the problem, which is specifically, you want to lazy_parse, and then supply just e.g., the program_header bytes and parse from those, to get the information you need.

Fortunately, I don't think this is a particularly hard problem to fix in a better manner, as I sort of outlined above it's just slightly more work :) let me know if you have any questions, and thanks for your patience and explaining your usecase better!

m4b:
another option for large files is to simply mmap them, using one of the various crates; this should allow you lazily iterate through parts of the binary while keeping your maximum working set low (generally, depending on what you're doing).

Issue #401, opened 2024-03-25
feature (PE): explicitly state presence or absence of support for various malformations


JohnScience:
[This video](https://youtu.be/uoQL3CE24ls?si=gEQf4dm7U_PIlLB-) goes in depth about various malformations which may be present in PE files.

m4b:
this should be fixed in the above PR, please reopen if you're able to test with that patch, and it fixes the issue; there will be a release coming shortly that rolls up a lot of PE changes too, if feel like waiting

PR #417, opened 2024-07-29
Fix fail on malformed certificate table parsing

ideeockus:
Hello! 

In some PE files Certificate Table can be malformed / contain invalid data. But if we use `ParseOptions::parse_attribute_certificates = true`, then whole parsing is failed.

I suggest use default `CertificateDirectoryTable` in case of error.


src/pe/mod.rs
```diff
@@ -3,12 +3,19 @@
 
 // TODO: panics with unwrap on None for apisetschema.dll, fhuxgraphics.dll and some others
 
-use core::cmp::max;
-
 use alloc::borrow::Cow;
 use alloc::string::String;
 use alloc::vec::Vec;
+use core::cmp::max;
```
m4b, line 9: I'm surprised by these import changes, did the formatter do this?

src/pe/mod.rs
```diff
@@ -142,7 +140,7 @@ impl<'a> PE<'a> {
                     return Err(error::Error::Malformed(format!(
                         "Unsupported header magic ({:#x})",
                         magic
-                    )))
+                    )));
```
m4b, line 143: ditto here too, this is also surprising to me

src/pe/mod.rs
```diff
@@ -268,7 +266,11 @@ impl<'a> PE<'a> {
                         bytes,
                         certificate_table.virtual_address,
                         certificate_table.size,
-                    )?;
+                    )
+                    .unwrap_or_else(|err| {
+                        warn!("Cannot parse CertificateTable: {:?}", err);
+                        Default::default()
```
m4b, line 272: I can't remember, what is a default certificate directory in this case? is it going to cause other problems further down the line when parsing, or if a user accesses parts of it, will it panic? Does it have offsets into other parts of the PE file that are no longer valid, etc.?

m4b (commented):
need some clarification on default certificate directory; also formatting things are odd

src/pe/mod.rs
ideeockus: default is empty table (no certificates), so no wrong offsets there

m4b:
So while this is an easy merge, I'm on the fence about whether we should; in general the malformed binary is kind of important to know, and in general, we choose to fail in those cases. However, sometimes we don't, and maybe this is one of those times, but it feels like it's just sort of skipping a malformed thing, and putting a default value in its place, which may be ok, but it also might not be.

So I'd like to understand more about:
1. why binaries have malformed certificates, is it common
2. is it not a big deal if they're malformed, e.g., it's safe to just supply a dummy/default/empty version of the table when bad ones are encountered, and continue going on?

thanks for your patience!

ideeockus:
> in general the malformed binary is kind of important to know, and in general, we choose to fail in those cases. However, sometimes we don't, and maybe this is one of those times, but it feels like it's just sort of skipping a malformed thing


OK, that sounds reasonable.  

What if we add something like `ParseMode::Strict`  and `ParseMode::Permissive`   to  `ParseOptions`?

ideeockus:
@m4b can you review please

I think handling malformations may be related to #401

src/pe/options.rs
```diff
@@ -8,6 +8,16 @@ pub struct ParseOptions {
     /// memory](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#other-contents-of-the-file).
     /// For on-disk representations, leave as true. Default: true
     pub parse_attribute_certificates: bool,
+    /// Whether or not to end with an error in case of incorrect data or continue parsing if able. Default: ParseMode::Strict
```
m4b, line 11: debating whether we should add `#[non_exhaustive]` to this struct to make it future compatible if we need to add another field like this in the future (and not break people)

src/pe/options.rs
```diff
@@ -16,6 +26,7 @@ impl ParseOptions {
         ParseOptions {
```
m4b, line 26: also i don't know who added this but this does not implement `Default`, but has a method named that instead; this should be fixed to actually implement `Default`

src/pe/options.rs
```diff
@@ -16,6 +26,7 @@ impl ParseOptions {
...
             resolve_rva: true,
             parse_attribute_certificates: true,
+            parse_mode: ParseMode::Strict,
         }
```
m4b, line 30: can you also add a method that returns what the TE parser uses; we can call it `fn te()` or something like that, instead of manually constructing it in TE portion.

m4b (changes requested):
this looks good to me, except as mentioned, but since we're here,  i'd like to see the:

1. non_exhaustive attrib added to ParseOptions
2. it actually implement `Default` trait
3. add `fn te() -> Self` method for constructing the parse options that TE requires

src/pe/options.rs
ideeockus: thanks for review, just pushed fixes

src/pe/options.rs
```diff
@@ -8,14 +9,35 @@ pub struct ParseOptions {
...
             resolve_rva: true,
             parse_attribute_certificates: true,
+            parse_mode: ParseMode::Strict,
+        }
+    }
+}
+
+impl ParseOptions {
+    pub fn te() -> Self {
+        Self {
```
m4b, line 37: more idiomatic would be something like:

```rust
Self {
  resolve_rva: false,
  parse_attribute_certificates: false,
  .. Self::default()
}
```

this way if new methods are added don't (maybe) need to update this location in source, but it's fine. I'm also wondering if we should make this non `pub`, just to reduce the api surface for now, i can fix this up though if you don't feel like pushing again.

m4b (approved):
this is great, thank you!

m4b (approved):

m4b:
NOTE: breaking change #434

Issue #424, opened 2024-10-21
Fail to parse PE files compiled by Rust


Berrysoft:
The error is:
```
Malformed("cannot map tls address_of_index rva (0xb5abc) into offset")
```
This error only occurs with goblin 0.9.0. Goblin 0.8.2 works fine.

The error seems only applies to exe create by Rust, for example, `sudo.exe`.

m4b:
@kkent030315 this seems directly related to the new tls parsing in https://github.com/m4b/goblin/pull/404 could you take a look? if a patch isn't available soon to fix this regression, I will likely disable the tls parsing in a minor patch and push that until it can be resolved better later on.

m4b:
@Berrysoft would you mind uploading a binary exhibiting the behavior here, so it's easier to repro/triage, thanks

Berrysoft:
[sudo.zip](https://github.com/user-attachments/files/17471219/sudo.zip)

It contains the `sudo.exe` from Windows 11.

m4b:
thank you, and yes i can repro on top of tree:
```
cargo run --example rdr -- sudo.exe
```

exhibits the failure.

m4b:
this is odd, doesn't seem to find the rva in the .data section:
```
DEBUG - Checking .data for 0xb5abc ∈ 0xb5000..0xb6228
```

this is returning false for whatever reason:
```rust
fn is_in_section(rva: usize, section: &section_table::SectionTable, file_alignment: u32) -> bool {
    let section_rva = section.virtual_address as usize;
    is_in_range(
        rva,
        section_rva,
        section_rva + section_read_size(section, file_alignment),
    )
}
```

something seems very wrong with some of the section end values (end < begin)?

```
DEBUG - Checking .text for 0xb5abc ∈ 0x1000..0x98190
DEBUG - CHECKING: rva=0xb5abc align=512, section=0x1000 section_end=0x97200
DEBUG - Checking .rdata for 0xb5abc ∈ 0x99000..0xb4534
DEBUG - CHECKING: rva=0xb5abc align=512, section=0x99000 section_end=0x1b600
DEBUG - Checking .data for 0xb5abc ∈ 0xb5000..0xb6228
DEBUG - CHECKING: rva=0xb5abc align=512, section=0xb5000 section_end=0xa00
DEBUG - Checking .pdata for 0xb5abc ∈ 0xb7000..0xba138
[...]

Issue #411, opened 2024-06-07
Parsing fails when the TLS index is in the zeroed area of a section


gabrielesvelto:
While processing DLLs from NVidia's graphics drivers I stumbled upon [this file](https://drive.google.com/file/d/1ykEVXIjzzulVVdrruZdqsTU0eHOjGAlR/view?usp=sharing) which failed to parse correctly. What happens at the point of error is the following:
- We're parsing the TLS section and we found a non-zero TLS index address so we're looking up its contents [here](https://github.com/m4b/goblin/blob/617775898fdfd0b843b2e937c80e455cf6f2a637/src/pe/tls.rs#L194)
- The value of the TLS index address is `0xdabe90` and the data section spans `0x175000..0xdac178` so it should work, but it fails to find it instead and moves on to the other sections, ultimately failing because it can't find a section to which the address belongs to
- The reason why `utils::find_offset()` fails even though the address falls within the section, is that it checks if the address falls between the beginning of the section and the beginning of the section plus the raw size, which is `0x3800` in this case. So it checks whether the TLS index address size is between `0x175000..0x178800`, which it isn't and thus fails.

The expected behavior of this particular code should be that the TLS index is simply set to zero, since the address points to an area of memory which is set to zero.

This seems to make sense from a logical perspective, given that Microsoft documentations mentions that this area is populated [by the loader](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#the-tls-section), so the section won't contain the actual value until it's loaded. It thus makes sense for it to be in a zeroed area.

gabrielesvelto:
I wonder if this might be a potential issue in more areas of the code. `utils::find_offset()` will always fail if the offset falls within a section, but not within the area populated in the file, however those offsets aren't invalid, they will be once the module is loaded into memory. It's just that they'll point to data that contain zeroes.

m4b:
are you able to resolve this or know of a patch you could contribute that can fix the underlying issue?

gabrielesvelto:
I've got a prototype patch ready, just need some time to polish it

m4b:
fixed by #425

PR #425, opened 2024-10-23
PE: Assume `AddressOfIndex` could not be in raw section, check callback is zero right after it reads from binary

kkent030315:
- Fixed issue #424 where binaries compiled Rust, and binaries with `AddressOfIndex` only exists in virtual address caused Malformed error by simply not raising explicit error but set the `slot` field to `None`.
- Fixed similar issue where the `callback` terminator check must be right before it reads from the raw binary. Otherwise it overflows when subtracting by image base.

### DIFF
--- a/src/pe/tls.rs
+++ b/src/pe/tls.rs
@@ -190,15 +190,17 @@ impl<'a> TlsData<'a> {
 
             // VA to RVA
             let rva = itd.address_of_index as usize - image_base;
-            let offset =
-                utils::find_offset(rva, sections, file_alignment, opts).ok_or_else(|| {
-                    error::Error::Malformed(format!(
-                        "cannot map tls address_of_index rva ({:#x}) into offset",
-                        rva
-                    ))
-                })?;
+            let offset = utils::find_offset(rva, sections, file_alignment, opts).ok_or_else(|| {
+                error::Error::Malformed(format!(
+                    "cannot map tls address_of_index rva ({:#x}) into offset",
+                    rva
+                ))
+            });
 
-            slot = Some(bytes.pread_with::<u32>(offset, scroll::LE)?);
+            slot = match offset {
+                Ok(offset) => Some(bytes.pread_with::<u32>(offset, scroll::LE)?),
+                Err(_) => None,
+            };
         }
 
         // Parse the callbacks if any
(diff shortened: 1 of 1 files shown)


kkent030315:
I think we can safely remove `slot` field as an alternative option. As explained in the issue, addres of indexes field is only useful at runtime and I personally think that field should not exists in the goblin while it is always useless.

@m4b How about removing the `slot` field and its parsing logic instead?

m4b:
> I personally think that field should not exists in the goblin while it is always useless for static binary analysis.

I'm confused, didn't you just add this field in 6c5b3e89534ec62caa033459b1cdb897c0cdede6 ?

I will merge this patch, thanks for posting it so quickly!

Re removing the slot field, I'd really like to avoid churn as much as possible, specifically in context of breaking changes. I know we can't predict the future but for future patches, try considering how breaking change could be avoided, or if a field/function really needs to be `pub`, etc.

While we're talking about that, I will likely deprecate the `parse` and `parse_with_opt` for the TlsData struct; I should have caught this in review, that's on me, but the use of generics there to get the u32/u64 is unintuitive, and is only used effectively to do `mem::std::size_of::<T>() == 8`; ideally the `is_64` bit should have been used, or I have a patch that uses scroll to parse this generically, but it's very ugly, and I wouldn't want it to be pub anyway.

I think a more goblin-idiomatic way to do it (on hindsight) would have been to write:

```rust
ImageTlsDirectory
ImageTlsDirectory32
```

and then write:
```rust
impl From<ImageTlsDirectory32> for ImageTlsDirectory {}
```

and then in the parse mechanism, if we're passed e.g., is_64 or whatever, we simply do:

```rust
let tls_directory = if is_64 { parse::<ImageTlsDirectory>() } else { parse::<ImageTlsDirectory32>().into() }
```

this alas will be a breaking change since the parse uses generics right now; anyway something to fix in the 0.10 release, and it's fairly minor.

anyway thanks for this fix!

m4b (approved):

kkent030315:
> > I personally think that field should not exists in the goblin while it is always useless for static binary analysis.
> 
> I'm confused, didn't you just add this field in https://github.com/m4b/goblin/commit/6c5b3e89534ec62caa033459b1cdb897c0cdede6 ?

Yes I did. Some binaries has `AddressOfIndex` valid in raw data (unlike Rust binaries in question) but nothing really useful. I just noticed it now😅.

gabrielesvelto:
I guess this fixed https://github.com/m4b/goblin/issues/411 too?

kkent030315:
> I guess this fixed #411 too?

Apparently it seems. Parsing [nvml.dll](https://github.com/m4b/goblin/issues/411#issue-2339840113) won't raise any errors now.

```bash
cargo run --example rdr -- nvml.dll
```

Issue #428, opened 2024-10-29
PE: import parser is broken for specially crafted binary


kkent030315:
Sample: [special_import_forwarder_tls.exe.zip](https://github.com/user-attachments/files/17556022/special_import_forwarder_tls.exe.zip)

# Reproducible steps

The sample is specially crafted and "valid" PE64 binary where it contains so-called "import forwarder TLS callback" where the TLS callbacks address points directly to the FT (First Thunk in `IMAGE_THUNK_DATA`) of import entry; and it fails to parse:

```bash
cargo run --example rdr -- "path/to/special_import_forwarder_tls.exe.bin"
```

```
[...]

PR #429, opened 2024-10-29
PE: Fix import parser for non-well-formed import table

kkent030315:
- Addresses issue #428 

Changelists:

- Added doc comment for `ImportDirectoryEntry::is_null`
- Added impl `ImportDirectoryEntry::is_possibly_valid` to filter out invalid entries where `name_rva` and `import_address_table_rva` are both zero. Those never be zero by design if the entry is valid.
- Added integration tests for both well-formed and non-well-formed import table parsing.

The change is straightforward, it does as well as how Windows Loader does. Parser should stop if it is not `is_possibly_valid` is the same behaviour with Windows Loader and should be more safe than before. This is special exception for non-well-formed case and should not affect well-formed binaries at all.

The same behaviour can be found at systeminformer (formely processhacker) implementation (respectively, peview):
https://github.com/winsiderss/systeminformer/blob/82c625783a035fa7eac355783f527bb53fb1a384/phlib/mapldr.c#L1635

This PR is backward compatible; thus minor changes.


src/pe/import.rs
```diff
@@ -171,13 +171,21 @@ pub struct ImportDirectoryEntry {
 pub const SIZEOF_IMPORT_DIRECTORY_ENTRY: usize = 20;
 
 impl ImportDirectoryEntry {
+    /// Whether the entire fields are set to zero
```
m4b, line 174: thank you for the doc!

src/pe/import.rs
```diff
@@ -129,7 +129,7 @@ impl<'a> SyntheticImportLookupTableEntry<'a> {
                     use self::SyntheticImportLookupTableEntry::*;
                     if bitfield.is_ordinal() {
                         let ordinal = bitfield.to_ordinal();
-                        debug!("importing by ordinal {:#x}", ordinal);
+                        debug!("importing by ordinal {:#x} ({})", ordinal, ordinal);
```
m4b, line 132: if you don't mind we should start porting this to `{ordinal:#x} ({ordinal})` whenever we encounter older style logs

m4b (approved):
This is great; if you don't mind, since you touched the logs, it'd be nice to see them ported to newer style, but if not, that's fine too!

src/pe/import.rs
m4b: you have a lot of PRs up, I don't think this should waste your time fixing it; can do in another pass

m4b:
NB backwards compatible with 0.9

Issue #371, opened 2023-06-06
PE: handle empty import address table


williballenthin:
This [PE file](https://www.virustotal.com/gui/file/0000bab1a3b04d3013d126e4621abfe7ec87ba0eafac01a53ac51c4db747f65a) has a valid  import lookup table but the pointer to the import address table is 0x0:

<img width="539" alt="image" src="https://github.com/m4b/goblin/assets/156560/a2764800-d2ec-45b4-b0e0-b66967cca6f3">

goblin currently supports handling empty import lookup tables but not empty import address tables:
https://github.com/m4b/goblin/blob/9f7fb6b1d68bd28d154c9d6587c1f132ce04cf54/src/pe/import.rs#L189-L191

Could we extend goblin to handle the above case (parse synthentic entries from the import lookup table when the import address table is 0x0)?

m4b:
If you understand the issue and might be willing to fix, a PR would be appreciated :)

PR #430, opened 2024-10-29
PE: Use OFTs for resolving imports without FTs

kkent030315:
- Addresses issue #371.

Sample: [sample.zip](https://github.com/user-attachments/files/17559872/sample.zip)

OFTs (Original First Thunk, aka lookup table) are kept as raw when mapped into virtual memory;
FTs (First Thunk, aka address table) are rewritten to the absolute address where the import function is located when mapped into virtual memory.

Static Import parsing would work fine on both cases, so use OFTs if FTs are zero, but keep FTs preffered.

### DIFF
--- a/src/pe/import.rs
+++ b/src/pe/import.rs
@@ -271,19 +271,15 @@ impl<'a> SyntheticImportDirectoryEntry<'a> {
         let import_address_table_offset =
             &mut utils::find_offset(rva as usize, sections, file_alignment, opts).ok_or_else(
                 || {
-                    error::Error::Malformed(
-                        if import_directory_entry.import_address_table_rva.is_zero() {
-                            format!(
-                                "Cannot map import_lookup_table_rva {:#x} into offset for {}",
-                                rva, name
-                            )
-                        } else {
-                            format!(
-                                "Cannot map import_address_table_rva {:#x} into offset for {}",
-                                rva, name
-                            )
-                        },
-                    )
+                    let target = if import_directory_entry.import_address_table_rva.is_zero() {
+                        "import_lookup_table_rva"
+                    } else {
+                        "import_address_table_rva"
+                    };
+                    error::Error::Malformed(format!(
+                        "Cannot map {} {:#x} into offset for {}",
+                        target, rva, name
+                    ))
                 },
             )?;
         let mut import_address_table = Vec::new();
--- a/src/pe/import.rs
+++ b/src/pe/import.rs
@@ -263,18 +263,29 @@ impl<'a> SyntheticImportDirectoryEntry<'a> {
             }
         };
 
-        let import_address_table_offset = &mut utils::find_offset(
-            import_directory_entry.import_address_table_rva as usize,
-            sections,
-            file_alignment,
-            opts,
-        )
-        .ok_or_else(|| {
-            error::Error::Malformed(format!(
-                "Cannot map import_address_table_rva {:#x} into offset for {}",
-                import_directory_entry.import_address_table_rva, name
-            ))
-        })?;
+        let rva = match import_directory_entry.import_address_table_rva.is_zero() {
+            true => import_directory_entry.import_lookup_table_rva,
+            false => import_directory_entry.import_address_table_rva,
+        };
+
+        let import_address_table_offset =
+            &mut utils::find_offset(rva as usize, sections, file_alignment, opts).ok_or_else(
+                || {
+                    error::Error::Malformed(
+                        if import_directory_entry.import_address_table_rva.is_zero() {
+                            format!(
+                                "Cannot map import_lookup_table_rva {:#x} into offset for {}",
+                                rva, name
+                            )
+                        } else {
+                            format!(
+                                "Cannot map import_address_table_rva {:#x} into offset for {}",
+                                rva, name
+                            )
+                        },
+                    )
+                },
+            )?;
         let mut import_address_table = Vec::new();
         loop {
             let import_address = bytes


src/pe/import.rs
```diff
@@ -263,18 +263,29 @@ impl<'a> SyntheticImportDirectoryEntry<'a> {
...
-        .ok_or_else(|| {
-            error::Error::Malformed(format!(
-                "Cannot map import_address_table_rva {:#x} into offset for {}",
-                import_directory_entry.import_address_table_rva, name
-            ))
-        })?;
+        let rva = match import_directory_entry.import_address_table_rva.is_zero() {
+            true => import_directory_entry.import_lookup_table_rva,
+            false => import_directory_entry.import_address_table_rva,
+        };
```
kkent030315, line 269: I am pretty sure I should check if `opts.resolve_rva` is `true` in match-false bracket (otherwise raises malformed as well as how it does before) in order to use address table which is replaced with absolute address of import address when the image is mapped.

```rs
let rva = match import_directory_entry.import_address_table_rva.is_zero() {
    true => Some(import_directory_entry.import_lookup_table_rva),
    false => opts.resolve_rva.then(|| import_directory_entry.import_address_table_rva),
};

if rva.is_none() {
    // Malformed
}
```

m4b: the second is somewhat harder to read at first, but i think I like it better :)

src/pe/import.rs
```diff
@@ -263,18 +263,29 @@ impl<'a> SyntheticImportDirectoryEntry<'a> {
...
+                                "Cannot map import_lookup_table_rva {:#x} into offset for {}",
+                                rva, name
+                            )
+                        } else {
+                            format!(
+                                "Cannot map import_address_table_rva {:#x} into offset for {}",
+                                rva, name
+                            )
+                        },
+                    )
```
m4b, line 286: this can be taken to extreme, but i definitely think this is better and less verbose:
```rust
let target = if import_directory_entry.import_address_table_rva.is_zero() { "import_lookup_table_rva" } else { "import_address_table_rva" };
error::Error::Malformed(format!("Cannot map {target} {rva:#x} into offset for {name}")
```

you could also have the rva branch above return a tuple of (rva, target) to use later on if there's an error, but this version probably better/cleaner/more local

src/pe/import.rs
m4b: when you say you're pretty sure, you mean that this PR needs that amendment? And are you worried about some kind of regression in doing this (i.e., less strict, problems further down the parsing pipeline or how come you didn't add it to this PR?)

m4b (commented):
I would alter with the suggestion, otherwise LGTM, assuming you want to make that additional change you mentioned

src/pe/import.rs
kkent030315: If I am correct that `resolve_rva` opt is meant to be _solely_ used for the mapped image (i.e., raw memory dumps, minidumps etc):

Yes, but what I am not sure is that we already make use of address table right now which is rewritten to the 64/32-bit absolute address when mapped into the memory. So If this is not to be the problem for this long in goblin, then I see no problems there to use lookup table.

I guess this is some kind of non-well implementation.. we should look for lookup tables rather than address tables.

- AddressTable/FTs: may not be valid in memory dumps.
  - Unlike PLT/GOT in ELF, In x86-64 arch and PE64, rel32 RVA of import calls like `call __imp_MessageBoxA` are actually points to the FTs.
- LookupTable/OFTs: always correct regardless of mapped images (i.e., for dependency walks)

I came to the conclusion that we do not need this optimization at this time.

kkent030315:
@m4b Thank you for the review! This PR is ready to go.

src/pe/import.rs
```diff
@@ -263,18 +263,25 @@ impl<'a> SyntheticImportDirectoryEntry<'a> {
...
-        .ok_or_else(|| {
-            error::Error::Malformed(format!(
-                "Cannot map import_address_table_rva {:#x} into offset for {}",
-                import_directory_entry.import_address_table_rva, name
-            ))
-        })?;
+        let rva = match import_directory_entry.import_address_table_rva.is_zero() {
+            true => import_directory_entry.import_lookup_table_rva,
+            false => import_directory_entry.import_address_table_rva,
+        };
```
m4b, line 269: just fyi matching on true/false in this case is somewhat unidiomatic, generally `if condition { } else { }` is preferred

m4b (approved):
Thanks for this! I would probably return the target_rva name as a tuple in the is_zero lookup for slightly less code/redundancy.

I don't fully understand semantically why we'd use the import_lookup_table when the address_table_rva is 0 for resolving the import_address_table offset but:
1. if it fixes an open bug
2. it doesn't regress any existing resolution for users then this seems ok (i don't think it can regress at the moment because if that value is zero it wouldn't get resolved iiuc, which would mean we'd return an error "cannot map import address table" in that case, whereas now we'll try the lookup table)

so the only real question i'd have is whether we'd return a kind of false positive in the case it's 0, and we return incorrectly an import_address_table_offset where we shouldn't?

kkent030315:
@m4b I'm the specialist in PE and I'd confident that it's tottaly fine to use lookup table in static binary parsers like goblin.

Yet the code behaves semantically what it was before, but only make exceptions such as the issue describes.

As I originally posted my request to provide the original binary in question, however that sample is 99%―mostly be specially crafted (packed) with private PE packers that no one else knows. First thunk (address table) is zero, means that it **semantically** does nothing other than Windows loader loads that dependency at execution but does not resolves the symbol. That is what makes me believe it is specially crafted and not for the usual cases where proprietary linkers do.

Anyways, thanks for the heads up!

m4b:
released in 0.9.3

Issue #350, opened 2022-12-23
add ability to parse OriginalFilename


theflakes:
MS doc: https://learn.microsoft.com/en-us/windows/win32/menurc/string-str?redirectedfrom=MSDN
Yara rule support for field: https://yara.readthedocs.io/en/v3.2.0/modules/pe.html

This is a useful field in threat hunting and forensics in general.

thanks

m4b:
Seems reasonable to me, would you like to make a PR adding this? (ideally backwards compatible/non breaking) :)

theflakes:
I can try at some point but its beyond my capabilities unfortunately. When I get some more time, I'll keep digging into it. Thanks

kkent030315:
FYI I'm working on this feature as a part of resource parser. The work is almost done and PR should be submitted in a few days I guess.

```
DEBUG - Found in section .rsrc(4), remapped into offset 0x7ffb0
DEBUG - resource_data data: Some(
    VersionInfo {
[...]

PR #431, opened 2024-10-31
PE: Add resource parser

kkent030315:
- Addresses issue #350.

Added resource parser implementation as proposed in the issue but with also manifests and fixed version info.

- Added parser for resource entries
- Added parser for `RT_VERSION`
- Added parser for `RT_MANIFEST`
- Added integration tests for the parsers


src/pe/resource.rs
```diff
@@ -0,0 +1,1433 @@
...
+///
+/// This function assumes that input bytes are multiple of `2`.
+pub(super) fn to_utf16_string(bytes: &[u8]) -> String {
+    let u16_slice = bytes
+        .chunks(2)
+        .map(|chunk| u16::from_le_bytes([chunk[0], chunk[1]]))
+        .take_while(|&wchar| wchar != 0)
+        .collect::<Vec<_>>();
+    String::from_utf16_lossy(&u16_slice)
+}
```
kkent030315, line 26: It's a lttle big confusing part, but by design (ref `ResourceString::parse`) it is guaranteed to be 2-bytes aligned within usage of this `resource` module. I believe theres better way to do this, or should I remove `alloc::string::String`s and let consumers impl higher implementations (thus keep our impl's lower impls)?

src/pe/resource.rs
```diff
@@ -0,0 +1,1557 @@
...
+            let data_entry = bytes.gread_with::<ResourceDataEntry>(&mut offset, scroll::LE)?;
+            let rva = data_entry.offset_to_data as usize;
+            offset = utils::find_offset(rva, sections, file_alignment, opts).ok_or_else(|| {
+                error::Error::Malformed(format!(
+                    "Cannot map ResourceDataEntry rva {:#x} into offset",
+                    rva
+                ))
+            })?;
+
+            let data = &pe[offset..offset + data_entry.size as usize];
```
kkent030315, line 1231: TODO: check malforemed

src/pe/resource.rs
```diff
@@ -0,0 +1,1557 @@
...
+
+        let rva = dd.virtual_address as usize;
+        let offset = utils::find_offset(rva, sections, file_alignment, opts).ok_or_else(|| {
+            error::Error::Malformed(format!(
+                "Cannot map ImageDebugDirectory rva {:#x} into offset",
+                rva
+            ))
+        })?;
+
+        if offset >= bytes.len() {
```
kkent030315, line 421: TODO: check malforemed (offset + size)

src/pe/resource.rs
```diff
@@ -0,0 +1,1557 @@
...
+        let len = bytes.gread_with::<u16>(offset, scroll::LE)?;
+        let value_len = bytes.gread_with::<u16>(offset, scroll::LE)?;
+        let r#type = bytes.gread_with::<u16>(offset, scroll::LE)?;
+        *offset = align_up(*offset, RESOURCE_STRING_FIELD_ALIGNMENT);
+        let key_size = &bytes[*offset..]
+            .chunks(2)
+            .take_while(|x| u16::from_le_bytes([x[0], x[1]]) != 0u16)
+            .count()
+            * SIZE_OF_WCHAR;
+        if (*offset - SIZE_OF_WCHAR) + key_size > bytes.len() {
```
kkent030315, line 723: TODO: check malforemed ((*offset - SIZE_OF_WCHAR) + key_size + SIZE_OF_WCHAR)

src/pe/resource.rs
```diff
@@ -0,0 +1,1433 @@
...
+        + core::ops::Not<Output = N>
+        + core::ops::BitAnd<Output = N>
+        + core::ops::Sub<Output = N>
+        + core::cmp::PartialEq
+        + core::marker::Copy,
+    u8: Into<N>,
+{
+    debug_assert!(align != 0u8.into(), "Align must be non-zero");
+    (value + align - 1u8.into()) & !(align - 1u8.into())
+}
```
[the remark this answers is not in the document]
kkent030315: I am pretty sure yes, as we have to have a duplicate function in CLR parser proposal at: https://github.com/m4b/goblin/pull/432/files#diff-cd5f7fb18a0ecca1bff11a2d304c2446f656baf68e045ae8ec0d78ba68600e8bR17-R31

src/pe/resource.rs
```diff
@@ -0,0 +1,1681 @@
...
+        if (*offset - SIZE_OF_WCHAR) + key_size + SIZE_OF_WCHAR > bytes.len() {
+            return Err(error::Error::Malformed(format!(
+                "offset ({:#x}) and key_size ({:#x}) is greater than bytes len {:#x}",
+                offset,
+                key_size,
+                bytes.len()
+            )));
+        }
+        let key =
+            &bytes[*offset - SIZE_OF_WCHAR..*offset - SIZE_OF_WCHAR + key_size + SIZE_OF_WCHAR];
```
m4b, line 744: this is a confusing offset computation, not sure it it can be simplified/extracted into subcomponents though :shrug:

src/pe/resource.rs
```diff
@@ -0,0 +1,1681 @@
...
+        self.r#type == 0
+    }
+
+    /// Converts [`ResourceString::key`] into a [`String`]
+    pub fn key_string(&self) -> String {
+        to_utf16_string(&self.key)
+    }
+
+    /// Converts [`ResourceString::value`] into a [`String`]
+    pub fn value_string(&self) -> String {
```
m4b, line 789: should this be `-> Option<String>`, since if `is_binary_data` is true, a string representation doesn't make sense?

src/pe/resource.rs
```diff
@@ -0,0 +1,1681 @@
...
+/// - `04`: Primary language identifier for English.
+/// - `09`: Sub-language identifier for United States.
+/// - `04E4`: Codepage identifier for Unicode (UTF-16).
+///
+/// This value may present in [`ResourceString::key`] without no dedicated value data.
+pub const VERSION_INFO_US_ENGLISH_UNICODE: &str = "040904E4";
+/// A [`ResourceString::key`] of [`VsFixedFileInfo`]
+pub const VS_VERSION_INFO_KEY: &str = "VS_VERSION_INFO";
+
+impl<'a> VsFixedFileInfo {
```
m4b, line 947: I'm surprised this didn't warn, isn't this `<'a>` unbound in this impl block?

src/pe/resource.rs
```diff
@@ -0,0 +1,1681 @@
...
+/// file description, version information, and copyright details. Each field is optional and
+/// can be absent if the information is not available.
+#[derive(Copy, Clone)]
+pub struct StringFileInfo<'a> {
+    /// Additional information for diagnostic purposes. Can be of arbitrary length.
+    pub comments: Option<&'a [u8]>,
+    /// The name of the company that produced the file, e.g., "Microsoft Corporation".
+    pub company_name: Option<&'a [u8]>,
+    /// A description of the file suitable for presentation to users, e.g., "Keyboard driver for AT-style keyboards".
+    pub file_description: Option<&'a [u8]>,
```
m4b, line 976: so in general, for these utf16 strings, I would in general either pre-parse them (especially if this doesn't occur a lot) to the utf8 owned String, and keep the fields `pub`; _or_ i would make these fields *not* `pub` and only expose the functions returning the parsed version.

in the case of this struct, since there are so many fields, I would almost prefer to pre-parse it, but maybe only if expect there to e.g,. be only one copy of this struct in the resource data; if it's in a vector or iterator and could have a perf issue due to eager pre-parsing, and maybe no one uses, then we can make it private, and have the function logic like you have below

src/pe/resource.rs
```diff
@@ -0,0 +1,1681 @@
...
+    pub fn special_build(&self) -> Option<String> {
+        self.special_build.map(|x| to_utf16_string(x))
+    }
+}
+
+/// Represents a version information
+#[derive(Copy, Clone)]
+pub struct VersionInfo<'a> {
+    /// Raw data of entire [`RT_VERSION`] area.
+    pub data: &'a [u8],
```
m4b, line 1122: in general, unless you have a specific usecase for this, I like to hide the data fields that need to tuck it away for whatever reason; it also has the benefit of making the struct not constructible, which also means adding a *new* field is backwards compatible, whereas, right now, since every field is pub, it is not.

Note however perhaps you don't anticipate adding a new field, *ever*, in which case keeping it constructible like this (every field pub) is fine;

src/pe/resource.rs
```diff
@@ -0,0 +1,1681 @@
...
+    pub fn parse(
+        pe: &'a [u8],
+        bytes: &'a [u8],
+        it: ResourceEntryIterator<'a>,
+        sections: &[section_table::SectionTable],
+        file_alignment: u32,
+        opts: &options::ParseOptions,
+    ) -> error::Result<Option<Self>> {
+        let it = it.collect::<Result<Vec<_>, _>>()?;
+        let it = it.iter().find(|e| e.id() == Some(RT_VERSION));
```
m4b, line 1152: iiuc, you are finding the element that matches RT_VERSION; you shouldn't have to collect to do this, which is an extra allocation

src/pe/resource.rs
```diff
@@ -0,0 +1,1681 @@
...
+    pub fn parse(
+        pe: &'a [u8],
+        bytes: &'a [u8],
+        it: ResourceEntryIterator<'a>,
+        sections: &[section_table::SectionTable],
+        file_alignment: u32,
+        opts: &options::ParseOptions,
+    ) -> error::Result<Option<Self>> {
+        let it = it.collect::<Result<Vec<_>, _>>()?;
+        let it = it.iter().find(|e| e.id() == Some(RT_MANIFEST));
```
m4b, line 1230: ditto here; all of this code looks very similar, with only minor variations; my brain immediately wants to abstract it somehow, but it will likely have very little gain, and likely only add complication :)

m4b (changes requested):
this is a huge PR! Thank you for adding this though.

Some general feedback:

1. There is a *lot* of new pub fields, pub structs and pub functions
     a. I believe i've mentioned in past, but to reiterate: goblin is very conservative with `pub`, can consider it in your mind as basically 1.0, but occasionally letting through breaking changes; with that in mind, it's important to ask yourself/ourselves whether:

       i. will the new pub structs/enums, etc., have new fields, either by addition by the binary stakeholders (microsoft, linux (i guess?), apple), or by some current oversight; if yes, or likely to have new fields added, either needs to be marked `#[non_exhaustive]`, or a field hidden so that it isn't constructible, and so adding a new field is legal. the owners of the format adding new fields is unlikely, but it has happened during the existence of this crate (e.g., apple adding new binary variants, new `LC_COMMANDS`, etc.)
       ii. do you _really_ need a field `pub`? similarly with a function; if yes, that's fine, but if any doubt, start with private/`pub(crate)` and we can make it pub later if needed; once it's pub it generally can't go back.

2. I haven't investigate in detail, but I assume a lot of the offset math is checked/etc., before doing any kind of unprotected/panicable `&[offset..foo]`; from what I've seen it all looks good, just want to double check, and I know you've fixed that in the past, so i suspect you already know this one.
3. be aware of allocations, etc., especially temporary ones, and as i mentioned, probably need to make a decision for the utf16 strings, as I mentioned in some comments.

Anyway, this is great work, other than some fairly minor nits, this looks pretty good to go, but just wanted to get some of the pub fields that don't need to be pub fixed, and etc.

Thank you for your patience and this amazing work!

src/pe/resource.rs
```diff
@@ -0,0 +1,1681 @@
...
+    /// Reinterprets the struct as [`u64`]
+    pub fn value(&self) -> u64 {
+        ((self.name_or_id) as u64) << 32 | self.offset_to_data_or_directory as u64
+    }
+
+    /// Checks if the resource name is a string.
+    ///
+    /// Returns `true` if the name is a string, otherwise `false`.
+    pub fn name_is_string(&self) -> bool {
+        self.name_or_id & IMAGE_RESOURCE_NAME_IS_STRING != 0
```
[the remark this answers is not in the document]
kkent030315: Yes. It is exactly as-is defined in Windows SDK. So we better keep them.

src/pe/resource.rs
```diff
@@ -0,0 +1,1681 @@
...
+    pub fn parse(bytes: &'a [u8], offset: &mut usize) -> error::Result<Self> {
+        let len = bytes.gread_with::<u16>(offset, scroll::LE)?;
+        let value_len = bytes.gread_with::<u16>(offset, scroll::LE)?;
+        let r#type = bytes.gread_with::<u16>(offset, scroll::LE)?;
+        *offset = align_up(*offset, RESOURCE_STRING_FIELD_ALIGNMENT);
+        let key_size = &bytes[*offset..]
+            .chunks(2)
+            .take_while(|x| u16::from_le_bytes([x[0], x[1]]) != 0u16)
+            .count()
+            * SIZE_OF_WCHAR;
```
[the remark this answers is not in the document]
kkent030315: Your proposal seems brilliant as a syntax! On the other hand I have different perspective that it's overkill for this resource parser. Nothing else but `RT_VERSION` uses this kind of utf-16 string parsing, I think it's fine to keep it alone. While I can implement your proposal, do you think it should still have `Str16`?

src/pe/resource.rs
kkent030315: We should remove `pub` from `data` in this struct for sure. Yet this struct will *never* be changed _unless_ Microsoft did something with it. :)

src/pe/resource.rs
kkent030315: It is still be able to express like following but I don't like that kind of ugly expression:

```rs
    ) -> error::Result<Option<Self>> {
        let mut maybe_entry = None;
        for x in it {
            if x?.id() == Some(RT_VERSION) {
                entry = Some(x?);
                break;
            }
        }

        if let Some(entry) = maybe_entry {
```

This is semantically _not_ equivalent since it hides the error in while converting `Result<T,E>` into `Option<T>` I really don't like:

```rs
        if let Some(entry) = it
            .filter_map(Result::ok)
            .find(|x| x.id() == Some(RT_VERSION))
```

So that I concluded to following implementation:

```rs
impl<'a> ResourceEntryIterator<'a> {
    /// Find the resource entry by its resource ID.
    pub fn find_by_id(&self, id: u16) -> error::Result<Option<ResourceEntry>> {
        self.map(|x| {
            x.and_then(|x| {
                if x.id() == Some(id) {
                    Ok(Some(x))
                } else {
                    Ok(None)
                }
            })
        })
        .find_map(Result::transpose)
        .transpose()
    }
}
```

```rs
    ) -> error::Result<Option<Self>> {
        if let Some(entry) = it.find_by_id(RT_VERSION)? {
```

This also provides shortcut for consumers as `find_by_id` is `pub`.
What do you think about this?

kkent030315:
I found a bug related to an alignment. It seems a buffer passed to `ResourceStringIterator` sometimes aligned with 8 bytes where the case has paddings at the tail (e.g., `RT_VERSION` in `ntdll.dll`), but sometimes it doesn't (e.g., `PYTHON_INSTALLER_VERSION_INFO`).

```
DEBUG - Parsed next resource string as size 0x24: ResourceString {
    len: 0x24,
    value_len: 0x4,
    type: 0 (Binary Data),
    key: "Translation",
    key_slice: [54, 00, 72, 00, 61, 00, 6e, 00, 73, 00, 6c, 00, 61, 00, 74, 00, 69, 00, 6f, 00, 6e, 00] (22 bytes),
    value_slice: [09, 04, b0, 04] (4 bytes, Correct),
}
DEBUG - 4 bytes remaining
type is too big (2) for 0
```

m4b:
is this in a reviewable state ? i can't remember where we left off exactly, a lot of PRs lately :)

src/pe/resource.rs
```diff
@@ -0,0 +1,1681 @@
...
+pub const RT_ANICURSOR: u16 = 21;
+/// Windows resource type identifier for animated icons.
+pub const RT_ANIICON: u16 = 22;
+/// Windows resource type identifier for HTML resources.
+pub const RT_HTML: u16 = 23;
+/// Windows resource type identifier for manifests.
+pub const RT_MANIFEST: u16 = 24;
+
+/// Represents an image resource directory in the PE (Portable Executable) format.
+#[repr(C)]
```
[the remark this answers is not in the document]
kkent030315: I am pretty sure it needs `#[repr(C)]` as this is a representation of `IMAGE_RESOURCE_DIRECTORY` in `Windows.h`.

https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/System/SystemServices/struct.IMAGE_RESOURCE_DIRECTORY.html

m4b:
looks like CI is failing; rebase?

m4b:
i will make it priority to review this and get this in once CI passing; i appreciate your patience! please feel free to aggressively ping me if i do not respond :)

src/pe/resource.rs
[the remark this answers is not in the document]
kkent030315: You are right, it's really not meant to be "for FFI" but is "for layout control", and it's `pub` struct and is clear representation of `VS_FIXEDFILEINFO` where originally defined in C. If we removed it from the struct it no longer be able to use for FFI by consumers. If that's acceptable, I'd 100% agree to remove `#[repr(C)]` from this struct.

src/pe/resource.rs
```diff
@@ -0,0 +1,1809 @@
...
+/// Helper for parsing `wchar_t` utf-16 strings in a safe manner.
+#[derive(Debug, Copy, Clone, Default, PartialEq)]
+pub struct Utf16String<'a>(&'a [u8]);
+
+impl<'a> Utf16String<'a> {
+    /// Converts underlying bytes into an owned [String].
+    pub fn to_string(&self) -> String {
+        let len = self.0.len() / 2;
+        let ptr = self.0.as_ptr() as *const u16;
+        // Safety: `self.0` is assumed properly aligned and has even length.
```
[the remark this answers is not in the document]
kkent030315: I just made it return an `Option<String>`, with some tests added.

kkent030315:
@m4b Thank you so much for your review.. it's being >2,000 lines of code. Everything you've concerned should be addressed now. Did double check on my end as well.

src/pe/resource.rs
```diff
@@ -0,0 +1,1942 @@
...
+                ))
+            })?;
+
+            let data = pe
+                .pread_with::<&[u8]>(offset, data_entry.size as usize)
+                .map_err(|_| {
+                    error::Error::Malformed(format!("Cannot map RT_VERSION at {offset:#x}"))
+                })?;
+            let iterator = ResourceStringIterator { data };
+            let strings = iterator.collect::<Result<Vec<_>, _>>()?;
```
m4b, line 1225: perf: i'm not sure you have to allocate here if you just use it to generate fixed_info?

src/pe/resource.rs
```diff
@@ -0,0 +1,1942 @@
...
+    /// - When [`ResourceString::type`] indicates string data: multiply this field with [`SIZE_OF_WCHAR`] that should be the actual size of [`ResourceString::value`] with null-terminator.
+    /// - Othereise, treat as-is.
+    pub value_len: u16,
+    /// The type of [`ResourceString::value`] in the version resource.
+    ///
+    /// This member is `1` if the version resource contains text data;
+    /// and `0` if the version resource contains binary data, otherwise sometimes an invalid value.
+    pub r#type: u16,
+    /// An arbitrary null-terminated utf-16 unicode string.
+    pub key: Utf16String<'a>,
```
m4b, line 722: this is the only place Utf16String is used (publicly, afaics); i would really prefer to keep this pub(crate) for now since once this datatype escapes out into pub, it's very hard to change. Since the other uses have methods to just get the key_string and value_string out, I think it's ok to remove pub from this for now until we've really nailed down the Utf16String naming and representation, etc.

m4b (commented):
Sorry I'd really like to keep Utf16String non-pub for now if we can help it, as this is already adding a ton of new public datatypes. After this, I think we are safe to merge this; thank you for your patience!

src/pe/resource.rs
```diff
@@ -18,16 +18,39 @@ pub(super) const SIZE_OF_WCHAR: usize = core::mem::size_of::<u16>();
...
+            .collect::<Result<String, _>>();
+            Some(result.unwrap_or_else(|_| {
+                let u16_vec = bytes
+                    .chunks_exact(2)
+                    .map(|chunk| u16::from_le_bytes([chunk[0], chunk[1]]))
+                    .take_while(|&val| val != 0)
+                    .collect::<Vec<_>>();
+                String::from_utf16_lossy(&u16_vec)
+            }))
+        }
```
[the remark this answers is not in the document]
kkent030315: Ok I think it would suffice to just use the older one. Agreed

src/pe/resource.rs
kkent030315: Unfortunately yes, we need it to find a `VS_VERSION_INFO` entry.

src/pe/resource.rs
```diff
@@ -0,0 +1,1922 @@
...
+pub(super) fn to_utf16_string(bytes: &[u8]) -> Option<String> {
+    if bytes.len() % 2 != 0 {
+        return None;
+    }
+    let u16_chars = bytes
+        .chunks_exact(2)
+        .map(|chunk| u16::from_le_bytes([chunk[0], chunk[1]]))
+        .take_while(|&wchar| wchar != 0)
+        .collect::<Vec<_>>();
+    Some(String::from_utf16_lossy(&u16_chars))
```
m4b, line 28: you could also do `from_utf16(&u16_chars).ok()` if you don't want the lossy part; not sure the exact semantics you want, but since it's already Option, probably `from_utf16(&u16_chars).ok()` is better here?

src/pe/resource.rs
```diff
@@ -0,0 +1,1922 @@
...
+}
+
+/// Helper for parsing `wchar_t` utf-16 strings in a safe manner.
+#[derive(Debug, Copy, Clone, Default, PartialEq)]
+pub struct Utf16String<'a>(&'a [u8]);
```
m4b, line 33: this is still `pub`; since key is now `pub(crate)` we can do the same and then we can merge. Thank you for removing the unsafe  code!

m4b (approved):
Thanks for your patience; the Utf16String is still pub, I think you forgot to add that after marking key pub, but otherwise this is ready to go, thanks for your patience! If you're too busy I can merge and fix it myself since it's very minor, but I'd prefer if possible to have it all in one patch. thank you again for your work on this!

m4b (approved):
Huzzah! Thank you for all your patience ! Great stuff !

Issue #433, opened 2024-11-03
PE: TLS parser generics and bitness-related rework


kkent030315:
Reference: https://github.com/m4b/goblin/pull/425#issuecomment-2436724192

TL;DR:
- Deprecate `TlsData::parse`, `TlsData::parse_with_opts` that uses unintuitive generics.
  - Instead, we should ideally have `is_64`.
- Make independent `ImageTlsDirectory(64)` and `ImageTlsDirectory32`, and `impl <From<ImageTlsDirectory32>> for ImageTlsDirectory(64)` to merge.

PR #435, opened 2024-11-10
PE: TLS parser rework

kkent030315:
- #433 

Changelists:
- Removed generics from `ImageTlsDirectory::parse*` and `TlsData::parse*`, instead added new argument `is_64` as proposed in the issue.
- Added slice bound check in `TlsData::raw_data` parser to compliant with panic-free design in goblin.

I personally prefer not making `ImageTlsDirectory32`―as it's more comfortable than the way duplicating the same kind of structs and merging to 64-bit as proposed in the issue.

This change should be marked as breaking-change and I expect this to be rolled up in 0.10.

### DIFF
--- a/src/pe/mod.rs
+++ b/src/pe/mod.rs
@@ -221,25 +221,15 @@ impl<'a> PE<'a> {
             }
 
             if let Some(tls_table) = optional_header.data_directories.get_tls_table() {
-                tls_data = if is_64 {
-                    tls::TlsData::parse_with_opts::<u64>(
-                        bytes,
-                        image_base,
-                        tls_table,
-                        &sections,
-                        file_alignment,
-                        opts,
-                    )?
-                } else {
-                    tls::TlsData::parse_with_opts::<u32>(
-                        bytes,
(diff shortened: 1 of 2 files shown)


m4b (commented):
this looks good, but I don't see where `is_64` gets computed? i only see it passed to `parse_with_opts` ?

kkent030315:
@m4b `is_64` is already computed in [`PE::parse_with_opts`](https://github.com/kkent030315/goblin/blob/63f47dfba934dd86d9a99cc1089f74e5c123a5d7/src/pe/mod.rs#L231) and passed straight to the `TlsData::parse_with_opts`.

m4b (approved):
thank you for making these changes, the public api is more idiomatic again

m4b:
NB: breaking change

BinFlip:
Nice! I was about to open a PR to fix an out of bound access that happened in the TLS parser when fuzzing my project, but it looks like this commit already took care of it vs the officially released '0.9.3' that is on crates.io

PR #436, opened 2024-11-15
elf: fix is_import to check st_shndx

### LINKED ISSUE #288: is_import() incorrectly checks st_value == 0
You want to be checking whether it's global/weak and has a shndx of 0 (SHN_UNDEF)

### LINKED ISSUE #419: `is_import` is wrong on MIPS ELF
Hello,
thank you very much for `goblin`, it is awesome. I've developed a tool called `findso`, in my [rsbkb](https://github.com/trou/rsbkb) project, using your crate, using `!is_import` to identify if a given symbol is exported or not. It looks for which `.so` implements a given function. For example:
```
$ findso -p /usr/arm-linux-gnueabi/lib/ -a memcpy
/usr/arm-linux-gnueabi/lib/libc.so.6
```

However, on mips I have the following result:
```
$ findso -p /usr/mips-linux-gnu/lib/ -a memcpy
/usr/mips-linux-gnu/lib/libatomic.so.1
/usr/mips-linux-gnu/lib/libatomic.so.1.2.0
/usr/mips-linux-gnu/lib/libc.so.6
...
```

readelf gives the following:
```
$ mips-linux-gnu-readelf -D -a /usr/mips-linux-gnu/lib/libnss_compat.so.2 | grep memcpy
    47: 00007d90     0 FUNC    GLOBAL DEFAULT  UND memcpy@GLIBC_2.0 (4)
  00020188 -32600(gp) 00007d90 00007d90 FUNC    UND memcpy
$ mips-linux-gnu-readelf -D -a /usr/mips-linux-gnu/lib/libc.so.6 | grep memcpy
   862: 000a75b0   984 FUNC    GLOBAL DEFAULT   13 memcpy@@GLIBC_2.0
```

I think there's some MIPS weirdness (as usual), where UND symbols have a global value?

I don't know if there's a simple and reliable way of checking if a symbol is actually defined or not.

Any idea?

Some sample files can be found in the following Debian package:
https://packages.debian.org/bookworm/all/libc6-mips-cross/filelist

Berrysoft:
Closes #288 

Closes #419 

I also found this bug on with ELF produced on illumos. `st_value` may not be zero because it may indicate the offset of a import function.

### DIFF
--- a/src/elf/sym.rs
+++ b/src/elf/sym.rs
@@ -94,9 +94,9 @@ pub fn st_visibility(other: u8) -> u8 {
 
 /// Is this information defining an import?
 #[inline]
-pub fn is_import(info: u8, value: u64) -> bool {
+pub fn is_import(info: u8, shndx: u64) -> bool {
     let bind = st_bind(info);
-    bind == STB_GLOBAL && value == 0
+    (bind == STB_GLOBAL || bind == STB_WEAK) && shndx == SHN_UNDEF as _
 }
 
 /// Convenience function to get the &'static str type from the symbols `st_info`.
@@ -170,8 +170,7 @@ macro_rules! elf_sym_std_impl {
             /// Checks whether this `Sym` has `STB_GLOBAL`/`STB_WEAK` bind and a `st_value` of 0
             #[inline]
             pub fn is_import(&self) -> bool {
-                let bind = self.st_info >> 4;
-                (bind == STB_GLOBAL || bind == STB_WEAK) && self.st_value == 0
+                is_import(self.st_info, self.st_shndx as _)
             }
             /// Checks whether this `Sym` has type `STT_FUNC`
             #[inline]
@@ -336,6 +335,8 @@ use core::fmt;
 use scroll::ctx;
 use scroll::ctx::SizeWith;
 
+use super::section_header::SHN_UNDEF;
+
 #[derive(Clone, Copy, PartialEq, Default)]
 /// A unified Sym definition - convertible to and from 32-bit and 64-bit variants
 pub struct Sym {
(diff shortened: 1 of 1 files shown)


src/elf/sym.rs
```diff
@@ -94,9 +94,9 @@ pub fn st_visibility(other: u8) -> u8 {
 
 /// Is this information defining an import?
 #[inline]
-pub fn is_import(info: u8, value: u64) -> bool {
+pub fn is_import(info: u8, shndx: u64) -> bool {
```
m4b, line 97: idly wondering if we should de-pub this along with the other functions; i wonder why they were made `pub`, could have been a historical accident or perhaps someone uses them for whatever reason; i don't see bingrep for example using it anywhere

src/elf/sym.rs
```diff
@@ -94,9 +94,9 @@ pub fn st_visibility(other: u8) -> u8 {
...
     let bind = st_bind(info);
-    bind == STB_GLOBAL && value == 0
+    (bind == STB_GLOBAL || bind == STB_WEAK) && shndx == SHN_UNDEF as _
```
m4b, line 99: this is a good catch (the missing STB_WEAK), but changing this function to operate semantically on `shndx` values instead of `st_value`, while backwards compatible, has wider implications.

for example, any clients relying on `st_value` being used in these computations will now break when section headers are stripped.

imho my general thesis that what a dynamic linker can import is how we've defined import, while arguably more nuanced than this, is a good one and has value (no pun intended). using `st_value` is a heuristic for that thesis.

that being said, it seems like people are hitting some issues with this, so we can consider adding something like `is_import_from_sections` or something which is new api to opt into using the section header (perhaps along with st_value) as the determinant.

m4b (changes requested):
As mentioned in comment, swapping out these functions to suddenly use shndx doesn't seem right to me; adding the STB_WEAK check in the is_import function is good catch (though it now has me wondering whether all these functions should be marked private).
However, we can review adding a new api that uses the shndx in another patch if that is something people really want, or some other design that incorporates it somehow.

Issue #440, opened 2024-12-16
PE: image base silently overflows parsing PE64 on 32-bit builds


kkent030315:
As pointed out in [this comment](https://github.com/m4b/goblin/pull/426#discussion_r1855327540) the [`goblin::pe::PE::image_base`](https://docs.rs/goblin/0.9.2/goblin/pe/struct.PE.html#structfield.image_base) _silently_ overflowing when parsing PE64 on 32-bit goblin.

| **Target** | Sizeof **`usize`** | Sizeof **`u64`** |
|------------------|------------------|----------------|
| x86 (32-bit)     | 4 bytes (32-bit) | 8 bytes (64-bit) |
| x64 (64-bit)     | 8 bytes (64-bit) | 8 bytes (64-bit) |

https://github.com/m4b/goblin/blob/48da3d867b47173b19072b30bd3ef21e7d0215ba/src/pe/mod.rs#L61-L62

The [`goblin::pe::PE::image_base`](https://docs.rs/goblin/0.9.2/goblin/pe/struct.PE.html#structfield.image_base) must be defined as `u64` to be compatible with 32-bit hosts.

## Steps to reproduce

For convenience, patching `examples\rdr.rs` may improve readability specifically for this case since theres a bunch of irrelevant logs:

```diff
fn run() -> error::Result<()> {
    for (i, arg) in env::args().enumerate() {
        if i == 1 {
            let path = Path::new(arg.as_str());
            let buffer = fs::read(path)?;
            let res = goblin::Object::parse(&buffer)?;
-           //println!("{:#?}", res);
+           match res {
+              goblin::Object::PE(x) => {
+                  println!("{:#x}", x.image_base)
+              }
+              _ => {}
+           }
        }
    }
    Ok(())
}
```

Then run:

```
cargo run --target i686-pc-windows-msvc --example rdr -- "path/to/pe64-bin.exe"
```

If the image base of `path/to/pe64-bin.exe` is `0x140000000` (the default value of COFF linkers), it outputs trimmed 32-bit value `0x40000000`.

This is not particular to `i686-pc-windows-msvc` triple but any other 32-bit builds.

kkent030315:
note: TE has no problem with this.

https://github.com/m4b/goblin/blob/48da3d867b47173b19072b30bd3ef21e7d0215ba/src/pe/header.rs#L1196-L1197

kkent030315:
This is also a mess. it is defined as 32-bit on both PE32/64.

[...]

Issue #434, opened 2024-11-03
Tracking issue for 0.10


kkent030315:
This is a desirable tracking issue for 0.10.x to rollup breaking changes as possible as it can.

## Breaking Changes

- [ ] #433 

## Features

- [x] #406
- [ ] #350 
  - [ ] #431 
- [ ] #315 
  - [ ] #432 

## Bug Fixes

- [ ] #428 
  - [ ] #429
- [ ] #371 
  - [ ] #430 

## Misc

- [ ] #426

m4b:
@kkent030315 thanks for creating this tracking issue, it is very helpful!

Please note my bandwidth at the moment is not super high, and I appreciate your patience and also all your work here! So if some PRs are more important than others to review, or e.g,. if there are some non-breaking changes, please help me a bit to point me to those first if you like, since there are quite a lot of PRs here :)

m4b:
@kkent030315 thanks for all these PRs, so 426 is reviewed and ready to merge, just would like to have the C source merged. After that, if you don't mind, could you give me another abbreviated list of PRs you'd like to see prioritized for review/merge?

kkent030315:
@m4b Hi! I'd prefer to finish #403. :)

PR #441, opened 2024-12-16
PE: fix address size incompatibility on 32-bit builds

kkent030315:
- #440 

Since `PE::entry`, `PE::image_base` and `StandardFields::address_of_entry_point` are all `pub`s this PR is not backward compatible.


kkent030315:
@m4b Hi there, could I get your review for this PR?
btw happy new year!

src/pe/optional_header.rs
```diff
@@ -119,7 +119,7 @@ pub struct StandardFields {
     /// * For device drivers, this is the address of the initialization function.
     ///
     /// The entry point function is optional for DLLs. When no entry point is present, this member is zero.
-    pub address_of_entry_point: u64,
+    pub address_of_entry_point: u32,
```
m4b, line 122: this seems like the only actually required fix, but even that is somewhat unclear to me? everything else seems to push the usize <-> u64/u32 casting down into the `pe::utils::find_offset`, which I admit does seem more elegant.

However, there was a certain clarity of passing a `usize` into these parse functions since they are ultimately used to compute 32/64-bit addresses/offsets.

Could you elaborate more on why this gets downgraded to a u32 though?

src/pe/mod.rs
```diff
@@ -56,10 +56,10 @@ pub struct PE<'a> {
...
     /// Whether the binary is 64-bit (PE32+)
     pub is_64: bool,
-    /// the entry point of the binary
-    pub entry: usize,
-    /// The binary's RVA, or image base - useful for computing virtual addreses
-    pub image_base: usize,
+    /// The entry point RVA of the binary
+    pub entry: u32,
+    /// The binary's VA, or image base - useful for computing virtual addreses
+    pub image_base: u64,
```
m4b, line 62: is this the actual bug? (that we used a `usize` but on 32-bit systems, this can be `u64`?)

If yes, is it really required for `entry` to be fixed to u32? If there's a compelling reason (For consistency perhaps) then it seems fine, but this is a pretty big breaking change, but it may be worth it if it prevents future issues.

m4b (commented):
this seems generally ok but i'd like confirmation on the bug location; this is also a pretty big breaking change (Fields in PE), so i'd like to make sure this doesn't add regressions and it's going to be right for the foreseeable future, etc.

m4b:
@kkent030315 thanks for this, and happy new year to you! Apologies for radio silence, i've been offline for a bit, but should be able to get a lot of these patches reviewed and merged in next few days, appreciate your patience as always

src/pe/optional_header.rs
kkent030315: Theoretically, this `u64` is _fine_. However,

1) Its 4 bytes half is always wasted because this is defined as `u32` as per PE specification[^1].

2) And as you mentioned, the `AddressOfEntryPoint` is always an RVA (`u32`), which is treated as `u32` as well, so consumers are forced to make unnecessary downcast if it is defined as `u64`.

[^1]: https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-image_optional_header64

src/pe/mod.rs
kkent030315: Yes, this is a serious hidden bug for long.

Fortunately, RVA calculations (i.e., [TLS callbacks map](https://github.com/m4b/goblin/blob/19d88eccfb96a8bce8ff086e9a1bfccd10f645ca/src/pe/tls.rs#L255-L256)) work even it overflows. (i.e., `0x140000000u64` overflows and trimmed into `0x40000000u32`, then subtract the `u32` RVA)

On the other hand, this is significant correctness issue likewise whenever the consumer wants to use it for something like verification.

I guess this has been invisible for this long because no one really cares about 32-bit builds.

src/pe/tls.rs
```diff
@@ -279,11 +279,11 @@ impl<'a> TlsData<'a> {
                     )));
                 }
                 // Each callback is an VA so convert it to RVA
-                // For x86 compatibility, `usize` is 32-bit, `u64` is 64-bit.
-                // Therefore upcast to u64 first, then downcast the whole var to u32.
-                let callback_rva = (callback - image_base as u64) as usize;
+                let callback_rva = callback - image_base;
                 // Check if the callback is in the image
-                if utils::find_offset(callback_rva, sections, file_alignment, opts).is_none() {
+                if utils::find_offset(callback_rva as usize, sections, file_alignment, opts)
```
m4b, line 284: isn't there technically still an issue on 32-bit systems if `callback_rva is > u32`, and then it gets casted to a `usize` (hence truncating a potentially 64 bit number to 32 bits (sizeof pointer))?

tbh i'm not sure we ever supported analyzing 64-bit binaries on on 32-bit systems, simply due to rust `&[u8]` requiring usize to index, which means if that slice is larger than 32-bits, you cannot access the upper bits in rust.

m4b (approved):
I think we can merge this, and while it's nice to solidify the fields as their explicit bit sizes, i don't think fully supporting 64-bit binary analysis on 32-bit machines has ever been a goal, but we can try our best.

I'm hoping this upcoming release will be one of the last in terms of major field breakage and other things in goblin, and I know i've said this many times in the past, but I'm hoping to put out a 1.0 release in the near future, but first would like to fixup and merge issues like this for better future proofing.

since you've made so many great changes in the PE side of things, if you could keep this in mind (a potential 1.0 and hence no api breakage after) that would be great.

m4b:
NOTE: breaking

m4b:
ref #434

src/pe/tls.rs
kkent030315: Oh yes, where the case the image must be broken. We shouldn't assume any of TLS callback addresses to be a valid VA/RVA. I'll put the check there.

PR #442, opened 2025-01-09
pe.imports: ignore malformed imports in ParseMode::Permissive

ideeockus:
In some PE files one part of imports may be corrupted, but another part of imports is ok, and I want to be able parse this files

### DIFF
--- a/src/pe/options.rs
+++ b/src/pe/options.rs
@@ -40,4 +40,9 @@ impl ParseOptions {
             parse_mode: ParseMode::Strict,
         }
     }
+
+    pub fn with_parse_mode(mut self, parse_mode: ParseMode) -> Self {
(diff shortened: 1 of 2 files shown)


src/pe/options.rs
```diff
@@ -40,4 +40,9 @@ impl ParseOptions {
             parse_mode: ParseMode::Strict,
         }
     }
+
+    pub fn with_parse_mode(mut self, parse_mode: ParseMode) -> Self {
```
m4b, line 44: nit would prefer pub items to be doc'd, even trivial ones

m4b (approved):
LGTM, i like it

m4b:
NB: non-breaking, new pub item

PR #444, opened 2025-01-17
PE: Add base relocation parser

kkent030315:
It was a little bit surprise for me that goblin does not have a parser for base relocations in PE besides COFF relocations.

- Added parser for `IMAGE_DIRECTORY_ENTRY_BASERELOC` directory.
- Added COFF relocation type constants for ARM, ARM64 and IA64.
- Added PE generic relocations type constants. (COFF and PE relocations are different things)


kkent030315:
@m4b Hi there, I'd like to get review and see if this worth merging. Thank you very much for your attention (as always)!

src/pe/relocation.rs
```diff
@@ -131,3 +308,364 @@ impl<'a> Iterator for Relocations<'a> {
         }
     }
 }
+
+#[derive(Debug, PartialEq, Clone, Default)]
+pub struct RelocationData<'a> {
```
m4b, line 313: nit would prefer all these pub items to be doc'd

src/pe/relocation.rs
```diff
@@ -131,3 +308,364 @@ impl<'a> Iterator for Relocations<'a> {
...
+        if offset + dd.size as usize > bytes.len() {
+            return Err(error::Error::Malformed(format!(
+                "base reloc offset {:#x} and size {:#x} exceeds the bounds of the bytes size {:#x}",
+                offset,
+                dd.size,
+                bytes.len()
+            )));
+        }
+
+        let bytes = &bytes[offset..offset + dd.size as usize];
```
m4b, line 359: it doesn't matter now but i wonder if a better pattern here to use is something like:

```rust
let size = dd.size as usize;
let bytes = bytes[offset..].pread::<&[u8]>(size).map_err(|_| TheErrorWeWantToDisplay)?;
```

it's just easy to copy-pasta the bounds error, and slice the wrong bounds that were different than what we checked, etc. anyway, not too important.

src/pe/relocation.rs
```diff
@@ -131,3 +308,364 @@ impl<'a> Iterator for Relocations<'a> {
...
+pub const RELOCATION_BLOCK_SIZE: usize = 8;
+
+impl<'a> scroll::ctx::TryFromCtx<'a, scroll::Endian> for RelocationBlock<'a> {
+    type Error = crate::error::Error;
+
+    fn try_from_ctx(bytes: &'a [u8], ctx: scroll::Endian) -> Result<(Self, usize), Self::Error> {
+        let mut offset = 0;
+        let rva = bytes.gread_with::<u32>(&mut offset, ctx)?;
+        let size = bytes.gread_with::<u32>(&mut offset, ctx)?;
+        let bytes = &[];
```
m4b, line 415: why?

src/pe/relocation.rs
```diff
@@ -131,3 +308,364 @@ impl<'a> Iterator for Relocations<'a> {
...
+                .bytes
+                .gread_with::<RelocationBlock>(&mut self.offset, scroll::LE)
+            {
+                Ok(block) => {
+                    let word_size = block.size as usize - RELOCATION_BLOCK_SIZE;
+                    let word_bytes = &self.bytes[self.offset..self.offset + word_size];
+                    // "Block"s and "Word"s are entirely contiguous
+                    self.offset += word_size;
+                    Ok(RelocationBlock {
+                        bytes: word_bytes,
```
m4b, line 445: i see, this gets rewritten later. hmmm, one thing i don't like is that temporarily, if someone were to use that raw TryFromCtx impl the bytes is empty, which feels "half-constructed" to me, which is generally a pattern i like to avoid.

however, not sure there is a better solution here, other than making it optional, which also feels a little laborious and still doesn't capture semantic meaning that well

m4b (approved):
the only issue is have is the temproary empty `&[]` i mentioned in a comment. not sure if there is a better way.

only other thing is very minor nit about missing pub docs on the simpler items, but it's fine, the other documentation is excellent, thank you!

otherwise, this is great!

i'll sit on this a bit to think if there's a better design (or if you think of one) w.r.t. `&[]` otherwise i'll merge.

one option is to hide the TryFromCtx impl in some way, though I don't see that being feasible without annoying things.

src/pe/relocation.rs
kkent030315: It's been quite a while since I wrote the code and I have no idea why did I chose to do this ugly.. anyway, I fixed that. It was trivial to just merge that construction into the `TryFromCtx for RelocationBlock`.

src/pe/relocation.rs
kkent030315: Wow, didn't know that there's such a great expression. Thanks!

src/pe/relocation.rs
```diff
@@ -131,3 +308,370 @@ impl<'a> Iterator for Relocations<'a> {
...
+        let mut offset = 0;
+        let rva = bytes.gread_with::<u32>(&mut offset, ctx)?;
+        let size = bytes.gread_with::<u32>(&mut offset, ctx)?;
+        if (size as usize) < RELOCATION_BLOCK_SIZE {
+            return Err(error::Error::Malformed(format!(
+                "base reloc size {size} is less than a block {RELOCATION_BLOCK_SIZE}"
+            )));
+        }
+        let word_size = size as usize - RELOCATION_BLOCK_SIZE;
+        if offset + word_size > bytes.len() {
```
m4b, line 421: Technically you could do that same pread trick here too but not important

src/pe/relocation.rs
```diff
@@ -131,3 +308,370 @@ impl<'a> Iterator for Relocations<'a> {
...
+/// Represents a single relocation entry within a PE base relocation block.
+///
+/// Each relocation entry is stored as a 16-bit value ([`u16`]), referred to as a "relocation word."
+/// The value is divided into two components:
+///
+/// * The upper 4 bits specify the relocation type, which determines how the relocation
+///   should be applied.
+/// * The remaining 12 bits specify the offset from the block's starting RVA where the
+///   relocation should be applied.
+#[repr(C)]
```
m4b, line 475: Does this need to be repr(C)? Are you using it across FFI boundaries or constructing it and sending to C or vice versa? If not I’d opt to remove, and can add later if necessary since this is pub and it’ll be part of the pub api with that layout constraint

m4b (commented):
This is super awesome PR thank you for revising and fixing the half constructed thing. 

I’d merge except I think we don’t need repr(C) on the struct, and it’ll be a new pub item so good to get it right.

again this is awesome sorry took so long to get it merged but almost there :)

src/pe/relocation.rs
kkent030315: I don't think this needs to be a `#[repr(C)]`, thanks for the suggestion!

src/pe/relocation.rs
kkent030315: Yes I was about to do that here as well; Turns out this expression is more clear for use in here. Specifically `word_size` is a bit confusing stuff to crush into `pread<&[u8]>(word_size)`, as it requrie subtraction of previous `size_of(RelocationBlock::rva)` + `size_of(RelocationBlock::size)`. I'd think this is good for maintainability.

kkent030315: Nvm, just got the better expression without confusing around `word_size`. Thanks for suggestion!

m4b:
NB: backwards compatible

PR #446, opened 2025-03-13
elf: add Loongarch macros and name mapping

000lbh:
Loongarch is an architecture used by Loongson processor family. It has 64bit and 32bit variations, but currently there is no 32bit products in the market. This patch imports `EM_LOONGARCH` and relocation macros from `elf.h`, and implements name mapping.


000lbh:
The CI failed, but I believe that it is not my fault

m4b (approved):
LGTM, thank you for this, and thank you for your patience, these kind of tedious patches no one likes to do but i appreciate it :)

m4b:
NB: non-breaking, new pub const items in elf

PR #448, opened 2025-03-21
pe.tls.tlsdata.parse_with_opts - integer overflow + out of bound

BinFlip:
The check for the access of the bytes slice can be bypassed, by either a large 'offset' or using an integer overflow.  I attached the fuzzer test case to this post as well. 

[crash-37862deb65bdd4829ef254d5aabea5612f6453cc.zip](https://github.com/user-attachments/files/19397396/crash-37862deb65bdd4829ef254d5aabea5612f6453cc.zip)

### DIFF
--- a/src/pe/tls.rs
+++ b/src/pe/tls.rs
@@ -227,7 +227,15 @@ impl<'a> TlsData<'a> {
                         rva
                     ))
                 })?;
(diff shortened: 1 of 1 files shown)


m4b (approved):
Awesome, thanks!

m4b:
NB: non-breaking

PR #451, opened 2025-03-25
pe.header: fix parse without rich header

ideeockus:
I think we shouldn't return `Error` before we even ensure that `Rich Header` present

### DIFF
--- a/src/pe/header.rs
+++ b/src/pe/header.rs
@@ -1035,7 +1035,10 @@ impl<'a> RichHeader<'a> {
     /// - while the lower 16 bits specify the tool’s build version.
     pub fn parse(bytes: &'a [u8]) -> error::Result<Option<Self>> {
         // Parse the DOS header; some fields are required to locate the Rich header.
-        let dos_header = DosHeader::parse(bytes)?;
+        let dos_header = match DosHeader::parse(bytes) {
+            Ok(v) => v,
+            _ => return Ok(None),
+        };
         let dos_header_end_offset = PE_POINTER_OFFSET as usize;
         let pe_header_start_offset = dos_header.pe_pointer as usize;
 
--- a/src/pe/header.rs
+++ b/src/pe/header.rs
@@ -1,10 +1,11 @@
+use alloc::vec::Vec;
 use core::iter::FusedIterator;
+
+use scroll::{ctx, IOread, IOwrite, Pread, Pwrite, SizeWith};
 
 use crate::error;
 use crate::pe::{data_directories, debug, optional_header, section_table, symbol};
 use crate::strtab;
-use alloc::vec::Vec;
-use scroll::{ctx, IOread, IOwrite, Pread, Pwrite, SizeWith};
 
 /// In `winnt.h` and `pe.h`, it's `IMAGE_DOS_HEADER`. It's a DOS header present in all PE binaries.
 ///
@@ -394,6 +395,7 @@ impl DosHeader {
 pub struct DosStub<'a> {
     pub data: &'a [u8],
 }
+
 impl<'a> Default for DosStub<'a> {
     /// This is the very basic DOS program bytecode representation embedded in MSVC linker.
     ///
@@ -442,6 +444,7 @@ impl<'a> Default for DosStub<'a> {
         }
     }
 }
+
 impl<'a> ctx::TryIntoCtx<scroll::Endian> for DosStub<'a> {
     type Error = error::Error;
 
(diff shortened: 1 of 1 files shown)


ideeockus:
@kkent030315 I saw you added this part with `Rich Header`. 

check this change is ok pls

kkent030315:
@ideeockus Hi there. Why don't we add some options in `ParseOptions` (i.e., `parse_rich_header`)?

Although I'm a bit interested in what you are trying to achieve with this PR. I guess you're having some issues with the rich header parser?

https://github.com/m4b/goblin/blob/ac1fabdd2100bae949607a320fe5d8087c1e784a/src/pe/options.rs#L1-L14

ideeockus:
@kkent030315 Yes I have following issue - there is a `Header::parse_without_dos` method which I used to parse header without dos. This method assumes `DosHeader` and `DosStub` are default and bytes starts with PE header (b"PE"). 

But in `RichHeader` we try to parse `DosHeader`, which is not necessary in the case of `parse_without_dos` (there is no DosHeader). 

So this code 
```rust
fn parse_impl(
        bytes: &'a [u8],
        dos_header: DosHeader,
        dos_stub: DosStub<'a>,
    ) -> error::Result<Self> {
        let mut offset = dos_header.pe_pointer as usize;
        let rich_header = RichHeader::parse(&bytes)?;
        ...
```
leads to an error like this - "DOS header is malformed (signature 0x4550)".

kkent030315:
@ideeockus Thank you for your detailed info. I'm pretty sure that is a ghost bug (where I didn't really considered an effect on `parse_without_dos`) and should be fixed. Following is a minimum reproducible code.

```rust
#[test]
fn parse_without_dos_rich() {
    const SIZEOF_DOS_HEADER: usize = core::mem::size_of::<DosHeader>();
    let buf = &BIN[SIZEOF_DOS_HEADER..]; // Skip DOS header bytes
    let res = Header::parse_without_dos(buf);
    res.unwrap();
}
```

I strongly feel you should't use `Header::parse_without_dos` at all, this method is theoretically broken. It uses default

Adding `parse_rich_header` in `ParseOptions` is not correct approach at this point. There is more generic approach to fix the problem rather than ignoring unconditionally.

Here's an idea: `Header::parse_impl` is private and we can add `parse_rich_header: bool` flag.

```diff
impl<'a> Header<'a> {
    fn parse_impl(
        bytes: &'a [u8],
        dos_header: DosHeader,
        dos_stub: DosStub<'a>,
+       parse_rich_header: bool,
    ) -> error::Result<Self> {
        let mut offset = dos_header.pe_pointer as usize;
-       let rich_header = RichHeader::parse(&bytes)?;
+       let rich_header = if parse_rich_header {
+           RichHeader::parse(&bytes)?
+       } else {
+           None
+       };
```

Then we can tweak `parse_without_dos` accordingly.

```diff
    /// Parses PE header from the given bytes; this will fail if the DosHeader or DosStub is malformed or missing in some way
    pub fn parse(bytes: &'a [u8]) -> error::Result<Self> {
        let dos_header = DosHeader::parse(&bytes)?;
        let dos_stub = DosStub::parse(bytes, dos_header.pe_pointer)?;

-       Header::parse_impl(bytes, dos_header, dos_stub)
+       Header::parse_impl(bytes, dos_header, dos_stub, true)
    }

    /// Parses PE header from the given bytes, a default DosHeader and DosStub are generated, and any malformed header or stub is ignored
    pub fn parse_without_dos(bytes: &'a [u8]) -> error::Result<Self> {
        let dos_header = DosHeader::default();
-       Header::parse_impl(bytes, dos_header, DosStub::default())
+       Header::parse_impl(bytes, dos_header, DosStub::default(), false)
    }
```

Finally, a regression test.

```rust
const BIN: &[u8] = include_bytes!("../../tests/bins/pe/well_formed_import.exe.bin");

#[test]
fn parse_without_dos_rich() {
    // Get a PE pointer (e_lfanew)
    let dos_header = DosHeader::parse(&BIN).unwrap();
    // Skip DOS header and DOS stub
    let buf = &BIN[dos_header.pe_pointer as usize..];
    Header::parse_without_dos(buf).unwrap();
}
```

And I just caught attention on that this test was pointless and misleading. Those still pass full bytes where `Header::parse_without_dos` expects bytes **without** DOS header and stub. A fix to this test would also be great to see in this PR!

```diff
// Replace it with new parse_without_dos_rich test above
-#[test]
-fn parse_without_dos() {
-    let header = Header::parse_without_dos(&BORLAND_PE32_VALID_NO_RICH_HEADER).unwrap();
-    assert_eq!(header.dos_stub, DosStub::default());
-    assert_eq!(header.rich_header.is_none(), true);
-
-    // DOS stub is default but rich parser still works
-    let header = Header::parse_without_dos(&CORRECT_RICH_HEADER).unwrap();
-    assert_eq!(header.dos_stub, DosStub::default());
-    assert_eq!(header.rich_header.is_some(), true);
-}
```

@ideeockus Curious if you like it. :)

ideeockus:
@kkent030315 Thank you for quick answer and suggestions. Seems good to me, I've added this changes to PR

m4b:
i'm sorry due to stupid action cache CI change failure, and also to an example's dep bumping their MSRV, i've decided to bump the msrv to 1.65.0 for now to unblock CI and also have CI run; can you please `git pull --rebase` and then rebase on top of origin?

m4b (approved):
nevermind don't need rebase this looks good to me, thank you *SO MUCH* for all your patience :D

m4b:
NB: non-breaking

Issue #450, opened 2025-03-23
Detected multiple panics during fuzzing


0x64617061:
I did fuzz runs using the "parse" fuzz target with Rust overflow checks enabled and detected four different panics/issues:

1) goblin/src/archive/mod.rs:346:36 attempt to subtract with overflow
2) goblin/src/archive/mod.rs:377:20 attempt to subtract with overflow
3) goblin/src/pe/header.rs:840:22 attempt to subtract with overflow
4) goblin/src/pe/header.rs:1213:26 attempt to subtract with overflow

I have uploaded stack traces and the respective input files in a repository: https://github.com/0x64617061/goblin_repro

kkent030315:
Interesting. For those cases with PE, there are some bugs in PE header parser.

### `goblin/src/pe/header.rs:1213:26 attempt to subtract with overflow`
https://github.com/m4b/goblin/blob/ac1fabdd2100bae949607a320fe5d8087c1e784a/src/pe/header.rs#L1210-L1216

The input is never valid as a correct TE (Terse Executable) depsite this has `VZ` TE magic in first two bytes.

```
Offset(h)  00 01 02 03 04 05 06 07 08 09 0A 0B 0C 0D 0E 0F  Decoded text
00000000   56 5A 52 5A 50 00 17 00 00 00 36 00 00 00 00 10  VZPZP.....6....
00000010   86 02 0C 00 00 01 01 01 01 1B 01 00 00 FF B5     ................
00000020   00 00 00 04 34 00 00 FF B5 00 00 00 04 34 15 40  ....4.......4.@
00000030   13 41 0E 10 15 40 13 41 0E 10                    .A...@.A..
```

### `goblin/src/pe/header.rs:840:22 attempt to subtract with overflow`
https://github.com/m4b/goblin/blob/ac1fabdd2100bae949607a320fe5d8087c1e784a/src/pe/header.rs#L828-L840

This one is not particular to PE, but a bug in COFF. It's unfortunate that COFF objects do not have "magic" unline LLVM bitcodes (`BC` magic in first two bytes). COFF files are just start with COFF header (a.k.a. Optional Header) so that no one is able to verify if the file is actually a COFF object.

```
Offset(h)  00 01 02 03 04 05 06 07 08 09 0A 0B 0C 0D 0E 0F  Decoded text
00000000   FC 01 00 00 00 00 00 00 01 00 00 00 00 00 00 00  ................
00000010   0F 00 FF 80                                      ....
```

[...]

PR #454, opened 2025-03-26
archive: fix subtract with overflow in archive parser

kkent030315:
- https://github.com/m4b/goblin/issues/450

Fixed subtract with overflow in Archive parser and added regression test for it.


m4b (approved):
awesome, thank you!

m4b:
NB: non-breaking

PR #455, opened 2025-03-26
archive: fix size overflow in name index parser

kkent030315:
- https://github.com/m4b/goblin/issues/450

Fixed size overflow in name index parser and added regression test for it.


m4b (approved):
LGTM

m4b:
darn this has conflicts unfortunately, probably from the previous archive fuzz patch?

kkent030315:
@m4b done as well!

m4b (approved):
LGTM

PR #456, opened 2025-03-31
pe.header: fix parse header with no dos stub

ideeockus:
Reason: dos stub is optional

Changes:
- fixed `DosStub::parse`
- added test with no-dos-stub header


m4b (approved):
LGTM, thank you so much for your patience! and also adding the test!

m4b:
ref #467

PR #457, opened 2025-03-31
pe.options: added parse_tls_data option

ideeockus:
Reason:
- sometimes I want to skip tls data parsing

Changes: 
- added `ParseOptions.parse_tls_data`

### DIFF
--- a/src/pe/options.rs
+++ b/src/pe/options.rs
@@ -39,7 +39,7 @@ impl ParseOptions {
         Self {
             resolve_rva: false,
             parse_attribute_certificates: false,
-            parse_tls_data: false,
+            parse_tls_data: true,
             parse_mode: ParseMode::Strict,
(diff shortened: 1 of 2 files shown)


src/pe/options.rs
```diff
@@ -37,6 +39,7 @@ impl ParseOptions {
         Self {
             resolve_rva: false,
             parse_attribute_certificates: false,
+            parse_tls_data: false,
```
m4b, line 42: just curious, what is the rationale for skipping this for te?

m4b:
Needs rebase for CI to run

src/pe/options.rs
ideeockus: you are right, it might be better to leave it enabled

m4b:
NB: non-breaking

PR #466, opened 2025-04-24
Handle Malformed/Obfuscated strtab Without UTF-8 Validation Error

chf0x:
Hi everyone,

I encountered an issue while parsing malformed/obfuscated ELF files. Goblin tends to fail with a "bad input invalid UTF-8" error when it encounters non-valid UTF-8 strings in the `.strtab` section.

I believe this behavior isn't ideal, as we should still be able to parse these ELF files correctly even if the strings within `.strtab` aren't valid UTF-8. I think, there is no explicit requirement for the strings to be valid UTF-8 in strtab?

Here is an example of the kind of malformed input causing the issue:

```
    64: 0000000000000000     0 OBJECT  GLOBAL DEFAULT  UND __sF
    65: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND calloc
    66: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND B|n5:Bn72FI?:#n[...]
    67: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND B|n5:Bn72FI?:#n[...]
    68: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND B|n5:Bn72FI?:#n[...]
    69: 000000000008cbfc  1950 FUNC    GLOBAL DEFAULT    1 p:z^I�^V^�9�^X3[...]
    70: 00000000000c41ea   434 FUNC    GLOBAL DEFAULT    1 ^�7,�j5!�j8#�P71[...]
    71: 00000000000c4118   435 FUNC    GLOBAL DEFAULT    1 ^\^��T^I)�X^I^T�[...]
    72: 00000000000c4137   432 FUNC    GLOBAL DEFAULT    1 �[@7^I"Y;�)S"O/E[...]
    73: 00000000000c4127   435 FUNC    GLOBAL DEFAULT    1 |7��i51�i8o�S7��[...]
```

Please find the attached binary as an example. As a temporary workaround, I simply replace invalid strings with an empty string (`""`) when parsing fails, but I don't think this is the most appropriate solution.

I'd like to open a discussion on how we should handle these cases. Should we consider skipping over invalid UTF-8 strings instead of failing, or is there another approach we’d prefer to implement?

Thanks a lot!

### DIFF
--- a/src/strtab.rs
+++ b/src/strtab.rs
@@ -80,7 +80,13 @@ impl<'a> Strtab<'a> {
         let mut result = Self::from_slice_unparsed(bytes, offset, len, delim);
         let mut i = 0;
         while i < result.bytes.len() {
-            let string = get_str(i, result.bytes, result.delim)?;
+            let string = match get_str(i, result.bytes, result.delim) {
+                Ok(s) => s,
+                Err(_) => {
+                    // On error set to empty string
+                    ""
+                }
+            };
             result.strings.push((i, string));
             i += string.len() + 1;
         }


chf0x:
[file.zip](https://github.com/user-attachments/files/19893520/file.zip)

src/strtab.rs
```diff
@@ -80,7 +80,13 @@ impl<'a> Strtab<'a> {
         let mut result = Self::from_slice_unparsed(bytes, offset, len, delim);
         let mut i = 0;
         while i < result.bytes.len() {
-            let string = get_str(i, result.bytes, result.delim)?;
+            let string = match get_str(i, result.bytes, result.delim) {
+                Ok(s) => s,
+                Err(_) => {
+                    // On error set to empty string
+                    ""
```
m4b, line 87: while i think we should come up with a solution here, returning an empty string probably isn't ideal and wouldn't be useful for any kind of analysis if there is > 1 non-utf8 string in the binary

m4b (changes requested):
not sure the best solution to this other than having the strtab return something like `&[u8]` or perhaps a enum with regular ergonomic &str or &[u8] if fails to parse :shrug:  either way isn't super great for most users though

m4b:
I've been thinking about this a lot and i think this is going to be very tricky to handle right. this came up once in the past in another context, suggestion was to return raw bytes, but  rust `&str` are guaranteed to be utf-8. if the obfuscator uses non-utf8 sequences in the symbol table, then we need to either return raw bytes, or something like a:

```rust
enum GoblinStr<'a> {
  Raw(&'[u8])
  Str(&'a str)
}
```

e.g., if we fail to validate a string as utf8 when it comes out of the string table, then it gets returned as the raw bytes.

however, this would massively pessimize and make highly unergonomic string handling for the entire library due to what is arguably a very niche case when loading and parsing binaries (albeit important).

so this is a tough one i think.

chf0x:
I agree. So probably for now, we could at least add a permissive mode for strings to skip any strings we can't parse, instead of failing entirely?

chf0x:
Another example, this time not related to strtab, but to section names. I’ll add it to this issue, as I think we can treat it as part of the broader UTF-8 meta issue. Let me know if you’d prefer to track it separately instead.

[example.zip](https://github.com/user-attachments/files/20775533/example.zip)

image

PR #470, opened 2025-06-06
PE: Add delay import parser

kkent030315:
Added delay-load import directory parser for PE32/64.

Using `Vec` as well as existing normal import parser. The design decision behind that `Vec` is that each import entry requires RVA resolution. I'd say it is not a good design, as both normal and delay import parser runs fully upon `goblin::pe::PE::parse` that can raise malformed errors -- even if the consumer doesn't need that import info.

I guess almost entire parsing logic can be moved to `TryFromCtx` if we could build some specific scroll ctx with a hell of lifetimes, so that `utils::find_offset` became callable inside the `TryFromCtx`. That's too much hustle but it should theoretically be possible.

```rust
/// A binary parsing context for PE parser, including the container size, underlying byte endianness and params for resolving RVAs
#[derive(Debug, Copy, Clone, PartialEq)]
pub struct PeCtx {
  pub container: Container,
  pub le: scroll::Endian,

  pub sections: &[section_table::SectionTable],
  pub file_alignment: u32,
  pub opts: &options::ParseOptions,
};
```

```rust
impl<'a> Iterator for DelayImportEntryIterator<'a>
impl<'a> Iterator for DelayImportDescriptorIterator<'a>
```

Ref:
- https://www.hexacorn.com/blog/2019/06/03/playing-with-delay-loaded-dlls/

### DIFF
--- a/src/pe/delay_import.rs
+++ b/src/pe/delay_import.rs
@@ -1,0 +1,293 @@
+use alloc::vec::Vec;
+use scroll::ctx;
+use scroll::{Pread, Pwrite, SizeWith};
+
+use crate::container::{Container, Ctx};
+use crate::error;
+use crate::pe::data_directories;
+use crate::pe::import::{IMPORT_BY_ORDINAL_32, IMPORT_BY_ORDINAL_64};
+use crate::pe::options;
+use crate::pe::section_table;
+use crate::pe::utils;
+
+/// Represents a single entry in the delay import table of a PE (Portable Executable) file.
+#[derive(Debug)]
+pub struct DelayImportEntry<'a> {
+    /// The associated delay import descriptor.
+    pub descriptor: DelayImportDescriptor,
+    /// Name offset of this delay import entry.
+    pub offset: u32,
+    /// RVA of this delay import entry.
+    pub rva: u32,
+    /// Hint value for symbol lookup.
+    pub hint: u16,
+    /// Name of the DLL this function is imported from.
+    pub dll: &'a str,
+    /// Name of the imported function, if available.
+    ///
+    /// This is `None` if the import is by ordinal.
+    pub name: Option<&'a str>,
+    /// Ordinal number if the function is imported by ordinal.
+    ///
+    /// This is `0` if `name` is `None` and vice versa.
+    pub ordinal: u16,
+}
+
+/// Internal helper struct to simplify bitflag operations.
+#[derive(Debug, SizeWith)]
+struct DelayImportThunk(pub u64);
+
+impl<'a> ctx::TryFromCtx<'a, Ctx> for DelayImportThunk {
+    type Error = crate::error::Error;
+    fn try_from_ctx(bytes: &'a [u8], ctx: Ctx) -> error::Result<(Self, usize)> {
+        let mut offset = 0;
+
+        let value = if ctx.is_big() {
+            bytes.gread_with::<u64>(&mut offset, scroll::LE)?
+        } else {
+            let v = bytes.gread_with::<u32>(&mut offset, scroll::LE)?;
+            v as u64
+        };
+
+        Ok((Self(value), offset))
+    }
+}
+
+impl DelayImportThunk {
+    /// Whether the entry is imported by ordinal (64-bit)
+    pub fn is_ordinal64(&self) -> bool {
+        self.0 & IMPORT_BY_ORDINAL_64 != 0
+    }
+
+    /// Whether the entry is imported by ordinal (32-bit)
+    pub fn is_ordinal32(&self) -> bool {
+        (self.0 as u32) & IMPORT_BY_ORDINAL_32 != 0
+    }
+
+    /// Returns the ordinal value
+    ///
+    /// Use this function if either `is_ordinal64` or `is_ordinal32` returns `true`.
+    pub fn ordinal(&self) -> u16 {
+        (self.0 as u16) & 0xFFFFu16
+    }
+
+    /// Returns the name rva value
+    ///
+    /// Use this function if either `is_ordinal64` or `is_ordinal32` returns `false`.
+    pub fn name_rva(&self) -> u32 {
+        self.0 as u32
+    }
+
+    /// Checks if the underlying value is zero.
+    pub fn is_null(&self) -> bool {
+        self.0 == 0
+    }
+}
+
+/// Represents a delay-load import descriptor in a PE (Portable Executable) file.
+///
+/// This structure corresponds to the `IMAGE_DELAYLOAD_DESCRIPTOR` defined in the PE/COFF specification.
+/// Each instance describes a single delay-loaded DLL and its associated import data.
+#[derive(Debug, Copy, Clone, Pread, Pwrite, SizeWith)]
+#[repr(C)]
+pub struct DelayImportDescriptor {
+    /// Attributes describing the delay load characteristics (e.g. pointer size, flags).
+    pub attributes: u32,
+    /// RVA of a null-terminated string containing the DLL name.
+    pub name_rva: u32,
+    /// RVA to a handle location that will store the DLL module handle after loading.
+    pub handle_rva: u32,
+    /// RVA of the delay-load Import Address Table (IAT).
+    pub address_table_rva: u32,
+    /// RVA of the delay-load Import Name Table (INT).
+    pub name_table_rva: u32,
+    /// RVA of the bound import address table (optional; used if binding is supported).
+    pub bound_table_rva: u32,
+    /// RVA of the unload import address table (optional; used during DLL unloading).
+    pub unload_table_rva: u32,
+    /// Timestamp of the bound DLL; zero if not bound.
+    pub time_stamp: u32,
+}
+
+impl<'a> DelayImportDescriptor {
+    /// Returns whether the RVA is based (RvaBased field).
+    pub fn rva_based(&self) -> bool {
+        self.attributes & 0x1 != 0
+    }
+
+    /// Returns the ReservedAttributes field (31 bits).
+    pub fn reserved_attributes(&self) -> u32 {
+        self.attributes >> 1
+    }
+
+    /// Whether the entire fields are set to zero
+    pub fn is_null(&self) -> bool {
+        self.attributes == 0
+            && self.name_rva == 0
+            && self.handle_rva == 0
+            && self.address_table_rva == 0
+            && self.name_table_rva == 0
+            && self.bound_table_rva == 0
+            && self.unload_table_rva == 0
+            && self.time_stamp == 0
+    }
+
+    /// Whether the entry is _possibly_ valid.
+    ///
+    /// Both [`Self::name_rva`] and [`Self::address_table_rva`] must be non-zero
+    pub fn is_possibly_valid(&self) -> bool {
+        self.name_rva != 0 && self.address_table_rva != 0
+    }
+}
+
+/// Represents a PE delay import directory data.
+#[derive(Debug)]
+pub struct DelayImportData<'a> {
+    /// Delay imported entries.
+    pub imports: Vec<DelayImportEntry<'a>>,
+}
+
+impl<'a> DelayImportData<'a> {
+    pub fn parse(
+        bytes: &'a [u8],
+        dd: data_directories::DataDirectory,
+        sections: &[section_table::SectionTable],
+        file_alignment: u32,
+        is_64: bool,
+    ) -> error::Result<DelayImportData<'a>> {
+        Self::parse_with_opts(
+            bytes,
+            dd,
+            sections,
+            file_alignment,
+            &options::ParseOptions::default(),
+            is_64,
+        )
+    }
+
+    pub fn parse_with_opts(
+        bytes: &'a [u8],
+        dd: data_directories::DataDirectory,
+        sections: &[section_table::SectionTable],
+        file_alignment: u32,
+        opts: &options::ParseOptions,
+        is_64: bool,
+    ) -> error::Result<DelayImportData<'a>> {
+        let offset =
+            utils::find_offset(dd.virtual_address as usize, sections, file_alignment, opts)
+                .ok_or_else(|| {
+                    error::Error::Malformed(format!(
+                        "cannot map delay import table {:#x}",
+                        dd.virtual_address
+                    ))
+                })?;
+        if offset + dd.size as usize > bytes.len() {
+            return Err(error::Error::Malformed(format!(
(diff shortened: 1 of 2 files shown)


kkent030315:
I made an brief example implementation of my idea to have `PeCtx` in [2fd0a44](https://github.com/m4b/goblin/pull/470/commits/2fd0a447c19e6473eebcb58a7769794e4985f238). Consider the code an experimental code that aren't ready to ship.

I've added custom `Ctx`:

```rust
/// A binary parsing context for PE parser
#[derive(Debug, Copy, Clone, PartialEq)]
pub struct PeCtx<'a> {
    pub container: Container,
    pub le: scroll::Endian,
    pub sections: &'a [section_table::SectionTable],
    pub file_alignment: u32,
    pub opts: options::ParseOptions,
    pub bytes: &'a [u8], // full binary view
}
```

The issue is that we cannot acrually borrow the `&'a sections` in `DelayImportData<'a>`. Nor we cannot make `sections: Vec<section_table::SectionTable>` because `Ctx` must implement `Copy` that `Vec` doesn't.

The solution to that problem is to require consumers provide `sections` when they needed to get delay import info.

```rust
impl<'a> DelayImportDll<'a> {
    pub fn functions(
        &self,
        sections: &'a [section_table::SectionTable],
    ) -> DelayImportFunctionIterator<'a> {
        // Replace sections with an actual sections ref
        let pectx = PeCtx::new(
            self.ctx.container,
            self.ctx.le,
            &sections,
            self.ctx.file_alignment,
            self.ctx.opts,
            self.ctx.bytes,
        );
        DelayImportFunctionIterator {
            ctx: pectx,
            bytes: self.bytes,
            offset: 0,
            descriptor: self.descriptor,
        }
    }
}

impl<'a> DelayImportData<'a> {
    pub fn dlls(&self, sections: &'a [section_table::SectionTable]) -> DelayImportDllIterator<'a> {
        // Replace sections with an actual sections ref
        let pectx = PeCtx::new(
            self.ctx.container,
            self.ctx.le,
            &sections,
            self.ctx.file_alignment,
            self.ctx.opts,
            self.ctx.bytes,
        );
        DelayImportDllIterator {
            ctx: pectx,
            bytes: self.bytes,
            offset: 0,
        }
    }
}
```

I think, the best is to have an offset of section table and parsing it when needed. So that we eliminate the odd `sections` parameter in public APIs.

```diff
/// A binary parsing context for PE parser
#[derive(Debug, Copy, Clone, PartialEq)]
pub struct PeCtx<'a> {
    pub container: Container,
    pub le: scroll::Endian,
-   pub sections: &'a [section_table::SectionTable],
+   // let sections = ctx.bytes.pread_with(ctx.sections_offset, scroll::LE)?;
+   pub sections_offset: usize, 
    pub file_alignment: u32,
    pub opts: options::ParseOptions,
    pub bytes: &'a [u8], // full binary view
}
```

Otherwise it looks very clean design for me. It makes parsing of delay imports deferred to when it's needed.

```rust
let res = goblin::pe::PE::parse(&buffer)?;
let di = res.delay_import_data.unwrap();
let sections = &res.sections;
let delayimports = di
  .dlls(&sections)
  .filter_map(Result::ok)
  .flat_map(|x| x.functions(&sections))
  .filter_map(Result::ok)
  .collect::<Vec<_>>();
```

m4b:
Have not fully read over the PeCtx idea, looks interesting though; few initial notes:
1. why is
> The issue is that we cannot acrually borrow the &'a sections in DelayImportData<'a>

the case? is it some kind of cycle issue or?
2. for
```
        let pectx = PeCtx::new(
            self.ctx.container,
            self.ctx.le,
            &sections,
            self.ctx.file_alignment,
            self.ctx.opts,
            self.ctx.bytes,
        );
```
you can rewrite this as:
```rust
let pectx = PeCtx { sections: &sections, ..self.ctx) }
```
as long as all fields of PeCtx are accessible in the use context (e.g., public, or within the file if private, module if pub(crate)). In other languages this is called a "functional record update" or something to that effect, ocaml has a similar syntactic mechanism, for example.

m4b:
sorry this didn't make it in for 0.10.1, but it does need a rebase; i did have some concerns about this, as it looked a little complicated, but i'll just have to re-review again when it's ready. Also I can't remember if it had breaking changes, so if it did it wouldn't have made it into 0.10.1; with this, I think i've caught up to your absolutely massive amount of PRs :D

kkent030315:
@m4b This PR needs a direction on which way this PR should go with: A) traditional `Vec` or B) new `PeCtx<'a>` and "deffered" iterators. I personally prefer challenging on `PeCtx<'a>` that might also be applied to import/export parser rework in near future. WDYT?

kkent030315:
Once #239 gets merged we could add `section_table: &'a [u8]` in `PeCtx` that will get parsed inside `FromCtx` and such.

```rs
/// A binary parsing context for PE parser, including the container size, underlying byte endianness and params for resolving RVAs
#[derive(Debug, Copy, Clone, PartialEq)]
pub struct PeCtx<'a> {
  pub container: Container,
  pub le: scroll::Endian,
  pub section_table: &'a [u8],
  pub file_alignment: u32,
  pub opts: &options::ParseOptions,
  pub bytes: &'a [u8], // full binary view
}
```

```rs
impl<'a> ctx::TryFromCtx<'a, (DelayImportDescriptor, PeCtx<'a>)> for DelayImportFunction<'a> {
    type Error = crate::error::Error;
    fn try_from_ctx(
        bytes: &'a [u8],
        ctx: (DelayImportDescriptor, PeCtx<'a>),
    ) -> error::Result<(Self, usize)> {
		let sections_it = SectionTableIterator { // Fused, ExactSize
			data: ctx.section_table, // &'a [u8]
		};
		let sections: Vec<SectionHeader> = sections_it.collect<Result<Vec<_>>>()?;
		// Or even `ctx.section_table.pread::<&[SectionHeader]>(0)` maybe?
        let name = if is_ordinal {
            None
        } else {
            let dll_name_rva = descriptor.name_rva;
            let dll_name_offset = utils::find_offset(
                dll_name_rva as usize,
                sections,
                ctx.file_alignment,
                &ctx.opts,
            )
            .ok_or_else(|| {
                    "cannot map delay import dll name rva {dll_name_rva:#x}"
                ))
            })?;
            let dll_name = ctx.bytes.pread::<&'a str>(dll_name_offset)?;
            Some(dll_name)
        };
		// ...
    }
}
```

src/pe/delay_import.rs
```diff
@@ -0,0 +1,529 @@
...
+        self.container.is_big()
+    }
+
+    /// Whether this binary container context is little endian or not
+    pub fn is_little_endian(self) -> bool {
+        self.le.is_little()
+    }
+
+    /// Return a dubious pointer/address byte size for the container
+    pub fn size(self) -> usize {
```
m4b: this and all above should be `pub(crate)` for now

src/pe/delay_import.rs
```diff
@@ -0,0 +1,529 @@
...
+
+impl<'a> ctx::TryFromCtx<'a, (DelayImportDescriptor, PeCtx<'a>)> for DelayImportFunction<'a> {
+    type Error = crate::error::Error;
+    fn try_from_ctx(
+        bytes: &'a [u8],
+        ctx: (DelayImportDescriptor, PeCtx<'a>),
+    ) -> error::Result<(Self, usize)> {
+        let mut offset = 0;
+
+        let (descriptor, ctx) = ctx;
```
m4b: you can actually do this in the function args, e.g.:
```rust
    fn try_from_ctx(
        bytes: &'a [u8],
        (descriptor, ctx): (DelayImportDescriptor, PeCtx<'a>),
```

src/pe/delay_import.rs
```diff
@@ -0,0 +1,529 @@
...
+/// A binary parsing context for PE parser
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub struct PeCtx<'a> {
+    pub container: Container,
+    pub le: scroll::Endian,
+    pub sections: &'a [section_table::SectionTable],
+    pub file_alignment: u32,
+    pub opts: options::ParseOptions,
+    pub bytes: &'a [u8], // full binary view
+}
```
m4b: if this is meant to be a generic Ctx for parsing PE binaries, then it shouldn't be in this module, for example. a better place is `mod` or even better a `ctx.rs` file in the `src/pe` directory.

src/pe/delay_import.rs
```diff
@@ -0,0 +1,529 @@
...
+            .bytes
+            .gread_with::<DelayImportThunk>(&mut self.offset, ctx)
+            .ok()?;
+        if thunk.is_null() {
+            self.bytes = &[];
+            return None;
+        }
+
+        let ctx = (self.descriptor, self.ctx);
+        let func = self.bytes.pread_with::<DelayImportFunction>(0, ctx).ok()?;
```
m4b: all control flow appears to do `ok()?` which converts Error -> Option, so the iteration item can just be `type Item = DelayImportFunction<'a>` if that's your intention (to drop all error messages during iteration, and stop on first error)

src/pe/delay_import.rs
```diff
@@ -0,0 +1,529 @@
...
+    type Error = crate::error::Error;
+    fn try_from_ctx(bytes: &'a [u8], ctx: Ctx) -> error::Result<(Self, usize)> {
+        let mut offset = 0;
+
+        let value = if ctx.is_big() {
+            bytes.gread_with::<u64>(&mut offset, scroll::LE)?
+        } else {
+            let v = bytes.gread_with::<u32>(&mut offset, scroll::LE)?;
+            v as u64
+        };
```
m4b: not your concern, but i've seen this pattern enough I feel like this would be a good use for an annotation in scroll, e.g.:

```rust
#[derive(Debug, SizeWith, Pread)]
struct DelayImportThunk(#[scroll(upcast)] pub u64);
```
or perhaps

```rust
#[scroll(from("function"))]
```

src/pe/delay_import.rs
```diff
@@ -0,0 +1,529 @@
...
+                })?;
+        if offset + dd.size as usize > bytes.len() {
+            return Err(error::Error::Malformed(format!(
+                "delay import offset {:#x} and size {:#x} exceeds the bounds of the bytes size {:#x}",
+                offset,
+                dd.size,
+                bytes.len()
+            )));
+        }
+        let directory_bytes = &bytes[offset..offset + dd.size as usize];
```
m4b: rewrite this to `pread::<&[u8]>(offset, dd.size as usize).map_err` and can remove check above and replace error with the `map_err`

src/pe/delay_import.rs
```diff
@@ -0,0 +1,529 @@
...
+    pub fn is_null(&self) -> bool {
+        self.0 == 0
+    }
+}
+
+/// Represents a delay-load import descriptor in a PE (Portable Executable) file.
+///
+/// This structure corresponds to the `IMAGE_DELAYLOAD_DESCRIPTOR` defined in the PE/COFF specification.
+/// Each instance describes a single delay-loaded DLL and its associated import data.
+#[derive(Debug, Copy, Clone, Pread, Pwrite, SizeWith)]
```
m4b: just curious why you're deriving SizeWith here and above; this is for the IOwrite traits, which need a size given a ctx so that it can write into a stack buffer.

src/pe/delay_import.rs
```diff
@@ -0,0 +1,529 @@
...
+                if thunk.is_null() {
+                    break 'inner;
+                }
+
+                let is_ordinal = if is_64 {
+                    thunk.is_ordinal64()
+                } else {
+                    thunk.is_ordinal32()
+                };
+                let ordinal = if is_ordinal { thunk.ordinal() } else { 0 };
```
m4b: coalesce this logic with is_ordinal check below

src/pe/delay_import.rs
```diff
@@ -0,0 +1,529 @@
...
+                    let func_name_rva = thunk.name_rva();
+                    let func_name_offset =
+                        utils::find_offset(func_name_rva as usize, sections, file_alignment, opts)
+                            .ok_or_else(|| {
+                                error::Error::Malformed(format!(
+                                    "cannot map delay import byname rva {func_name_rva:#x}"
+                                ))
+                            })?;
+                    let hint = bytes.pread_with::<u16>(func_name_offset, scroll::LE)?;
+                    let name = bytes.pread::<&'a str>(func_name_offset + 2)?; // + 2 = sizeof(hint)
```
m4b: gread probably preferred here.

src/pe/delay_import.rs
```diff
@@ -0,0 +1,529 @@
...
+                if is_ordinal {
+                    imports.push(DelayImportEntry {
+                        descriptor,
+                        offset: current_name_offset as u32,
+                        rva,
+                        hint: 0,
+                        dll: dll_name,
+                        name: None,
+                        ordinal,
+                    });
```
m4b: i prefer the fields to be constructed/returned from ordinal, which is just:

```
ordinal, name, and hint
```

you can then have a single push, e.g.:

```rust
let (ordinal, name, hint) = if is_ordiinal {
  (thunk.ordinal(), None, 0)
} else {
 (0, // logic in other branch)
};
imports.push(DelayImportEntry {
     descriptor,
     offset: current_name_offset as u32,
     rva,
     hint,
     dll: dll_name,
     name,
     ordinal,,
 });
```

src/pe/delay_import.rs
```diff
@@ -0,0 +1,529 @@
...
+            return None;
+        }
+
+        let ctx = Ctx::new(self.ctx.container, scroll::LE);
+        let thunk = self
+            .bytes
+            .gread_with::<DelayImportThunk>(&mut self.offset, ctx)
+            .ok()?;
+        if thunk.is_null() {
+            self.bytes = &[];
```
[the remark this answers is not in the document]
kkent030315: Linkers insert a null ("thunk") entries after all other to indicate the list is terminated. This is also strictly required.

src/pe/delay_import.rs
kkent030315: This is just to follow as all others, I think it's a traditional way in goblin to make them pwrite/sizewith? otherwise at least they aren't needed for the parser to function.

src/pe/options.rs
```diff
@@ -14,7 +14,7 @@ pub struct ParseOptions {
     pub parse_mode: ParseMode,
 }
 
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Copy, Clone, PartialEq)]
```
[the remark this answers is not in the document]
kkent030315: I think this was used in tests I wrote long ago. I can't recall it. It doesn't needed.

PR #472, opened 2025-06-13
LoadConfig cannot read cb size fix

chf0x:
Hi guys, I think we have a typo here. I attached a binary as an example where it fails
[example.zip](https://github.com/user-attachments/files/20728727/example.zip)

### DIFF
--- a/src/pe/load_config.rs
+++ b/src/pe/load_config.rs
@@ -337,7 +337,7 @@ impl LoadConfigData {
                     ))
                 })?;
         let bytes = bytes[offset..]
-            .pread::<&[u8]>(dd.size as usize)
+            .pread_with::<&[u8]>(0, dd.size as usize)
             .map_err(|_| {
                 error::Error::Malformed(format!(
                     "load config offset {:#x} and size {:#x} exceeds the bounds of the bytes size {:#x}",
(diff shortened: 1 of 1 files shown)


m4b:
would you mind fixing the warnings in load_config.rs while you're there? i think i need to turn on warnings-as-failures in CI job

```
warning: variable does not need to be mutable
   --> src/pe/load_config.rs:217:13
    |
217 |         let mut read_arch_dependent_u64 = |offset: &mut usize| {
    |             ----^^^^^^^^^^^^^^^^^^^^^^^
    |             |
    |             help: remove this `mut`
    |
    = note: `#[warn(unused_mut)]` on by default

warning: unused variable: `size`
   --> src/pe/load_config.rs:349:13
    |
349 |         let size = bytes.pread::<u32>(0).map_err(|_| {
    |             ^^^^ help: if this is intentional, prefix it with an underscore: `_size`
    |
    = note: `#[warn(unused_variables)]` on by default
```

also @kkent030315 this last warning is a little odd, why is the size unused here? seems to be parsed and ignored, could be another latent bug?

kkent030315:
@m4b I might forgot to remove the unused size var, but that really doesn't matter in the case. The problem is in `let bytes = bytes[offset..].pread::<&[u8]>` where it returns zero sized slice even if the size and bytes are correct. I guess something is wrong in scroll implementation?

chf0x:
@kkent030315 I may be wrong, but in this case, are you suggesting reading from the offset dd.size, rather than reading dd.size bytes?
```
pread::<&[u8]>(dd.size as usize)
```

m4b (approved):
Thank you!

PR #473, opened 2025-06-14
PE: fix TLS raw data parser

kkent030315:
In favor to #458, I parted the change specifics to the possible TLS parser panic to this PR.

`TlsData::raw_data` can be `None` so instead of raising a malformed error if the address does not exists or does not lies in the binary, I made it simply `None`.


m4b (approved):
thank you! would be nice to have a test showing regression won't return, but maybe follow up

m4b (approved):
thank you!

m4b:
NB: nonbreaking

PR #475, opened 2025-06-15
elf.note: added type_to_str support for coredump constants

gunbux:
I was trying to parse some coredumps with goblin, and got really confused when calling `type_to_str` because it didn't give me the coredump note constants that were in goblin. I noticed that although goblin had the common coredump constants, we aren't actually using it anywhere, so I thought it'd be nice to at least parse these constants properly in `type_to_str`

### DIFF
--- a/src/elf/note.rs
+++ b/src/elf/note.rs
@@ -188,6 +188,15 @@ if_alloc! {
 
     impl<'a> Note<'a> {
         pub fn type_to_str(&self) -> &'static str {
+            // If the Note is from a coredump, match to different tags
+            if self.name == "CORE" {
+                return match self.n_type {
+                    NT_PRPSINFO => "NT_PRPSINFO",
+                    NT_PRSTATUS => "NT_PRSTATUS",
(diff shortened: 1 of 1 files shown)


src/elf/note.rs
```diff
@@ -188,6 +188,15 @@ if_alloc! {
 
     impl<'a> Note<'a> {
         pub fn type_to_str(&self) -> &'static str {
+            // If the Note is from a coredump, match to different tags
+            if self.name == "CORE" {
```
m4b, line 192: i haven't looked at this code in a while, but is it reliable to depend on self.name == "CORE" to parse from a coredump? E.g., are there cases when the file is a coredump, but self.name is not "CORE"?

gunbux: Afaik, looking at the linux source code, this should always be the case for the constants we care about:

https://elixir.bootlin.com/linux/v6.15.1/source/include/uapi/linux/elf.h#L399

m4b: ack

m4b:
NB: non-breaking

PR #476, opened 2025-06-17
Fix data directories and praise clippy

DreydenGys:
The first commit fixes the bug that I previously mentioned here: https://github.com/m4b/goblin/pull/471#issuecomment-2955345191

The second and third ones don't change any behavior and are a major refactoring that aims to reduce the number of clippy/build warnings. 
Most of them are now resolved, the only ones remaining are the ones related to `# Safety` sections of unsafe functions. As I didn't write these functions I don't feel safe to writing these documentations. 

Now that most of the issues thrown by clippy have been resolved, it would be nice to add a new job to run clippy in the CI.


m4b:
has conflicts now (not unexpected);  i suggest a more piecemeal set of patches because this is a lot to review, though largely mechanical. mod by mod might be a good idea? or start with elf and macho, since there isn't much movement in those parts of the codebase. this is probably a little disruptive to the open PE patches so i will prioritize those

m4b:
Also if this is fixing a bug please put that in a separate PR, outside of any formatting/linting auto-fixes, as that is a lot of diff noise unrelated to a bugfix.

m4b:
closing for now, please feel free to submit PR with bug fix separate and lint fixes, as suggested above, thank you for your time!

PR #477, opened 2025-06-19
PE: Add parser for `C_SCOPE_TABLE`

kkent030315:
Added parser for `C_SCOPE_TABLE` in `UnwindHandler::ExceptionHandler`.

This patch is backward compatible.

Reference:
- https://blog.talosintelligence.com/exceptional-behavior-windows-81-x64-seh/
  > If the filter returns the code EXCEPTION_EXECUTE_HANDLER, the C specific handler prepares all the data needed to execute the relative exception handler and finally calls the routine [RtlUnwindEx](http://msdn.microsoft.com/en-us/library/windows/desktop/ms680615%28v=vs.85%29.aspx).

### DIFF
--- a/src/pe/exception.rs
+++ b/src/pe/exception.rs
@@ -106,14 +106,17 @@ pub struct ScopeTableEntry {
     ///
     /// This marks the beginning of a `try` block.
     pub begin: u32,
+
     /// The ending RVA (exclusive) of the protected code region.
     ///
     /// This marks the end of the `try` block.
     pub end: u32,
+
     /// The RVA of the exception handler function.
     ///
     /// e.g., be invoked when an exception occurs in the associated code range.
     pub handler: u32,
+
     /// The RVA of the continuation target after the handler is executed.
     ///
     /// This is used for control transfer (e.g., continuation blocks, to resume execution after `finally`).
--- a/src/pe/exception.rs
+++ b/src/pe/exception.rs
@@ -696,9 +696,7 @@ impl<'a> UnwindInfo<'a> {
         let mut offset = 0;
         let num_entries = data.gread_with::<u32>(&mut offset, scroll::LE).ok()?;
         let table_size = num_entries * core::mem::size_of::<ScopeTableEntry>() as u32;
-        let data = data[..]
-            .pread_with::<&[u8]>(offset, table_size as usize)
-            .ok()?;
+        let data = data.pread_with::<&[u8]>(offset, table_size as usize).ok()?;
         Some(ScopeTableIterator { data })
     }
 }
--- a/src/pe/exception.rs
+++ b/src/pe/exception.rs
@@ -138,7 +138,7 @@ impl Iterator for ScopeTableIterator<'_> {
             return None;
         }
 
-        // It is guarabteed that .expect here is really a unreachable.
+        // It is guaranteed that .expect here is really a unreachable.
         // See: that we do `num_entries * core::mem::size_of::<ScopeTableEntry>() as u32;`
         Some(
             self.data
--- a/src/pe/exception.rs
+++ b/src/pe/exception.rs
@@ -127,26 +127,22 @@ pub struct ScopeTableEntry {
 #[derive(Debug)]
 pub struct ScopeTableIterator<'a> {
     data: &'a [u8],
+    offset: usize,
 }
 
 impl Iterator for ScopeTableIterator<'_> {
     type Item = error::Result<ScopeTableEntry>;
 
     fn next(&mut self) -> Option<Self::Item> {
-        if self.data.is_empty() {
+        if self.offset >= self.data.len() {
             return None;
         }
 
-        Some(match self.data.pread_with(0, scroll::LE) {
-            Ok(func) => {
-                self.data = &self.data[core::mem::size_of::<ScopeTableEntry>()..];
-                Ok(func)
-            }
-            Err(error) => {
-                self.data = &[];
-                Err(error.into())
-            }
-        })
+        Some(
+            self.data
+                .gread_with(&mut self.offset, scroll::LE)
+                .map_err(Into::into),
+        )
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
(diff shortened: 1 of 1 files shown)


src/pe/exception.rs
```diff
@@ -624,6 +685,22 @@ impl<'a> UnwindInfo<'a> {
...
+    pub fn c_scope_table_entries(&self) -> Option<ScopeTableIterator<'a>> {
+        let UnwindHandler::ExceptionHandler(_, data) = self.handler? else {
+            return None;
+        };
+        let mut offset = 0;
+        let num_entries = data.gread_with::<u32>(&mut offset, scroll::LE).ok()?;
+        let table_size = num_entries * core::mem::size_of::<ScopeTableEntry>() as u32;
+        let data = data[..]
+            .pread_with::<&[u8]>(offset, table_size as usize)
+            .ok()?;
```
DreydenGys, line 701:
```suggestion
        let data = data
            .pread_with::<&[u8]>(offset, table_size as usize)
            .ok()?;
```

Why are you using `[...]`?

src/pe/exception.rs
```diff
@@ -94,6 +94,67 @@ const RUNTIME_FUNCTION_SIZE: usize = 12;
...
+    pub end: u32,
+    /// The RVA of the exception handler function.
+    ///
+    /// e.g., be invoked when an exception occurs in the associated code range.
+    pub handler: u32,
+    /// The RVA of the continuation target after the handler is executed.
+    ///
+    /// This is used for control transfer (e.g., continuation blocks, to resume execution after `finally`).
+    pub target: u32,
+}
```
DreydenGys, line 121:
```suggestion
pub struct ScopeTableEntry {
    /// The starting RVA (relative virtual address) of the protected code region.
    ///
    /// This marks the beginning of a `try` block.
    pub begin: u32,
    
    /// The ending RVA (exclusive) of the protected code region.
    ///
    /// This marks the end of the `try` block.
    pub end: u32,
    
    /// The RVA of the exception handler function.
    ///
    /// e.g., be invoked when an exception occurs in the associated code range.
    pub handler: u32,
    
    /// The RVA of the continuation target after the handler is executed.
    ///
    /// This is used for control transfer (e.g., continuation blocks, to resume execution after `finally`).
    pub target: u32,
}
```
Other structures in the file seem to space out fields, you should probably stick with it too. It also makes it easier to read.

src/pe/exception.rs
```diff
@@ -94,6 +94,67 @@ const RUNTIME_FUNCTION_SIZE: usize = 12;
...
+
+impl Iterator for ScopeTableIterator<'_> {
+    type Item = error::Result<ScopeTableEntry>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.data.is_empty() {
+            return None;
+        }
+
+        Some(match self.data.pread_with(0, scroll::LE) {
```
DreydenGys, line 137: Instead of mutating data each time maybe you could store an offset and use the `gread_with` method like it is done in the `UnwindCodeIterator`?

src/pe/exception.rs
```diff
@@ -94,6 +94,66 @@ const RUNTIME_FUNCTION_SIZE: usize = 12;
...
+    type Item = error::Result<ScopeTableEntry>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.offset >= self.data.len() {
+            return None;
+        }
+
+        Some(
+            self.data
+                .gread_with(&mut self.offset, scroll::LE)
```
DreydenGys, line 143: If it fail during the iterator, will it fail indefinitely on the same offset? (I think it can also happen with the `UnwindCodeIterator`)

kkent030315: Also, the iterator assumes the bytes fed is multiple of `mem::size_of::<ScopeTableEntry>()` so it practically have no chance of returning `None`. If it ever happens, that's a bug in the higher implementation.

src/pe/exception.rs
```diff
@@ -624,6 +684,20 @@ impl<'a> UnwindInfo<'a> {
...
         }
     }
+
+    /// Returns an iterator over C scope table entries in this unwind info.
+    ///
+    /// If this unwind info has no [UnwindHandler::ExceptionHandler], this will always return `None`.
+    pub fn c_scope_table_entries(&self) -> Option<ScopeTableIterator<'a>> {
+        let UnwindHandler::ExceptionHandler(_, data) = self.handler? else {
+            return None;
+        };
```
m4b, line 694: nice use of let else + ? in Option returning func :) it did throw me off at first, i hadn't seen this pattern together at first so I parsed it as a double Option, then I realized handler is an Optional enum.

src/pe/exception.rs
m4b: Thank you for bringing up this question @DreydenGys , it's a good one to think about, and it did take me a bit to realize @kkent030315 answer is correct, the pread can't really fail, since when constructing the iterator, it guarantees `count * sizeof ScopeTableEntry` exist. Consequently, it might actually be better to unwrap the pread, make a comment to this effect of why it can't fail, and have the return type just be a `ScopeTableEntry` instead of `Result<ScopeTableEntry>`. This will also incidentally be a more ergonomic api for users since they don't have to deal with an error. If we do this, I might suggest having constructing the iterator return a result instead of an option, as that would provide more information about what failed (not enough bytes, etc.). Alternatively can keep the Option return type when constructing the iterator but it does have less information.. What do you think @kkent030315 ?

kkent030315: @m4b I agree. Like I, and you said, it's practically safe to `.unwrap()` in the pread inside the iterator and make the `type Item = ScopeTableEntry;`.

src/pe/exception.rs
```diff
@@ -94,6 +94,68 @@ const RUNTIME_FUNCTION_SIZE: usize = 12;
...
+
+impl Iterator for ScopeTableIterator<'_> {
+    type Item = ScopeTableEntry;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.offset >= self.data.len() {
+            return None;
+        }
+
+        // It is guarabteed that .expect here is really a unreachable.
```
m4b, line 141: nit: typo

m4b (approved):

m4b (approved):
Thank you! :heart:

m4b:
NB: non-breaking

PR #478, opened 2025-06-21
feat: add no-std error support

hannahfluch:
Currently, only the `std::error::Error` trait is implemented for custom goblin error. Therefore the crate cannot be used with e.g. thiserror in `no-std` environments. Causing this error:
```
error[E0599]: the method `as_dyn_error` exists for reference `&goblin::error::Error`, but its trait bounds were not satisfied
  --> uefi-loader/src/error.rs:42:12
   |
42 |     Goblin(#[from] goblin::error::Error),
   |            ^^^^^^^ method cannot be called on `&goblin::error::Error` due to unsatisfied trait bounds
   |
  ::: /home/hannah/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/goblin-0.10.0/src/error.rs:14:1
   |
14 | pub enum Error {
   | -------------- doesn't satisfy `goblin::error::Error: AsDynError<'_>` or `goblin::error::Error: core::error::Error`
   |
   = note: the following trait bounds were not satisfied:
           `goblin::error::Error: core::error::Error`
           which is required by `goblin::error::Error: AsDynError<'_>`
           `&goblin::error::Error: core::error::Error`
           which is required by `&goblin::error::Error: AsDynError<'_>`
``` 

 When the `std` feature is disabled, it is more practical to implement the `core::error::Error` trait instead.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -2,12 +2,13 @@
 //!
 
 use alloc::string::String;
+#[cfg(not(feature = "std"))]
+use core::error;
 use core::fmt;
 use core::num::TryFromIntError;
 use core::result;
 #[cfg(feature = "std")]
 use std::{error, io};
-
 #[non_exhaustive]
 #[derive(Debug)]
 /// A custom Goblin error
@@ -25,10 +26,10 @@ pub enum Error {
     BufferTooShort(usize, &'static str),
 }
 
-#[cfg(feature = "std")]
 impl error::Error for Error {
     fn source(&self) -> Option<&(dyn error::Error + 'static)> {
         match *self {
+            #[cfg(feature = "std")]
             Error::IO(ref io) => Some(io),
             Error::Scroll(ref scroll) => Some(scroll),
             _ => None,
--- a/src/error.rs
+++ b/src/error.rs
@@ -2,13 +2,10 @@
 //!
 
 use alloc::string::String;
-#[cfg(not(feature = "std"))]
 use core::error;
 use core::fmt;
 use core::num::TryFromIntError;
 use core::result;
-#[cfg(feature = "std")]
-use std::{error, io};
 #[non_exhaustive]
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -2,12 +2,13 @@
 //!
 
 use alloc::string::String;
+#[cfg(not(feature = "std"))]
+use core::error;
```
m4b, line 6: I believe you can remove the cfg above this, and the entire cfg and import below by simply doing std::io::Error

src/error.rs
```diff
@@ -2,12 +2,13 @@
...
 use core::fmt;
 use core::num::TryFromIntError;
 use core::result;
 #[cfg(feature = "std")]
 use std::{error, io};
```
m4b, line 11: I think we can delete this entire line and cfg by writing full import on the io variant of the error, eg `IO(std::io::Error)`

src/error.rs
```diff
@@ -25,10 +26,10 @@ pub enum Error {
     BufferTooShort(usize, &'static str),
 }
 
-#[cfg(feature = "std")]
 impl error::Error for Error {
     fn source(&self) -> Option<&(dyn error::Error + 'static)> {
         match *self {
+            #[cfg(feature = "std")]
```
m4b, line 32: This is interesting, normally adding a cfg on a variant arm is a breaking change but since this entire impl had the cfg before I don’t think this can in principle break anyone ? Eg this change should be a backwards compatible change and hence minor update

m4b (changes requested):
Thanks for this. I thought we’d done this already but must have gotten forgotten lost! I definitely remember someone mentioning core now has error but maybe that was in scroll?

Unless I’m mistaken I think the changes I suggested are equivalent but less cfg so let’s go with that if so

m4b (approved):
This is great thank you!

PR #479, opened 2025-07-14
Add/Extend ELF/PE permissive parsing mode to better handle packed, broken, or malware samples

chf0x:
@m4b I’ve been using Goblin to work with malware, and over time I’ve encountered a number of cases where parsing fails. In the `strstab` issue, you mentioned that you might be open to adding a permissive mode to address such pain points. I’d like to reopen this pull request with the changes I’ve made over time to enable parsing of broken binaries.

There are still a few places where the permissive flag is missing, for example, in strtab. I’d be happy to add those if you think this is worth merging into the main project.

I also probably would need to update tests.

Thank you!


src/pe/header.rs
```diff
@@ -461,16 +461,34 @@ impl<'a> DosStub<'a> {
...
         let end_offset = pe_pointer as usize;
 
         // Check end_offset is not greater than start_offset
         // end_offset == start_offset may mean there is no dos stub
         if end_offset < start_offset {
-            return Err(error::Error::Malformed(format!(
-                "PE pointer ({:#x}) cannot be before the DOS stub start ({:#x})",
-                pe_pointer, start_offset
-            )));
+            if matches!(opts.parse_mode, crate::pe::options::ParseMode::Permissive) {
```
kkent030315, line 480: I think of this a bit. Do we really need to allow PE pointer pointing out the data inside DOS header? In that case the binary won't even run, nor a valid PE file at all. Rest of thing will all fail.

If that's the case, can we move this into `Header::parse_with_opts`? I'd like to keep everything in the highest abstracted function as possible as it can ;)

```diff
    /// Parse a PE header from the underlying bytes with parsing options
    pub fn parse_with_opts(bytes: &'a [u8], opts: &crate::pe::options::ParseOptions) -> error::Result<Self> {
        let dos_header = DosHeader::parse(&bytes)?;
-       let dos_stub = DosStub::parse_with_opts(bytes, dos_header.pe_pointer, opts)?;
+       let dos_stub = DosStub::parse_with_opts(bytes, dos_header.pe_pointer, opts)
+           .map(Some)
+           .or_else(|e| match opts.parse_mode {
+               crate::pe::options::ParseMode::Permissive => {
+                   log::warn!("LoadConfig parse failed in permissive mode: {:?}", e);
+                   Ok(None)
+               }
+               _ => Err(e),
+           })?
+           .unwrap_or_default();

        Header::parse_impl(bytes, dos_header, dos_stub, true, opts)
    }
```

kkent030315 (changes requested):
Thank you for the PR! I reviewed it partly. I'll do more review at the rest of time.

src/pe/header.rs
chf0x: I once had a binary that I needed to analyze. While you’re right that it wouldn’t run, it was still useful to load it for further analysis. I’ll try to find it as an example

src/pe/relocation.rs
```diff
@@ -339,24 +341,95 @@ impl<'a> RelocationData<'a> {
...
+                    )));
+                }
+            }
+        };
+
+        let available_size = if offset + (dd.size as usize) <= bytes.len() {
+            dd.size as usize
+        } else {
+            // If directory size is wrong, find the containing .reloc section to get its size as a fallback
+            let section = sections.iter()
```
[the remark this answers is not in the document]
chf0x: The issue was that parsing failed on certain binaries because of an incorrect relocation size during the `PE:parse` step. This change is intended to make the analysis more resilient. I agree that anything can be faked or broken, so if you feel this part isn’t useful, I can remove it or place it under a "permissive" (or similar) flag. Your call, I’m happy with either decision

kkent030315 (changes requested):
Some minor changes and questions, LGTM otherwise.

kkent030315:
I think `INVALID_PE_SIGNATURE` needs rollback as well.

src/elf/options.rs
```diff
@@ -0,0 +1,50 @@
+//! ELF parsing options and modes
```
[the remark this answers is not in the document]
chf0x: I added global options.rs and moved ParseOptions to it. For now, I haven’t touched pe/options.rs, since changing it would be a breaking change. It’s probably better to handle that in a separate PR after this one is merged

src/pe/clr.rs
```diff
@@ -490,6 +490,7 @@ impl FusedIterator for ClrSectionIterator<'_> {}
 #[cfg(test)]
 mod tests {
     use super::*;
+    use alloc::vec::Vec;
```
kkent030315, line 493: Maybe you should revert all of those changes in tests?

src/pe/header.rs
```diff
@@ -1385,6 +1436,8 @@ mod tests {
         },
     };
 
+    use alloc::vec::Vec;
```
kkent030315, line 1439: This one too?

src/pe/header.rs
```diff
@@ -4,7 +4,10 @@ use core::iter::FusedIterator;
 use scroll::{IOread, IOwrite, Pread, Pwrite, SizeWith, ctx};
 
 use crate::error;
-use crate::pe::{data_directories, debug, optional_header, section_table, symbol};
+use crate::error::Permissive;
+#[cfg(feature = "te")]
+use crate::pe::data_directories;
+use crate::pe::{debug, optional_header, section_table, symbol};
```
kkent030315, line 10: Is this change really required? Maybe it should suffice just:

```suggestion
use crate::error::Permissive;
use crate::pe::{data_directories, debug, optional_header, section_table, symbol};
```

src/mach/exports.rs
kkent030315, line 1: Now you can remove those newlines in diff so this file will be gone from patch :)

src/pe/debug.rs
```diff
@@ -175,24 +176,48 @@ impl<'a> DebugData<'a> {
...
+                    return Err(error::Error::Malformed(format!(
                         "Cannot map ImageDebugDirectory rva {:#x} into offset",
                         dd.virtual_address
-                    ))
-                })?;
+                    )))?;
+                }
+            };
+
+        use crate::error::Permissive;
```
kkent030315, line 189: Please move it to the top-level import :)

src/pe/import.rs
```diff
@@ -77,10 +76,47 @@ pub struct HintNameTableEntry<'a> {
...
+        if *offset > bytes.len() {
+            return Err(error::Error::Malformed(format!(
+                "HintNameTableEntry name at offset {:#x} is beyond file bounds (file size: {:#x}). \
+                This may indicate a packed binary.",
+                offset,
+                bytes.len()
+            )));
+        }
+
+        use crate::error::Permissive;
```
kkent030315, line 112: Same; please move it to top-level import :)

src/pe/import.rs
```diff
@@ -119,7 +155,24 @@ impl<'a> SyntheticImportLookupTableEntry<'a> {
         let le = scroll::LE;
         let offset = &mut offset;
         let mut table = Vec::new();
+        use crate::error::Permissive;
```
kkent030315, line 158: Same; please move it to top-level import :)

src/pe/import.rs
```diff
@@ -351,21 +440,50 @@ impl<'a> ImportData<'a> {
             "import_directory_table_rva {:#x}",
             import_directory_table_rva
         );
+        use crate::error::Permissive;
```
kkent030315, line 443: Same; please move it to top-level import :)

src/pe/load_config.rs
```diff
@@ -3,10 +3,12 @@ use scroll::{Pread, Pwrite, SizeWith};
 
 use crate::container::{Container, Ctx};
 use crate::error;
+
```
kkent030315, line 6: You can remove newline here

src/pe/mod.rs
```diff
@@ -661,6 +694,7 @@ impl<'a> Coff<'a> {
 mod tests {
     use super::Coff;
     use super::PE;
+    use alloc::vec::Vec;
```
kkent030315, line 697: This one too; Maybe you should `use super::*;` and remove import in fn body.

src/pe/mod.rs
```diff
@@ -848,4 +882,48 @@ mod tests {
         assert_eq!(binary.entry, 0x1020u32);
         assert_eq!(binary.image_base, 0x140000000u64);
     }
+
+    #[test]
+    fn test_packed_binary_permissive_parsing() {
+        use crate::pe::options::{ParseMode, ParseOptions};
```
kkent030315, line 888: Maybe you should `use super::*;` and remove import in fn body.!

src/lib.rs
```diff
@@ -113,6 +113,8 @@ pub mod error;
 
 pub mod strtab;
 
+pub mod options;
+
```
kkent030315, line 117: Could you put this right after the `pub mod error;`?

kkent030315 (changes requested):
Thanks, the diff look very clean now; With those changes it would be cleanest I think.

kkent030315:
@chf0x by the way what IDE are you on? If you're under VSCode or Zed you might install [rust-analyzer](https://marketplace.visualstudio.com/items?itemName=rust-lang.rust-analyzer) extension (if you don't have one yet). It looks like you're having slightly different formatting rule so. :)

kkent030315 (approved):
Thanks very much, the patch looks very clean!

src/pe/header.rs
kkent030315: Ah, okay, so for those TE-specific datas and to avoid warn on unused import without TE feature flag. Makes sense but ig you can remove it from the patch.

https://github.com/m4b/goblin/blob/090a1d437e8c5ab35af6fa0996f5d811230b7264/src/pe/header.rs#L1194-L1197

https://github.com/m4b/goblin/blob/090a1d437e8c5ab35af6fa0996f5d811230b7264/src/pe/header.rs#L1215-L1219

src/pe/utils.rs
```diff
@@ -147,6 +147,59 @@ pub fn try_name<'a>(
     }
 }
 
+/// Safe version of try_name that handles packed binaries gracefully
+pub fn safe_try_name<'a>(
```
m4b, line 151: is this a function you use? I don't see it used anywhere, and i'd prefer not to add more freestanding `pub` functions unless absolutely necessary.

src/pe/options.rs
```diff
@@ -1,3 +1,5 @@
+pub use crate::options::ParseMode;
```
m4b, line 1: we should reevaluate the `pub` use here and elsewhere; for now `pub use` is required because it's a breaking change otherwise since `options` is a `pub mod`

src/pe/header.rs
```diff
@@ -1051,13 +1086,22 @@ impl<'a> RichHeader<'a> {
     /// - The upper 16 bits of the tool ID describe the tool type,
     /// - while the lower 16 bits specify the tool’s build version.
     pub fn parse(bytes: &'a [u8]) -> error::Result<Option<Self>> {
+        Self::parse_with_opts(bytes, &crate::pe::options::ParseOptions::default())
+    }
+
+    pub fn parse_with_opts(
```
m4b, line 1092: `pub(crate)`

m4b (changes requested):
ok the major thing is the non-exhaustive which I realized must be removed to make this backwards compat, moving the (`pub(crate)`) helper trait into options for now, and some other minor nits, mostly reducing the api surface from `pub` to `pub(crate)`.

thanks for your patience, this should be the last review :)

src/options.rs
```diff
@@ -0,0 +1,85 @@
+//! Unified parsing options for binary formats
+//!
+//! This module provides common parsing options that can be used across
+//! different binary formats (ELF, PE, Mach-O, etc.).
+
+/// Binary parsing mode
+#[non_exhaustive]
```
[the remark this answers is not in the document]
chf0x: But I think the current code is backward-compatible, since I didn’t remove options.rs from PE. So the old API should remain the same as before. Or do you mean you’d like to remove the ELF options completely, since that’s a new API?

src/pe/import.rs
```diff
@@ -77,10 +76,45 @@ pub struct HintNameTableEntry<'a> {
...
+                "HintNameTableEntry name at offset {:#x} is beyond file bounds (file size: {:#x}). \
+                This may indicate a packed binary.",
+                offset,
+                bytes.len()
+            )));
+        }
+
+        let name = bytes.pread::<&'a str>(*offset).or_permissive_and_value(
+            opts.parse_mode.is_permissive(),
+            &format!("Invalid UTF-8 in import name at offset {:#x}", offset),
```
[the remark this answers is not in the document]
chf0x: That is a good call, I removed allocations like this. Thanks!

kkent030315 (approved):
LGTM. With the permissive trait the codebase looks very clean now! (maybe you can fold them down the comments in `options.rs`?)

chf0x:
Awesome news, thanks for reviewing @m4b @kkent030315 ! The last requested changes are now in

m4b:
NB: backwards compatible

m4b:
NB: non-breaking, in 0.10.2

PR #480, opened 2025-07-18
PE: Remove unwind info alignment assertion

kkent030315:
I had an issue where below assertion being triggered parsing `rustc.exe`. The unwind info stands correct.

```bash
cargo run --example rdr -- "path/to/rustc.exe"
```

As commented in the code in the patch, developers sometimes forget to align handy-written unwind infos (which becomes an edge case because is really a compiler's job to generate such), but the unwind info is correct anyway and we should not have that assertion.

https://github.com/m4b/goblin/blob/8e606d8a5aa2158f64ef82f6e279c6b182cc7b3f/src/pe/exception.rs#L576

---

Edit: temporaly fix is to use `[profile.dev] debug-assertions = false` in `Cargo.toml`

### DIFF
--- a/src/pe/exception.rs
+++ b/src/pe/exception.rs
@@ -53,6 +53,8 @@ use crate::pe::options;
 use crate::pe::section_table;
 use crate::pe::utils;
 
+/// **N**o handlers.
+const UNW_FLAG_NHANDLER: u8 = 0x00;
 /// The function has an exception handler that should be called when looking for functions that need
 /// to examine exceptions.
 const UNW_FLAG_EHANDLER: u8 = 0x01;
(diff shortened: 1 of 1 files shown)


kkent030315:
For the case of `rustc.exe` (is really a proxy, originally `rustup-init.exe` from [Rust Official](https://github.com/rust-lang/rustup).

The unwind info come from [aws-lc-rs](https://github.com/aws/aws-lc-rs) -- is binding for [aws-lc](https://github.com/aws/aws-lc).

They align the first one, but not for the rest.
https://github.com/aws/aws-lc/blob/f10416a1d49a4d080719a0c7d888569e0c60bfcc/generated-src/win-x86_64/crypto/fipsmodule/ghash-x86_64.asm#L1230-L1330

m4b (approved):
Thanks for updating comments!

m4b:
NB: non-breaking change

PR #483, opened 2025-07-26
PE: fix load config parser out of bounds

### LINKED ISSUE #481: pe/load_config.rs: unchecked slice indexing → panic DoS (same root cause as #468)
https://github.com/m4b/goblin/blob/09bd7649d2afb6af288d87a0b01162a0d8407a0f/src/pe/load_config.rs#L339

While fuzzing goblin v0.10.0, I hit a panic in pe/load_config.rs caused by an unchecked slice index. This is the same root cause and fix pattern as #468 (bounds check before slicing). Impact is a panic / DoS only (safe Rust).

```
thread 'main' panicked at goblin/src/pe/load_config.rs:339:26:
range start index 16842788 out of range for slice of length 412
```

Crashing file
[goblin_pe_crash.exe.zip](https://github.com/user-attachments/files/21434069/goblin_pe_crash.exe.zip)

kkent030315:
- Resolves #481 
- Fixed invalid use of `.pread_with::<&[u8]>`
- ~~`PE::load_config_data` is now `None` if failing to parse.~~

### DIFF
--- a/src/pe/load_config.rs
+++ b/src/pe/load_config.rs
@@ -336,8 +336,8 @@ impl LoadConfigData {
                         dd.virtual_address
                     ))
                 })?;
-        let bytes = bytes[offset..]
-            .pread_with::<&[u8]>(0, dd.size as usize)
+        let bytes = bytes
+            .pread_with::<&[u8]>(offset, dd.size as usize)
             .map_err(|_| {
                 error::Error::Malformed(format!(
                     "load config offset {:#x} and size {:#x} exceeds the bounds of the bytes size {:#x}",


src/pe/mod.rs
```diff
@@ -273,14 +273,15 @@ impl<'a> PE<'a> {
...
+                load_config_data = load_config::LoadConfigData::parse_with_opts(
                     bytes,
                     load_config_dir,
                     &sections,
                     file_alignment,
                     opts,
                     is_64,
-                )?);
+                )
+                .ok();
```
m4b, line 284: Are you sure we want to drop all errors when parsing and return none if malformed ? It’s ok to gracefully error when fuzzing (eg your fix is good) but for regular uses if the config is malformed should we not continue parsing ?

Or perhaps we should check the strict flag to stop otherwise return none, seems like a good compromise ?

src/pe/load_config.rs
```diff
@@ -336,8 +336,8 @@ impl LoadConfigData {
                         dd.virtual_address
                     ))
                 })?;
-        let bytes = bytes[offset..]
-            .pread_with::<&[u8]>(0, dd.size as usize)
+        let bytes = bytes
```
m4b, line 339: Yea I think this is the correct change to use everywhere, as I’ve mentioned elsewhere. We should probably do a pass looking for patterns like this and convert to full pread as you did here

src/pe/mod.rs
kkent030315: It's good if we could defer the parse in some way so that the user could parse the rest of binary. This change is not mandatory so I am going to revert this part atm

m4b (approved):
TY!

m4b:
NB: non-breaking

PR #486, opened 2025-08-19
Don't return err when parsemode is permissive

Kharosx0:
Ignore more errors and continue parsing when ParseMode is Permissive (e.g. an erroneous resource directory should not prevent parsing of the PE when ParseMode is Permissive).

### DIFF
--- a/src/pe/mod.rs
+++ b/src/pe/mod.rs
@@ -110,7 +110,6 @@ impl<'a> PE<'a> {
             + header::SIZEOF_COFF_HEADER;
         let offset =
             &mut (optional_header_offset + header.coff_header.size_of_optional_header as usize);
-
         let sections = header.coff_header.sections(bytes, offset)?;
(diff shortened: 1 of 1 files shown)


kkent030315:
Probably a duplicate of #479?

chf0x:
Not exactly a duplicate, I’d say this is complementary to #479. In my pull request, I introduced granular permissive handling within individual parsers, whereas this provides a higher-level safety net for cases that remain uncovered

m4b:
needs rustfmt

m4b:
@Kharosx0 this needs a rebase now; after that let's re-review now that permissive stuff has landed. You may be able to use the new trait to help with this, not sure.

m4b:
@Kharosx0 are you still interested in this? If so, can you checkout the latest goblin, and see if the new parsemode permissive changes are working like your PR, as they may already be implementing this, thanks!

m4b:
closing as per above comment, I think this may be fixed? please reopen with issue if still encounter

PR #488, opened 2025-08-20
PE: Expose load config data

kkent030315:
`LoadConfigData::directory` was accidentally a private field in #464 (https://github.com/m4b/goblin/pull/464#issuecomment-2954702946) and it is inaccessible outside of the crate so I made it a `pub` field.


m4b:
NB: backwards compatible

m4b:
NB: non-breaking, in 0.10.2

PR #489, opened 2025-08-21
PE: Add parser for dynamic relocation (DVRT)

kkent030315:
This PR adds support for dynamic value relocation table (DVRT) for PE32/64. DVRT info is provided in load config directory. Generally speaking, DVRT is a feature to mitigate Spectre variant 2 (branch target injection), for example, by dynamically relocating indirect jumps or calls to ”retpoline” pages. DVRT contains where such indirect call targets and jumps are located within the executable.

There are 8 types of dynamic relocation as of now, which this PR adds 5 of them. DVRT has 2 versions as of now (`1` and `2`) and this PR currently only supports `1`.
- [x] DVRT V1
	- [x] `IMAGE_DYNAMIC_RELOCATION_GUARD_RF_PROLOGUE`
	- [x] `IMAGE_DYNAMIC_RELOCATION_GUARD_RF_EPILOGUE`
	- [x] `IMAGE_DYNAMIC_RELOCATION_GUARD_IMPORT_CONTROL_TRANSFER`
	- [x] `IMAGE_DYNAMIC_RELOCATION_GUARD_INDIR_CONTROL_TRANSFER`
	- [x] `IMAGE_DYNAMIC_RELOCATION_GUARD_SWITCHTABLE_BRANCH`
	- [x] `IMAGE_DYNAMIC_RELOCATION_ARM64X`
	- [x] `IMAGE_DYNAMIC_RELOCATION_FUNCTION_OVERRIDE`
	- [x] `IMAGE_DYNAMIC_RELOCATION_ARM64_KERNEL_IMPORT_CALL_TRANSFER`
- [ ] DVRT V2
	- [ ] `IMAGE_DYNAMIC_RELOCATION_GUARD_RF_PROLOGUE`
	- [ ] `IMAGE_DYNAMIC_RELOCATION_GUARD_RF_EPILOGUE`
	- [ ] `IMAGE_DYNAMIC_RELOCATION_GUARD_IMPORT_CONTROL_TRANSFER`
	- [ ] `IMAGE_DYNAMIC_RELOCATION_GUARD_INDIR_CONTROL_TRANSFER`
	- [ ] `IMAGE_DYNAMIC_RELOCATION_GUARD_SWITCHTABLE_BRANCH`
	- [ ] `IMAGE_DYNAMIC_RELOCATION_ARM64X`
	- [ ] `IMAGE_DYNAMIC_RELOCATION_FUNCTION_OVERRIDE`
	- [ ] `IMAGE_DYNAMIC_RELOCATION_ARM64_KERNEL_IMPORT_CALL_TRANSFER`

DVRT V2 and other relocation types are probably going to be implemented in follow-up PRs.

Reference:
- https://denuvosoftwaresolutions.github.io/DVRT/dvrt.html (Only DVRT v1; not entirely covered)

### DIFF
--- a/src/pe/dynreloc.rs
+++ b/src/pe/dynreloc.rs
@@ -1250,9 +1250,7 @@ impl DynRelocRelocation {
             DynRelocRelocation::Branch(data) => Some(data.record.page_relative_offset() as _),
             DynRelocRelocation::Prologue(data) => Some(data.record.offset() as _),
             DynRelocRelocation::Epilogue(data) => Some(data.record.offset() as _),
-            DynRelocRelocation::KernelImport(data) => {
-                Some(data.record.page_relative_offset() as _)
-            }
+            DynRelocRelocation::KernelImport(data) => Some(data.record.page_relative_offset() as _),
         }
     }
 }


src/pe/dynreloc.rs
```diff
@@ -0,0 +1,1258 @@
...
+    }
+
+    /// Gets the CfgCheck field (bit 14)
+    /// True if Control Flow Guard check is required
+    pub fn cfg_check(&self) -> bool {
+        (self.0 >> 14) & 1 != 0
+    }
+
+    /// Sets the CfgCheck field (bit 14)
+    pub fn set_cfg_check(&mut self, cfg_check: bool) {
```
m4b, line 547: I might opt for the builder pattern here, but I don't know how usage patterns are generally, do you normally mutate it a few times occasionally in different situations or is it generally constructed with these flags once? If latter I'd suggest to use builder pattern since it's more convenient. If it's mutated in different situations/conditions, set style is probably better

src/pe/dynreloc.rs
```diff
@@ -0,0 +1,1258 @@
...
+
+                DynRelocRelocation::Branch(DynRelocBranch {
+                    symbol: self.symbol,
+                    block_rva: self.block_rva,
+                    record,
+                })
+            }
+            IMAGE_DYNAMIC_RELOCATION_FUNCTION_OVERRIDE => {
+                // This is unreachable. See impl Iterator for DynRelocBlockIterator.
+                unimplemented!()
```
m4b, line 1085: you're 100% sure this is not reachable? To be clear, I mean even with malformed, fuzzer generated binaries etc., this should truly be impossible to reach.

m4b:
this probably needs a rebase/my comments still need some clarification. we can focus on the other outstanding PR for delay import parser in the meantime though :)

PR #492, opened 2025-10-01
PE: Implement `UWOP_EPILOG` parsing for unwind info v2

kkent030315:
Implemented to-do of `UWOP_EPILOG` in unwind info v2 along with its tests.

Reference:
- https://github.com/BlancLoup/weekly-geekly.github.io/blob/1cbdf1c6127fcdaeda1f01bcbb006febd17d5a95/articles/322956/index.html
- https://github.com/Montura/cpp/blob/4393c678ee8e44dd98feb7a198c3983a6108cef8/src/exception_handling/msvc/eh_msvc_cxx_EH_x64.md?plain=1#L119

### DIFF
--- a/src/pe/exception.rs
+++ b/src/pe/exception.rs
@@ -333,10 +333,45 @@ pub enum UnwindOperation {
     /// Save the lower 64 bits of a nonvolatile XMM register on the stack.
     SaveXMM(Register, StackFrameOffset),
 
-    /// Describes the function epilog.
+    /// Describes the function epilog location and size (Version 2).
     ///
-    /// This operation has been introduced with unwind info version 2 and is not implemented yet.
-    Epilog,
+    /// [UWOP_EPILOG] entries work together to describe where epilogues are located in the function,
+    /// but not what operations they perform. The actual operations performed during epilogue
+    /// execution are the reverse of the prolog operations.
+    ///
+    /// These entries always appear at the beginning of the [UnwindCode] array, with a minimum of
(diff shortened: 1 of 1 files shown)


src/pe/exception.rs
```diff
@@ -333,10 +333,45 @@ pub enum UnwindOperation {
...
+    /// (lower 8 bits) with the lower 4 bits of `offset_high_or_flags` (upper 4 bits). The epilogue's
+    /// starting address is calculated by subtracting this offset from the function's `EndAddress`.
+    Epilog {
+        /// For the first [UWOP_EPILOG] entry:
+        /// * Size of the epilogue in bytes
+        /// * If `offset_high_or_flags` bit 0 is set, also serves as offset
+        ///
+        /// For subsequent entries:
+        /// * Lower 8 bits of the offset from function end
+        offset_low_or_size: u8,
```
m4b, line 361: unfortunately this is a breaking change so won't make it into the minor release. Also, if you anticipate in the future any enum variants/cases being added in the future, I strongly encourage to put a `#[non_exhaustive]` annotation on this `pub` enum so that we can safely add new cases in the future. If you don't anticipate it ever again having other variants, then we can leave it without

m4b (approved):
LGTM, thanks for implementing this!

kkent030315:
Ok I agree and added `non_exhaustive`. Tests indicate that this is correct implementation to parse unwind codes.

PR #493, opened 2025-10-02
Fixed getting section data tainted section lifetime

Hexorg:
Original lifetime specification was wrong and made the compiler think the returned lifetime had to live as long as the section data, which it did not.

Before the fix, the following code did not compile:

```rs
// ... open PE file
 let mut result = Vec::new();
 for section in pe.sections {
     result.push(section.data(file_bytes)); // compiler complains that section needs to live as long as result.
 }
 
 return result;
 ```
 
 After the fix getting section data can outlive the section.

### DIFF
--- a/src/pe/section_table.rs
+++ b/src/pe/section_table.rs
@@ -99,7 +99,7 @@ impl SectionTable {
         Ok(table)
     }
 
-    pub fn data<'a, 'b: 'a>(&'a self, pe_bytes: &'b [u8]) -> error::Result<Option<Cow<'a, [u8]>>> {
+    pub fn data<'b>(&self, pe_bytes: &'b [u8]) -> error::Result<Option<Cow<'b, [u8]>>> {
         let section_start: usize = self.pointer_to_raw_data.try_into().map_err(|_| {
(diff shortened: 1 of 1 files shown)


kkent030315 (approved):

m4b (approved):
nice, thank you for the test! i think this is non-breaking since it makes more programs acceptable, and not less (so it won't break someone?)

m4b:
> @m4b no problem happens to the best of us! Shameless self-plug but I'm trying to make a decompiler at https://github.com/Hexorg/Ouroboros and my current goal is to use Goblin to load binary files.

Neat I'll check it out, I love decompiler tech. Also, I approve of this name :) Several years ago I was going to start a decompiler project, and I was in fact going to call it ouroboros, because that is obviously the best name ever for a decompiler project :)  Alas I never got around to it; same with a (static, I already played around with a dynamic linker in rust, it's ancient tech from like 1.0 rust days, called dryad if curious)) linker I was going to write (too many things!), but thankfully `wild` exists which is very nice.

Anyway, thanks for your contribution, much appreciated.

PR #495, opened 2025-10-09
elf.reloc: add some missing PowerPC 32-bit relocation type constants

ivlzme:
Also consolidated a few PPC and PPC64 constants which overlap


ivlzme:
> Sorry I missed this; this would have been fine to go into the emergency release i just released, but i hope you can wait a bit? otherwise I can make another minor release, but won't be tonight

No problem at all. You can add this to your next normal release. Thank you for maintaining this project!

PR #496, opened 2025-10-16
pe.reloc: fix parsing relocs loading 0 bytes

PJB3005:
I tried using the reloc feature but it always just parsed an empty slice for the actual reloc byte contents.

Test code:

```rust
use std::{fs, os};

fn main() {
    let bytes = fs::read("../target/debug/scratch.exe").expect("Failed to load scratch.exe!");
    let pe = goblin::pe::PE::parse(&bytes).expect("Failed to load PE!");
    let relocs = pe.relocation_data.expect("Failed to load relocations!");
    println!("{:?}", relocs.blocks().flat_map(|b| b.unwrap().words().map(|w| w.unwrap()).collect::<Vec<_>>()).collect::<Vec<_>>());
}
```

### DIFF
--- a/src/pe/relocation.rs
+++ b/src/pe/relocation.rs
@@ -380,8 +380,8 @@ impl<'a> RelocationData<'a> {
             )));
         }
 
-        let bytes = bytes[offset..]
-            .pread_with::<&[u8]>(0, safe_available_size)
+        let bytes = bytes
+            .pread_with::<&[u8]>(offset, safe_available_size)
             .map_err(|_| {
                 error::Error::Malformed(format!(
                     "base reloc offset {:#x} and size {:#x} exceeds the bounds of the bytes size {:#x}",


ToshibaMastr:
Maybe instead of `bytes[offset..].pread_with::<&[u8]>(0, safe_available_size)`, use `bytes.pread_with::<&[u8]>(offset, safe_available_size)` ?

m4b:
yes @ToshibaMastr 's suggestion is good! @PJB3005 can you update PR with that and also a test for this was nice, i'm surprised we don't have this covered. I should probably run code coverage to at some point to see the state of things

PJB3005:
> and also a test for this was nice

What would be the best way to make a `.exe` to test this? I see we have a couple in `tests/bin/pe`.

m4b:
ideally you can put the raw bytes (and ideally the binary is tiny) in a test, so it's all source code; after that can check in a really tiny binary. i've been saying for years I need to create a test suite, that e.g., we just git clone in CI that runs the test suite but that is one extra step to testing/contributing to goblin, but probably worth it :shrug:

m4b:
i quickly grepped this, we/i/someone should probably check similar cases and replace them with `pread_with(offset, etc)`
```
rg '\[offset\.\.\]'
src/elf/symver.rs
412:                let bytes: &'a [u8] = &iter.bytes[offset..];
664:                let bytes: &'a [u8] = &iter.bytes[offset..];

src/pe/exception.rs
659:            let data = &bytes[offset..];

src/pe/debug.rs
819:        let name = match self.data[offset..].iter().position(|&b| b == 0) {
842:        self.data = &self.data[offset..];

src/pe/tls.rs
226:                    (&bytes[offset..]).pread_with(0, size as usize).ok()

src/pe/relocation.rs
383:        let bytes = bytes[offset..]

src/pe/resource.rs
671:                self.data = &self.data[offset..];
```

kkent030315:
I am pretty sure this comes from previous permissive PR. It's on me I think I might overlooked that one during review.

For min binary you could compile it with MSVC following flags:

- compile: /GS- /GR-
- link: /stub:myminstub.bin /entry:main /nodefaultlib /baserelocclustering /novcfeature /nocoffgrpinfo /debug:none

There are more tricks to eliminate unnecessary bytes from the binary but it does pretty much good.

m4b:
I'm going to merge and do a quick release since it's a regression

PR #498, opened 2025-10-20
Fix potential out-of-bounds read in unwind/POGO info parser

### LINKED ISSUE #497: track down and eliminate (if required) `bytes[offset..]` style code
Here is a simplistic grep, some of these are guaranteed safe due to sizes checked by ctor of the iterator, for example, but should double verify and add a comment when that's true, and eliminate the rest:

```
src/pe/resource.rs
671:                self.data = &self.data[offset..];

src/elf/symver.rs
412:                let bytes: &'a [u8] = &iter.bytes[offset..];
664:                let bytes: &'a [u8] = &iter.bytes[offset..];

src/pe/tls.rs
226:                    (&bytes[offset..]).pread_with(0, size as usize).ok()

src/pe/debug.rs
819:        let name = match self.data[offset..].iter().position(|&b| b == 0) {
842:        self.data = &self.data[offset..];

src/pe/exception.rs
713:            let data = &bytes[offset..];
```

kkent030315:
Resolves #497.

### DIFF
--- a/src/pe/debug.rs
+++ b/src/pe/debug.rs
@@ -852,14 +852,13 @@ impl FusedIterator for POGOEntryIterator<'_> {}
 
 #[cfg(test)]
 mod tests {
-    use crate::pe::debug::POGOEntryIterator;
-
     use super::{
         CODEVIEW_PDB70_MAGIC, ExDllCharacteristicsInfo, IMAGE_DEBUG_POGO_SIGNATURE_LTCG,
         IMAGE_DEBUG_TYPE_CODEVIEW, IMAGE_DEBUG_TYPE_EX_DLLCHARACTERISTICS, IMAGE_DEBUG_TYPE_ILTCG,
         IMAGE_DEBUG_TYPE_POGO, IMAGE_DEBUG_TYPE_REPRO, IMAGE_DEBUG_TYPE_VC_FEATURE,
         IMAGE_DLLCHARACTERISTICS_EX_CET_COMPAT, IMAGE_DLLCHARACTERISTICS_EX_CET_COMPAT_STRICT_MODE,
-        ImageDebugDirectory, POGO_SIGNATURE_SIZE, POGOInfoEntry, ReproInfo, VCFeatureInfo,
+        ImageDebugDirectory, POGO_SIGNATURE_SIZE, POGOEntryIterator, POGOInfoEntry, ReproInfo,
+        VCFeatureInfo,
     };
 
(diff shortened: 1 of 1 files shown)


src/pe/debug.rs
```diff
@@ -810,18 +810,18 @@ impl<'a> Iterator for POGOEntryIterator<'a> {
...
         let name = match self.data[offset..].iter().position(|&b| b == 0) {
             Some(pos) => {
                 if offset + pos as usize >= self.data.len() {
                     return Some(Err(error::Error::Malformed(format!(
-                        "Null-terminator for POGO entry (rva {:#x} and size {:#X}) found but exceeds iterator buffer",
-                        rva, size
+                        "Null-terminator for POGO entry (rva {rva:#x} and size {size:#X}) found but exceeds iterator buffer",
                     ))));
                 }
                 let name = &self.data[offset..offset + pos + 1];
```
m4b: i think `let name = self.data.pread(offset, pos + 1).map_err` is clearer, and you can delete the check above. Because you're returning the error will require some handling; probably some combinator version, but a simple match would also suffice. or perhaps this is best?

```rust
let Some(name) = self.data.pread(offset, pos + 1).ok() else {
  return Some(Err(error::Error::Malformed(format!(
                        "Null-terminator for POGO entry (rva {rva:#x} and size {size:#X}) found but exceeds iterator buffer",
                    ))));
}
```

also directly below it `offset + pos + 1` is incremented; if turn `pread` into `gread` it'll do that for you :)

src/pe/debug.rs
```diff
@@ -832,13 +832,18 @@ impl<'a> Iterator for POGOEntryIterator<'a> {
             }
             None => {
                 return Some(Err(error::Error::Malformed(format!(
-                    "Cannot find null-terimnator for POGO entry (rva {:#x} and size {:#X})",
-                    rva, size
+                    "Cannot find null-terimnator for POGO entry (rva {rva:#x} and size {size:#X})",
```
m4b: nit typo: terimnator => terminator

src/pe/debug.rs
```diff
@@ -810,18 +810,18 @@ impl<'a> Iterator for POGOEntryIterator<'a> {
             Err(error) => return Some(Err(error.into())),
         };
 
-        if offset > self.data.len() {
+        // Use >= to avoid empty slice, that we want to emit an error early here for
+        // malformed name in a POGO entry.
+        if offset >= self.data.len() {
```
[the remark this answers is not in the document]
kkent030315: Added tests for this.

PR #499, opened 2025-11-03
pe: Reject cyclic resource trees

Mrmaxmeier:
Parsing these currently results in a hang. This follows the approach taken by the LIEF PE parser. Others like `trailofbits/pe-parse` strictly limit the resource tree depth to 3 instead (which is the max seen in the wild) to side-step this problem.


src/pe/mod.rs
```diff
@@ -923,4 +923,10 @@ mod tests {
             "Permissive parsing should succeed for packed binary"
         );
     }
+
+    #[test]
+    fn malformed_resource_tree() {
+        let data = include_bytes!("../../tests/bins/pe/malformed_resource_tree.exe.bin");
+        let _ = PE::parse(data);
+    }
```
kkent030315: This being succeeding though is a malformed binary. What is the intention of this test?

kkent030315 (approved):
LGTM

m4b (approved):
thank you!

PR #501, opened 2025-11-20
fix #500: Handle unpadded resource values in .NET assemblies compiled with Mono

### LINKED ISSUE #500: Handle unpadded resource values in .NET assemblies compiled with Mono
# Handle unpadded resource values in .NET assemblies compiled with Mono

## Problem

Goblin failed to parse certain .NET assemblies compiled with Mono's resource compiler, throwing a scroll error:
```
TooBig { size: 20, len: 18 }
```

### Root Cause

In `ResourceString::parse()`, the code assumes all resource string values are padded to 4-byte alignment according to the PE specification. However, Mono's resource compiler does not pad the **final** value in a `ResourceString` structure, causing out-of-bounds read attempts.

**Example from `System.Xml.XDocument.dll`:**
- ResourceString: "Assembly Version" → "4.0.11.0"
- `value_len` = 9 words = 18 bytes
- Code calculates: `align_up(18, 4)` = 20 bytes
- Actual data available: 18 bytes
- Result: scroll error when attempting to read 20 bytes

## Solution

Modified `ResourceString::parse()` to read only the available bytes when alignment padding is missing:

```rust
// src/pe/resource.rs, lines 839-848
let real_value_len = utils::align_up(
    if r#type == 1 {
        value_len as usize * SIZE_OF_WCHAR
    } else {
        value_len as usize
    },
    4,
);

// Some resource compilers (like mono) don't pad the last value to 4-byte alignment.
// If there aren't enough bytes for the aligned length, use what's available.
let actual_read_len = real_value_len.min(bytes.len() - *offset);
let value = bytes.pread_with::<&[u8]>(*offset, actual_read_len)?;
*offset += value.len();
```

## Testing

- All 125 existing tests pass
- Verified correct extraction of version info and string resources from `System.Xml.XDocument.dll`
- File: [System.Xml.XDocument.dll.zip](https://github.com/user-attachments/files/23642692/System.Xml.XDocument.dll.zip)

## Impact

- **Backward compatible**: No changes to parsing of well-formed PE resources
- **Fixes**: Parsing of .NET assemblies compiled with Mono (widespread in cross-platform .NET applications)
- **Scope**: 4-line change in `ResourceString::parse()`

BinFlip:
# Handle unpadded resource values in .NET assemblies compiled with Mono

## Problem

Goblin failed to parse certain .NET assemblies compiled with Mono's resource compiler, throwing a scroll error:
```
TooBig { size: 20, len: 18 }
```

### Root Cause

In `ResourceString::parse()`, the code assumes all resource string values are padded to 4-byte alignment according to the PE specification. However, Mono's resource compiler does not pad the **final** value in a `ResourceString` structure, causing out-of-bounds read attempts.

**Example from `System.Xml.XDocument.dll`:**
- ResourceString: "Assembly Version" → "4.0.11.0"
- `value_len` = 9 words = 18 bytes
- Code calculates: `align_up(18, 4)` = 20 bytes
- Actual data available: 18 bytes
- Result: scroll error when attempting to read 20 bytes

## Solution

Modified `ResourceString::parse()` to read only the available bytes when alignment padding is missing:

```rust
// src/pe/resource.rs, lines 839-848
let real_value_len = utils::align_up(
    if r#type == 1 {
        value_len as usize * SIZE_OF_WCHAR
    } else {
        value_len as usize
    },
    4,
);

// Some resource compilers (like mono) don't pad the last value to 4-byte alignment.
// If there aren't enough bytes for the aligned length, use what's available.
let actual_read_len = real_value_len.min(bytes.len() - *offset);
let value = bytes.pread_with::<&[u8]>(*offset, actual_read_len)?;
*offset += value.len();
```

## Testing

- All 125 existing tests pass
- Verified correct extraction of version info and string resources from `System.Xml.XDocument.dll`
- File: [System.Xml.XDocument.dll.zip](https://github.com/user-attachments/files/23642692/System.Xml.XDocument.dll.zip)

## Impact

- **Backward compatible**: No changes to parsing of well-formed PE resources
- **Fixes**: Parsing of .NET assemblies compiled with Mono (widespread in cross-platform .NET applications)
- **Scope**: 4-line change in `ResourceString::parse()`


BinFlip:
This addresses #500

src/pe/resource.rs
```diff
@@ -787,7 +787,11 @@ impl<'a> ResourceString<'a> {
             },
             4,
         );
-        let value = bytes.pread_with::<&[u8]>(*offset, real_value_len as usize)?;
+
+        // Some resource compilers (like mono) don't pad the last value to 4-byte alignment.
+        // If there aren't enough bytes for the aligned length, use what's available.
+        let actual_read_len = real_value_len.min(bytes.len() - *offset);
+        let value = bytes.pread_with::<&[u8]>(*offset, actual_read_len)?;
```
kkent030315: Instead of assuming that the string actually ends at the slice bound, I'd rather modifying padding directly.

You can avoid adding padding bytes if there are no spaces for it.

https://github.com/m4b/goblin/blob/589ffe5787d1d9ab2007a7f98051fa60d073938a/src/pe/resource.rs#L782-L789

m4b:
sorry, I merged a prior PR that now causes a conflict; @BinFlip would you mind rebasing? assuming @kkent030315 says good to go, we can merge then, thank you!

PR #504, opened 2025-12-03
PE: machinetypes - handle indeterminate types different from UNKNOWN

wesinator:
this PR updates the machine type PE header to differentiate the UNKNOWN (0x0) machine type value and other, "indeterminate" PE machine types.

disclaimer: not a Rust dev

[NO_TRAIN]::

### DIFF
--- a/src/pe/header.rs
+++ b/src/pe/header.rs
@@ -1419,7 +1419,7 @@ pub fn machine_to_str(machine: u16) -> &'static str {
         COFF_MACHINE_SH5 => "SH5",
         COFF_MACHINE_THUMB => "THUMB",
         COFF_MACHINE_WCEMIPSV2 => "WCE_MIPS_V2",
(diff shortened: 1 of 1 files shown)


wesinator:
> What is the purpose of this patch? How is that change of string representation could ever useful? Can’t you just match the COFF_UNKNOWN in downstream crates and replace to whatever you want?

Hmm, I didn't see the difference of `COFF_MACHINE_UNKNOWN` / `COFF_UNKNOWN`

A few comments though:
 - would it be useful to specify `machine` type in the default COFF_UNKNOWN label, or rename it to avoid confusion with the defined `UNKNOWN` machine type label?
 - do we expect consuming libraries to access the peheader `machine` u16 value at the same item level as the function provided by `machine_to_str`? I think this would be good so that the actual value could easily be used if it is an `unknown` case

apologies,

PR #507, opened 2025-12-19
riscv: fix relocation constants numbering

Jvlegod:
References:

1. https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc
2. https://riscv-non-isa.github.io/riscv-elf-psabi-doc/

RISC-V ELF psABI version 1.1 (November 26, 2025).

Given the current psABI specification, the [#PR165](https://github.com/m4b/goblin/pull/165) is no longer applicable.

### DIFF
--- a/src/elf/constants_relocation.rs
+++ b/src/elf/constants_relocation.rs
@@ -856,21 +856,21 @@ pub const R_RISCV_TPREL_LO12_I: u32 = 30;
 pub const R_RISCV_TPREL_LO12_S: u32 = 31;
 /// TLS LE thread usage: %tprel_add(symbol)
 pub const R_RISCV_TPREL_ADD: u32 = 32;
-/// 8-bit label addition: word8 = S + A
+/// 8-bit label addition: word8 = V + S + A
 pub const R_RISCV_ADD8: u32 = 33;
-/// 16-bit label addition: word16 = S + A
+/// 16-bit label addition: word16 = V + S + A
 pub const R_RISCV_ADD16: u32 = 34;
-/// 32-bit label addition: word32 = S + A
+/// 32-bit label addition: word32 = V + S + A
 pub const R_RISCV_ADD32: u32 = 35;
-/// 64-bit label addition: word64 = S + A
+/// 64-bit label addition: word64 = V + S + A
 pub const R_RISCV_ADD64: u32 = 36;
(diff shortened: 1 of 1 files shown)


src/elf/constants_relocation.rs
```diff
@@ -872,38 +872,46 @@ pub const R_RISCV_SUB16: u32 = 38;
 pub const R_RISCV_SUB32: u32 = 39;
 /// 64-bit label subtraction: word64 = S - A
 pub const R_RISCV_SUB64: u32 = 40;
-/// GNU C++ vtable hierarchy
-pub const R_RISCV_GNU_VTINHERIT: u32 = 41;
-/// GNU C++ vtable member usage
-pub const R_RISCV_GNU_VTENTRY: u32 = 42;
```
m4b: why were these deleted and also constant changed?

src/elf/constants_relocation.rs
```diff
@@ -872,38 +872,46 @@ pub const R_RISCV_SUB16: u32 = 38;
...
-/// Absolute address (CI-Type)
-pub const R_RISCV_RVC_LUI: u32 = 46;
-/// GP-relative reference (I-Type)
-pub const R_RISCV_GPREL_I: u32 = 47;
-/// GP-relative reference (S-Type)
-pub const R_RISCV_GPREL_S: u32 = 48;
-/// TP-relative TLS LE load (I-Type)
-pub const R_RISCV_TPREL_I: u32 = 49;
-/// TP-relative TLS LE store (S-Type)
-pub const R_RISCV_TPREL_S: u32 = 50;
```
m4b: delete why were these all deleted?

Jvlegod:
> so just fyi removing the pub constants is technically a breaking change and it would be pushed back a bit as I like to roll up breaking changes. Alternatively can leave them in and it won’t be and will be in a minor release

Here's a suggestion: I'll dedicate a separate PR to the destructive changes and address it when you feel the time is right. In this PR, I'll be making changes to the non-destructive parts.

What do you think?😎

Jvlegod:
I have removed the parts that caused the destructive changes.

m4b (approved):
awesome thank you!

m4b:
NB: backwards compat

NB: released in 0.10.5

PR #509, opened 2026-01-02
feat: install name tool style macho writing

wolfv:
In our application (rattler-build) we rely on `goblin` to read rpaths and linked libraries, but still fall back to `install_name_tool` to actually rewrite shared libraries and binaries.

I've tried to recreate the install name tool functionality identically in this PR. There is a Python script that executes the example install_name_tool (written in Rust, using Goblin) and compares the output with the proper install name tool from apple doing random modifications to binaries found on the system. The goal is to achieve bit-by-bit identical results.

I am opening this PR "early" to see if there is appetite for such functionality upstream in goblin. 

Also disclaimer: most of the code was written by Claude. I am happy to spend more time to refactor things to your liking if this PR has chances to get merged!

### DIFF
--- a/src/mach/mod.rs
+++ b/src/mach/mod.rs
@@ -398,10 +398,12 @@ pub fn peek_bytes(bytes: &[u8; 16]) -> error::Result<crate::Hint> {
                 }
             }
             fat::FAT_MAGIC | fat::FAT_MAGIC_64 => {
-                // should probably verify this is always Big Endian...
                 let narchitectures = bytes.pread_with::<u32>(4, BE)? as usize;
                 Ok(crate::Hint::MachFat(narchitectures))
             }
+            fat::FAT_CIGAM | fat::FAT_CIGAM_64 => Err(error::Error::Malformed(
+                "Fat binary is byte-swapped (little-endian); only big-endian fat binaries are supported".into(),
+            )),
             _ => Ok(crate::Hint::Unknown(bytes.pread::<u64>(0)?)),
         }
     }
@@ -603,6 +605,9 @@ impl<'a> Mach<'a> {
                 let multi = MultiArch::new(bytes)?;
                 Ok(Mach::Fat(multi))
             }
+            fat::FAT_CIGAM | fat::FAT_CIGAM_64 => Err(error::Error::Malformed(
+                "Fat binary is byte-swapped (little-endian); only big-endian fat binaries are supported".into(),
+            )),
             // we might be a regular binary
             _ => {
                 let binary = MachO::parse_impl(bytes, 0, lossy)?;


src/mach/fat.rs
```diff
@@ -13,6 +13,9 @@ use scroll::{Pread, Pwrite, SizeWith};
 
 pub const FAT_MAGIC: u32 = 0xcafe_babe;
 pub const FAT_CIGAM: u32 = 0xbeba_feca;
+/// 64-bit fat magic number (for archives with 64-bit offsets/sizes)
+pub const FAT_MAGIC_64: u32 = 0xcafe_babf;
+pub const FAT_CIGAM_64: u32 = 0xbfba_feca;
```
m4b: these are good changes, though I don't think I've ever seen this magic in the wild, though I haven't poked at macho binaries in a while

src/mach/fat.rs
```diff
@@ -130,3 +133,101 @@ impl FatArch {
...
+            Some(slice) => slice,
+            None => {
+                log::warn!("invalid `FatArch64` offset");
+                &[]
+            }
+        }
+    }
+
+    /// Returns the cpu type
+    pub fn cputype(&self) -> CpuType {
```
m4b: I was surprised to see this "Java beans" style getter, especially as the field is `pub` and I consider it an anti-pattern in rust. it's also odd this uses a type alias CpuType as return type, but the field is not declared as such; however, I see the FatArch struct is defined and declared as such :sweat_smile: probably some very old `pub` code from ancient times in goblin.

I was torn on whether we should be consistent or not carry the pattern forward, but probably since most of the struct have some kind of bitfiddling it made sense to make them all functions, even ones that returned the raw value. Anyway, this code hasn't been touched in a long time :D

src/mach/fat.rs
```diff
@@ -130,3 +133,103 @@ impl FatArch {
...
+    /// Where in the fat binary it starts (64-bit)
+    pub offset: u64,
+    /// How big the binary is (64-bit)
+    pub size: u64,
+    pub align: u32,
+    /// Reserved for future use
+    pub reserved: u32,
+}
+
+pub const SIZEOF_FAT_ARCH_64: usize = 32;
```
m4b: ditto here, I'd just call `size_of`

src/mach/fat.rs
```diff
@@ -130,3 +133,103 @@ impl FatArch {
         Ok(arch)
     }
 }
+
+#[repr(C)]
+#[derive(Clone, Copy, Default, Pread, Pwrite, SizeWith)]
+/// 64-bit version of `FatArch` for fat binaries with large offsets/sizes
+/// Uses u64 for offset and size fields
+pub struct FatArch64 {
```
m4b: so in mod.rs we implement a FatArhcIterator, and SingleArchIterator that returns these FatArch64's; since this patch doesn't add those features, I don't see where this would be used by a client? 

Again I've never seen this struct in the wild before, can you upload a binary that has it?

m4b (commented):
altogether this is much more palatable; I'd like to inspect a binary with the fatarch64 struct, e.g., running `bingrep` on it should fail iiuc, so if you could upload one to the conversation here, i'd appreciate it

m4b:
tool looks great! in terms of names I think goblin-patch or something like that is a little better, but if it meants to add more functionality something like ext or etc makes sense. There's also metagoblin that I maintain (which I've just remembered I have to update), but that is more like adding extra metadata to goblin that goblin itself shouldn't host/provide. :shrug: neat project of yours though!

wolfv:
@m4b are you fine with the changes now, or should I remove the code for fat arch? I am fine either way.

PR #517, opened 2026-01-31
PE: Add support for ARM64 unwind info

kkent030315:
addresses #512 

This is draft for now until plenty of more unwind code covered by tests (and of course for malformed ones as well).

### DIFF
--- a/src/pe/exception.rs
+++ b/src/pe/exception.rs
@@ -1160,7 +1160,7 @@ impl Arm64RuntimeFunction {
 
     /// Returns `true` if this unwind data is packed, `false` otherwise.
     pub const fn is_packed(&self) -> bool {
-        self.flag() != 0
+        self.flag() != ARM64_PDATA_REF_TO_FULL_XDATA
     }
 
     /// Returns an RVA of the full unwind data.
--- a/src/pe/exception.rs
+++ b/src/pe/exception.rs
@@ -1370,16 +1370,16 @@ pub struct Arm64UnwindInfo<'a> {
 
 impl<'a> Arm64UnwindInfo<'a> {
     fn checked_mul4(words: u32) -> error::Result<usize> {
-        let bytes = words.checked_mul(4).ok_or_else(|| {
-            error::Error::Malformed("arm64 unwind info size overflow".to_string())
-        })?;
+        let bytes = words
+            .checked_mul(4)
+            .ok_or_else(|| error::Error::Malformed("arm64 unwind info size overflow".into()))?;
         usize::try_from(bytes)
-            .map_err(|_| error::Error::Malformed("arm64 unwind info size overflow".to_string()))
+            .map_err(|_| error::Error::Malformed("arm64 unwind info size overflow".into()))
     }
 
     fn checked_add(a: usize, b: usize) -> error::Result<usize> {
         a.checked_add(b)
-            .ok_or_else(|| error::Error::Malformed("arm64 unwind info size overflow".to_string()))
+            .ok_or_else(|| error::Error::Malformed("arm64 unwind info size overflow".into()))
     }
 
     /// Precomputes the size of this xdata. This does _not_ count for handler-specific bytes after handler RVA.
--- a/src/pe/exception.rs
+++ b/src/pe/exception.rs
@@ -879,6 +879,18 @@ impl<'a> ExceptionData<'a> {
         self.get_function_by_offset(self.offset + index * RUNTIME_FUNCTION_SIZE)
     }
 
+    /// Returns an iterator over ARM64 runtime function entries.
+    pub fn functions_arm64(&self) -> Arm64RuntimeFunctionIterator<'a> {
+        Arm64RuntimeFunctionIterator {
+            data: &self.bytes[self.offset..self.offset + self.size],
+        }
+    }
+
+    /// Returns an ARM64 runtime function at the given index.
+    pub fn get_function_arm64(&self, index: usize) -> error::Result<Arm64RuntimeFunction> {
+        self.get_function_by_offset_arm64(self.offset + index * size_of::<Arm64RuntimeFunction>())
+    }
+
     /// Performs a binary search to find a function entry covering the given RVA relative to the
     /// image.
     pub fn find_function(&self, rva: u32) -> error::Result<Option<RuntimeFunction>> {
@@ -926,6 +938,15 @@ impl<'a> ExceptionData<'a> {
         self.get_unwind_info_with_opts(function, sections, &options::ParseOptions::default())
     }
 
+    /// Resolves unwind information for the given ARM64 function entry.
+    pub fn get_unwind_info_arm64(
+        &self,
+        function: Arm64RuntimeFunction,
+        sections: &[section_table::SectionTable],
+    ) -> Option<error::Result<Arm64UnwindInfo<'a>>> {
+        self.get_unwind_info_arm64_with_opts(function, sections, &options::ParseOptions::default())
+    }
+
     /// Resolves unwind information for the given function entry.
     pub fn get_unwind_info_with_opts(
         &self,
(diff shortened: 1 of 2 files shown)


src/pe/exception.rs
```diff
@@ -879,6 +879,18 @@ impl<'a> ExceptionData<'a> {
         self.get_function_by_offset(self.offset + index * RUNTIME_FUNCTION_SIZE)
     }
 
+    /// Returns an iterator over ARM64 runtime function entries.
+    pub fn functions_arm64(&self) -> Arm64RuntimeFunctionIterator<'a> {
```
m4b, line 919: it might be slightly more idiomatic to write `iter_functions_arm64` ?

I would be tempted to rename this to `arm64_functions` ? though the downside is it starts with less interesting information (arm64() as opposed to (functions) 🤔

src/pe/exception.rs
```diff
@@ -926,6 +938,15 @@ impl<'a> ExceptionData<'a> {
         self.get_unwind_info_with_opts(function, sections, &options::ParseOptions::default())
     }
 
+    /// Resolves unwind information for the given ARM64 function entry.
+    pub fn get_unwind_info_arm64(
+        &self,
+        function: Arm64RuntimeFunction,
+        sections: &[section_table::SectionTable],
+    ) -> Option<error::Result<Arm64UnwindInfo<'a>>> {
```
m4b, line 982: Might be worth checking what other parts of goblin return in cases like this, `Option<Result>` or  `Result<Option>` is probably better? Also, it might even be worth considering if the error is actionable or useful at all (e.g., effectively whether it's an OOB), to just `ok()` it into an Option and drop any error message info (again if it's not useful or actionable for the user)

I say this believing that more information the better, but if it's just satisfying the type system, might be better to coalesce the return type into a single one?

src/pe/exception.rs
```diff
@@ -1003,6 +1055,1013 @@ impl<'a> IntoIterator for &'_ ExceptionData<'a> {
...
+pub struct Arm64UnwindHeader(pub u32);
+
+const _: () = assert!(size_of::<Arm64UnwindHeader>() == 4);
+
+impl Arm64UnwindHeader {
+    const fn mask(bits: u32) -> u32 {
+        (1u32 << bits) - 1
+    }
+
+    const fn bits(v: u32, shr: u32, bits: u32) -> u32 {
```
m4b, line 1258: I assume the `const` part guarantees this is inlined?

src/pe/exception.rs
```diff
@@ -1003,6 +1055,1013 @@ impl<'a> IntoIterator for &'_ ExceptionData<'a> {
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.offset >= self.bytes.len() {
+            return None;
+        }
+        Some(
+            self.bytes
+                .gread_with::<u32>(&mut self.offset, scroll::LE)
+                .map(Arm64EpilogScope)
+                .map_err(Into::into),
```
m4b, line 1377: nice; this pattern comes up so much it makes me want to write a `ScrollIterator<T>` that does all this for you.

src/pe/exception.rs
```diff
@@ -1003,6 +1055,1013 @@ impl<'a> IntoIterator for &'_ ExceptionData<'a> {
...
+#[derive(Clone, Debug)]
+pub struct Arm64UnwindInfo<'a> {
+    /// Unwind info header.
+    pub header: Arm64UnwindHeader,
+    /// Extension info if present.
+    pub extension: Option<Arm64UnwindExtension>,
+    /// Bytes of epilog scope.
+    epilog_scope_bytes: &'a [u8],
+    /// Bytes of unwind codes.
+    unwind_code_bytes: &'a [u8],
```
m4b, line 1411: nit: I like to put non-pub fields after pub ones but ymmv

src/pe/exception.rs
```diff
@@ -1003,6 +1055,1013 @@ impl<'a> IntoIterator for &'_ ExceptionData<'a> {
...
+    /// Returns `true` if "extension" bytes are in addition.
+    pub const fn has_extension(&self) -> bool {
+        self.epilog_count() == 0 && self.code_words() == 0
+    }
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct Arm64UnwindExtension(pub u32);
+
+const _: () = assert!(size_of::<Arm64UnwindExtension>() == 4);
```
m4b, line 1309: neat! I wanted this so badly back in early days of rust, in this library and in dryad specifically :)

m4b (commented):
this looks good to me, is there any blocking issue you can think of/why we shouldn't merge?

src/pe/exception.rs
kkent030315: Ideally, overall, we would do something like made up of wrapper enum, but it's obviously breaking change. `arm64_functions` is acceptable, however as you mentioned, it's less informative for developers, specifically when type `func...` in their IDE.

```rust
pub enum RuntimeFunction {
  X86(X86RuntimeFunction),
  Arm(ArmRuntimeFunction),
}
```

src/pe/exception.rs
kkent030315: Uh, I agree it's a bit unfortunate to introduce `Option<Result<T>`. The reason for that is, for binary parsing, "absent of the data" v. "data present, but parsing fail" can be informative. For example, if I want to make a binary visualizer, I don't really want to display the data as "absent" when it might be an error parsing the data.

This might only cost a bit for developers while being informative, so I'd rather keep it, since it's public API we can't make changes later (meaning I'd want to keep the "absent" v. "error" meta regardless of what errors are in).

Especially when it comes to unwind info, it's often more of more important for the developer to know whether the info is absent or parsing fails!

```rust
foo.get_unwind_info_arm64(...) // Option<Result<T>>
	.map(Result::ok) // Option<Option<T>>
	.flatten() // Option<T>

foo.get_unwind_info_arm64(...)
	.and_then(|r| r.ok())
```

src/pe/exception.rs
kkent030315: Unfortunately, it has no such guarantee. We can introduce `#[inline(always)]`.

src/pe/exception.rs
kkent030315: It... might, might be made better, by just doing `bytes.chunks_exact(size_of::<Foo>()).map(...)` so we can fold redundant Iterator impls. Here is the siginifcant example from my own mutable-aware PE/ELF parser that every parser is dereffered and has only single slice reference (so-called _View_ auto datastructures).

```rust
impl<'buf, T> PortableExecutable<'buf, T>
where
    T: BufRef,
{
    /// Returns an iterator over sections.
    pub fn sections(
        &self,
    ) -> error::Result<impl Iterator<Item = ImageSectionHeaderView<'_, &[u8]>> + '_> {
        // ...

        Ok(buf
            .chunks_exact(size_of::<ImageSectionHeader>())
            .map(ImageSectionHeaderView::new))
    }
}
```

```rust
impl<'buf, T> PortableExecutable<'buf, T>
where
    T: BufMut,
{
    /// Returns a mutable iterator over sections.
    pub fn sections_mut(
        &mut self,
    ) -> error::Result<impl Iterator<Item = ImageSectionHeaderView<'_, &mut [u8]>> + '_> {
        // ...

        Ok(buf
            .chunks_exact_mut(size_of::<ImageSectionHeader>())
            .map(ImageSectionHeaderView::new))
    }
}
```

m4b:
@kkent030315 did we need some more time on this, or is this ready to go?

kkent030315:
> @kkent030315 did we need some more time on this, or is this ready to go?

Yes! just pushed some minor changes/fixes from laptop now.

kkent030315:
This is ready to get the final review and merge I guess, sorry for the lag. @m4b also take good care of yourself🙏!

m4b (approved):
Ok this was along time coming, thanks for your patience @kkent030315 . This is a huge PR, and I've already reviewed once. I don't think I will have anything else to add materially speaking, as I don't know much about this space in PEs, etc., so I'll (again) defer to your expertise on PE stuff, so you can be the de facto maintainer of this section of goblin code :D

m4b:
NB: backwards compatible

m4b:
released 0.10.6

PR #519, opened 2026-02-28
elf: fix MIPS64 little-endian relocation parsing

### LINKED ISSUE #274: MIPS64 parse error: "type is too big (1236271128) for 137416"

messense:
## Summary

Fixes #274 — MIPS64 LE ELF binaries fail to parse with an out-of-bounds error because `r_sym` and `r_type` return garbage values from the non-standard MIPS64 relocation layout.

## Background

MIPS64 ELF uses a non-standard relocation info (`r_info`) layout different from standard ELF64:

| Standard ELF64 | MIPS64 |
|---|---|
| `r_sym` (32 bits) \| `r_type` (32 bits) | `r_sym` (32 bits) \| `r_ssym` (8 bits) \| `r_type3` (8 bits) \| `r_type2` (8 bits) \| `r_type` (8 bits) |

On little-endian MIPS64, when this struct is read as a single `u64`, the byte order causes the fields to be scrambled. For example, an `r_info` with `r_sym=0` and `r_type=R_MIPS_REL32` gets read as `0x0312000000000000`, causing:
- `r_sym` to return **51,511,296** (instead of 0)
- `r_type` to return **0** (instead of 3)

The bogus `r_sym` then causes the dynamic symbol table parsing to try to read millions of symbols, failing with an out-of-bounds error.

## Approach

Based on the maintainer's feedback from #382 — **pass down whether it's a MIPS binary at runtime** rather than using `#[cfg]` compile-time conditionals:

1. **`reloc64::mips64el_r_info()`** — Converts MIPS64 LE `r_info` to standard ELF64 format using the same byte transformation as [LLVM](https://github.com/llvm/llvm-project/blob/119bf57ab6de49a3e61b9200c917a6d30ac6f0ad/llvm/include/llvm/Object/ELFTypes.h#L435-L444) and the [`object` crate](https://github.com/gimli-rs/object/blob/832f5277a0774de18f84e7c27401f974e4b6a064/src/elf.rs#L1129-L1139)

2. **`Reloc::fixup_mips64el()`** — Applies the transformation after parsing a relocation entry

3. **`RelocSection` / `RelocIterator`** — Carry an `is_mips64el` flag and conditionally apply the fixup during `get()` and iteration

4. **`Elf::parse_with_opts()`** — Detects MIPS64 LE from the ELF header (`e_machine == EM_MIPS && 64-bit && little-endian`) and passes it through via `RelocSection::parse_inner()`

## Backward Compatibility

- `RelocSection::parse()` signature is unchanged (defaults `is_mips64el = false`)
- `RelocCtx` type alias is unchanged
- `r_sym()` / `r_type()` / `r_info()` functions are unchanged
- All existing tests pass

## Tests

Added 5 new tests:
- `test_mips64el_r_info` — validates the byte transformation with real data from #274
- `test_mips64el_r_info_with_sym` — validates transformation when r_sym is non-zero
- `test_mips64el_r_info_zero` — edge case: all-zero r_info
- `test_standard_r_sym_r_type_unchanged` — ensures no regression for non-MIPS
- `test_mips64el_reloc_section_parse` — integration test through the full `RelocSection` pipeline, verifying both broken (without fix) and correct (with fix) behavior

## References

- Issue: #274
- Previous PR: #382 (closed with feedback to avoid `#[cfg]`)
- [MIPS64 ELF ABI documentation](https://repo.oss.cipunited.com/archives/ftp.linux-mips.org/pub/linux/mips/doc/ABI/elf64-2.4.pdf)
- [LLVM implementation](https://github.com/llvm/llvm-project/blob/119bf57ab6de49a3e61b9200c917a6d30ac6f0ad/llvm/include/llvm/Object/ELFTypes.h#L435-L444)
- [`object` crate implementation](https://github.com/gimli-rs/object/blob/832f5277a0774de18f84e7c27401f974e4b6a064/src/elf.rs#L1129-L1139)
- [Binutils MIPS64 relocation structs](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=elfcpp/elfcpp_internal.h;h=ce3673fe87fa20d1a96e881712b344adc60e7d92;hb=HEAD#l183)

### DIFF
--- a/src/elf/reloc.rs
+++ b/src/elf/reloc.rs
@@ -266,6 +266,28 @@ pub mod reloc64 {
         (sym << 32) + typ
     }
 
+    /// Convert a MIPS64 little-endian `r_info` value to the standard ELF64 format.
+    ///
+    /// MIPS64 ELF uses a non-standard relocation info layout:
+    /// - `r_sym` (32 bits) | `r_ssym` (8 bits) | `r_type3` (8 bits) | `r_type2` (8 bits) | `r_type` (8 bits)
+    ///
+    /// On little-endian systems, when this struct is read as a single `u64`, the byte order
+    /// causes the fields to be scrambled compared to the standard `(sym << 32) | type` layout.
+    /// This function rearranges the bytes so that the standard [`r_sym`] and [`r_type`] functions
+    /// return correct values.
+    ///
+    /// See the [MIPS64 ELF ABI](https://web.archive.org/web/20231012215433/https://techpubs.jurassic.nl/manuals/hdwr/developer/Mpro_n32_ABI/sgi_html/sgidoc/books/Mpro_n32_ABI/sgi_html/ch06.html)
+    /// and [LLVM's implementation](https://github.com/llvm/llvm-project/blob/119bf57ab6de49a3e61b9200c917a6d30ac6f0ad/llvm/include/llvm/Object/ELFTypes.h#L435-L444)
+    /// for reference.
+    #[inline(always)]
+    pub fn mips64el_r_info(info: u64) -> u64 {
+        (info << 32)
+            | ((info >> 8) & 0xff000000)
+            | ((info >> 24) & 0x00ff0000)
+            | ((info >> 40) & 0x0000ff00)
+            | ((info >> 56) & 0x000000ff)
+    }
+
     elf_rela_std_impl!(u64, i64);
 }
 
@@ -297,6 +319,19 @@ if_alloc! {
         pub fn size(is_rela: bool, ctx: Ctx) -> usize {
             use scroll::ctx::SizeWith;
             Reloc::size_with(&(is_rela, ctx))
+        }
+
+        /// Fix up `r_sym` and `r_type` for MIPS64 little-endian binaries.
+        ///
+        /// MIPS64 ELF uses a non-standard relocation info layout that causes
+        /// `r_sym` and `r_type` to be incorrectly extracted on little-endian systems.
+        /// This method reconstructs the original `r_info`, applies the MIPS64 LE
+        /// byte transformation, and re-extracts the correct values.
+        fn fixup_mips64el(&mut self) {
+            let r_info = ((self.r_sym as u64) << 32) | (self.r_type as u64);
+            let fixed = reloc64::mips64el_r_info(r_info);
+            self.r_sym = reloc64::r_sym(fixed) as usize;
+            self.r_type = reloc64::r_type(fixed);
         }
     }
 
@@ -394,6 +429,7 @@ if_alloc! {
         ctx: RelocCtx,
         start: usize,
         end: usize,
+        is_mips64el: bool,
     }
 
     impl<'a> fmt::Debug for RelocSection<'a> {
@@ -411,7 +447,23 @@ if_alloc! {
     impl<'a> RelocSection<'a> {
         #[cfg(feature = "endian_fd")]
         /// Parse a REL or RELA section of size `filesz` from `offset`.
+        ///
+        /// **Note:** This method does not apply the MIPS64 little-endian relocation
+        /// fixup. If you are parsing a MIPS64 LE binary, use [`Elf::parse`] or
+        /// [`Elf::parse_with_opts`] instead, which automatically detect MIPS64 LE
+        /// and apply the necessary `r_info` byte transformation.
         pub fn parse(bytes: &'a [u8], offset: usize, filesz: usize, is_rela: bool, ctx: Ctx) -> crate::error::Result<RelocSection<'a>> {
+            Self::parse_inner(bytes, offset, filesz, is_rela, ctx, false)
+        }
+
+        #[cfg(feature = "endian_fd")]
+        /// Parse a REL or RELA section of size `filesz` from `offset`, with MIPS64
+        /// little-endian relocation info handling.
+        ///
+        /// When `is_mips64el` is `true`, the MIPS64 little-endian byte transformation
+        /// is applied to the `r_info` field of each relocation entry, which corrects
+        /// the `r_sym` and `r_type` extraction for MIPS64 LE binaries.
+        pub(crate) fn parse_inner(bytes: &'a [u8], offset: usize, filesz: usize, is_rela: bool, ctx: Ctx, is_mips64el: bool) -> crate::error::Result<RelocSection<'a>> {
             // TODO: better error message when too large (see symtab implementation)
             let bytes = if filesz != 0 {
                 bytes.pread_with::<&'a [u8]>(offset, filesz)?
@@ -420,11 +472,12 @@ if_alloc! {
             };
 
             Ok(RelocSection {
-                bytes: bytes,
+                bytes,
                 count: filesz / Reloc::size(is_rela, ctx),
                 ctx: (is_rela, ctx),
                 start: offset,
                 end: offset + filesz,
+                is_mips64el,
             })
         }
 
@@ -434,7 +487,11 @@ if_alloc! {
             if index >= self.count {
                 None
             } else {
-                Some(self.bytes.pread_with(index * Reloc::size_with(&self.ctx), self.ctx).unwrap())
+                let mut reloc: Reloc = self.bytes.pread_with(index * Reloc::size_with(&self.ctx), self.ctx).unwrap();
+                if self.is_mips64el {
+                    reloc.fixup_mips64el();
+                }
+                Some(reloc)
             }
         }
 
@@ -473,6 +530,7 @@ if_alloc! {
                 index: 0,
                 count: self.count,
                 ctx: self.ctx,
+                is_mips64el: self.is_mips64el,
             }
         }
     }
@@ -483,6 +541,7 @@ if_alloc! {
         index: usize,
         count: usize,
         ctx: RelocCtx,
+        is_mips64el: bool,
     }
 
     impl<'a> fmt::Debug for RelocIterator<'a> {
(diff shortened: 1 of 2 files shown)


src/elf/reloc.rs
```diff
@@ -434,7 +487,11 @@ if_alloc! {
             if index >= self.count {
                 None
             } else {
-                Some(self.bytes.pread_with(index * Reloc::size_with(&self.ctx), self.ctx).unwrap())
+                let mut reloc: Reloc = self.bytes.pread_with(index * Reloc::size_with(&self.ctx), self.ctx).unwrap();
+                if self.is_mips64el {
+                    reloc.fixup_mips64el();
+                }
```
m4b: I think the correct thing to do is to turn `type RelocCtx = (bool, Ctx)` to:

```rust
struct RelocCtx {
  is_rela: bool,
  is_mips64el: bool,
  ctx: Ctx,
}
```

And then this diff disappears and the `fixup_mips64el()` call occurs inside the Ctx implementation.

I think this is also a non-breaking change, since the type alias for RelocCtx is not `pub`.

src/elf/reloc.rs
```diff
@@ -473,6 +530,7 @@ if_alloc! {
                 index: 0,
                 count: self.count,
                 ctx: self.ctx,
+                is_mips64el: self.is_mips64el,
```
m4b: e.g., this disappears

src/elf/reloc.rs
```diff
@@ -506,7 +565,11 @@ if_alloc! {
                 None
             } else {
                 self.index += 1;
-                Some(self.bytes.gread_with(&mut self.offset, self.ctx).unwrap())
+                let mut reloc: Reloc = self.bytes.gread_with(&mut self.offset, self.ctx).unwrap();
+                if self.is_mips64el {
+                    reloc.fixup_mips64el();
+                }
+                Some(reloc)
```
m4b: so will this

src/elf/reloc.rs
```diff
@@ -411,7 +447,23 @@ if_alloc! {
     impl<'a> RelocSection<'a> {
         #[cfg(feature = "endian_fd")]
         /// Parse a REL or RELA section of size `filesz` from `offset`.
+        ///
+        /// **Note:** This method does not apply the MIPS64 little-endian relocation
+        /// fixup. If you are parsing a MIPS64 LE binary, use [`Elf::parse`] or
+        /// [`Elf::parse_with_opts`] instead, which automatically detect MIPS64 LE
+        /// and apply the necessary `r_info` byte transformation.
         pub fn parse(bytes: &'a [u8], offset: usize, filesz: usize, is_rela: bool, ctx: Ctx) -> crate::error::Result<RelocSection<'a>> {
```
m4b, line 468: if we want to make this a breaking change I think the better api here is to have caller construct the reloc ctx, so we'd make `RelocCtx` public and do something like:

```rust
pub fn parse(bytes: &'a [u8], offset: usize, filesz: usize, ctx: RelocCtx) -> crate::error::Result<RelocSection<'a>> 
```

I don't think we should do this though in this patch, if this change isn't already breaking (which it doesn't appear to be, even with suggestion about doing the mips logic in the Ctx)

src/elf/reloc.rs
m4b: e.g., you'd do your fixup right here

m4b (changes requested):
First of all, thank you for your patience; I was extremely sick last week, and was unable to review.

Anyway, thank you for this PR, this is great! I like that it's backwards compatible, though it is tempting to make some of the backwards imcompatible changes for a better api surface.

The main issue is whether my suggestions are backwards compatible? At first I thought they were not, but then I was surprised to see that RelocCtx was a `non-pub` typealias, which iiuc, would mean the trait implementation (which I thought were always public?) depends on a private type, which I assume means this is not a breaking change to change the `TryFromCtx` impl? E.g., no one can currently write down RelocCtx, or construct it, so I can't see how it would be a breaking change, but there may be some subtlety here I'm not seeing right now.

Assuming this is correct, I believe we should do the changes I suggest (putting the mips boolean into the RelocCtx, and doing the fixup in the `TryFromCtx` impl)

If this is *not true* (though semantically I don't' quite understand how the trait impl could be public, somehow breaking for users, but also depend on a private type you can't even name due to visibility, e.g., in swift this is a compile time error), then we should proceed with yours and we can do the proper Ctx based implementation as a breaking change.

Thanks again for this great PR and pushing this over the finish line!

m4b:
Oh I should also mention the only other "oddity" here is that `Reloc::size(is_rela, Ctx)` is `pub`, but thankfully size doesn't depend on the mips property so we can still keep as is for now unless want to do a breaking change later as suggested with RelocCtx for the `parse` method and would also update `size`. Anyway, not a big deal.

messense:
Thanks for the review, hope you are getting better!

philipc:
While `RelocCtx` isn't public, it is an alias for a tuple of public types, so that tuple can still manage to leak into the public API, such as in these trait implementations: https://docs.rs/goblin/0.10.5/goblin/elf/reloc/struct.Reloc.html#impl-SizeWith%3C(bool,+Ctx)%3E-for-Reloc. Changing `RelocCtx` to a private struct means that `SizeWith` implementation is removed from the public API. In practice I doubt this matters, just depends how strict you want to be for breaking changes.

src/elf/reloc.rs
```diff
@@ -296,57 +318,86 @@ if_alloc! {
...
+            }
+            Ok((reloc, size))
         }
     }
 
     impl ctx::TryIntoCtx<RelocCtx> for Reloc {
         type Error = crate::error::Error;
         /// Writes the relocation into `bytes`
-        fn try_into_ctx(self, bytes: &mut [u8], (is_rela, Ctx {container, le}): RelocCtx) -> result::Result<usize, Self::Error> {
+        fn try_into_ctx(self, bytes: &mut [u8], rctx: RelocCtx) -> result::Result<usize, Self::Error> {
```
philipc: Probably worth keeping `TryIntoCtx` and `IntoCtx` consistent with each other.

And if we're going to change `TryIntoCtx` to use the new `RelocCtx` type, then we may as well fix it to handle `rctx.is_mips64el` while we're at it.

src/elf/reloc.rs
```diff
@@ -404,9 +420,9 @@ if_alloc! {
 
     impl ctx::IntoCtx<(bool, Ctx)> for Reloc {
```
m4b: can you make this take RelocCtx too, for consistency?

src/elf/reloc.rs
```diff
@@ -335,53 +335,69 @@ if_alloc! {
...
-    type RelocCtx = (bool, Ctx);
+    #[derive(Clone, Copy, Debug, Default)]
+    struct RelocCtx {
+        is_rela: bool,
+        is_mips64el: bool,
+        ctx: Ctx,
+    }
+
+    impl RelocCtx {
+        fn new(is_rela: bool, ctx: Ctx) -> Self {
```
m4b: I would just delete this, defaulting the crucial is_mips param to false feels like a footgun for a `::new` function.

src/elf/reloc.rs
```diff
@@ -318,7 +318,7 @@ if_alloc! {
     impl Reloc {
         pub fn size(is_rela: bool, ctx: Ctx) -> usize {
             use scroll::ctx::SizeWith;
-            Reloc::size_with(&(is_rela, ctx))
+            Reloc::size_with(&RelocCtx::new(is_rela, ctx))
```
m4b: I would personally just get rid of the ::new constructor and simply do:

```rust
RelocCtx { is_rela, ctx, is_mips64el: false }
```

so it's clear it's false here.

m4b (commented):
this is great few minor changes then ready I think!  I'm on the fence whether we take these latest changes and make it a breaking change, or just add it in minor release (e.g., moving from tuple to private struct).
As philipc points out, tuples are not nominal, but structural, in rust, so even though the alias is private, one can still "name" (e.g., construct) the type that is implemented, since it uses two public types `(bool, Ctx)`.

 in practice I highly doubt anyone is relying on the SizeWith and TryFromCtx impls being visible or what the ctx for the impl is, but it is technically a breaking change to change Reloc's ctx from a tuple to a struct.

One thing we could do is take your first initial commit, merge that, do a minor release, then take your 2nd commit and make that apart of a breaking release if want to be really particular.

A nice compromise I think I will end up doing is just do a minor release with the RelocCtx change (the requested changes), and if it breaks anyone, I'll yank, and release with the first change, and put the RelocCtx as a breaking change.

m4b (approved):
awesome, thanks for your patience and this PR @messense !

m4b:
NB: backwards compatible

m4b:
released 0.10.6

PR #520, opened 2026-03-19
Inconsistent naming conventions

crzysdrs:
A few of the string representations of the X86_64 relocations do not match the other strings with their `R_` prefix.

Some other architectures include `R_` in their strings (mips, riscv, ppc). I think those prefixes should probably be removed but at least they are consistent with other entries in their architecture but that's out of scope for this pull request.


m4b (approved):
Thank you for fixes like this!

m4b:
NB: backwards compatible

m4b:
released 0.10.6

PR #521, opened 2026-03-22
pe.options: added parse_resources option

prettyroseslover:
Sometimes it's great to have a choice to skip resource parsing. Therefore, I added `ParseOptions.parse_resources`. Also, I decided to add a couple of missing `with_` methods to `ParseOptions`.

### DIFF
--- a/src/pe/mod.rs
+++ b/src/pe/mod.rs
@@ -364,19 +364,18 @@ impl<'a> PE<'a> {
                 0
             };
 
-            if opts.parse_resources {
-                if let Some(&resource_table) = optional_header.data_directories.get_resource_table()
-                {
-                    let data = resource::ResourceData::parse_with_opts(
-                        bytes,
-                        resource_table,
-                        &sections,
-                        file_alignment,
-                        opts,
-                    )?;
-                    resource_data = Some(data);
-                    debug!("resource_data data: {:#?}", data.version_info);
-                }
+            if opts.parse_resources
+                && let Some(&resource_table) = optional_header.data_directories.get_resource_table()
+            {
+                let data = resource::ResourceData::parse_with_opts(
+                    bytes,
+                    resource_table,
+                    &sections,
+                    file_alignment,
+                    opts,
+                )?;
+                resource_data = Some(data);
+                debug!("resource_data data: {:#?}", data.version_info);
             }
 
             authenticode_excluded_sections = Some(authenticode::ExcludedSections::new(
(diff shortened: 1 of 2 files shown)


kkent030315:
Hi, thank you for the PR! Besides do you have any issues with the parser? Or, I guess is with some performance matters?😃

prettyroseslover:
No, I have encountered no issues with the parser, it's more for the performance + malware samples which may have some funky stuff with the resources, but I still want to parse everything apart from that!

m4b:
apologies, this has conflicts now, if you could rebase, I"ll re-review, assuming @kkent030315 thinks its good direction

kkent030315:
Yes I think so. In other words there's no much anything else we could do without breaking changes. This looks the right direction for me!

src/pe/mod.rs
```diff
@@ -364,16 +364,19 @@ impl<'a> PE<'a> {
...
-                    bytes,
-                    resource_table,
-                    &sections,
-                    file_alignment,
-                    opts,
-                )?;
-                resource_data = Some(data);
-                debug!("resource_data data: {:#?}", data.version_info);
+            if opts.parse_resources {
+                if let Some(&resource_table) = optional_header.data_directories.get_resource_table()
```
kkent030315, line 368: I believe we could do this to merge into one:
```suggestion
            if opts.parse_resources && let Some(&resource_table) = optional_header.data_directories.get_resource_table()
```

kkent030315 (approved):
Other than one minor LGTM. Thank you so much for the PR!

src/pe/mod.rs
prettyroseslover: Thanks for you feedback. However, I guess it won't be possible (check [CI tests](https://github.com/m4b/goblin/actions/runs/26086461790/job/76700713561) for this), as MSRV for goblin is 1.85 and the following statement was stabilized only in 1.88 I believe.

I did resolve the conflicts tho!

kkent030315: oh, sorry, that was on me. yes, you can keep this until we bump msrv at some point :)

prettyroseslover: I guess this conversation can be resolved and PR can be merged?

m4b (approved):
thank you!

m4b:
thanks for your patience!

m4b:
NB: backwards compatible

m4b:
released in 0.10.6

Issue #506, opened 2025-12-18
PE created by MSYS2+Mingw-w64 GCC can't be parsed


loewenheim:
At Sentry, we've received a [report](https://github.com/getsentry/sentry/issues/104738) that PE files containing DWARF debug info produced by MSYS2+Mingw-w64 GCC can't be processed. The `.debug` files in https://packages.msys2.org/packages/mingw-w64-x86_64-qt6-base-debug can serve as an example:
```
❯ ~/code/symbolic/target/debug/object_debug mingw-w64-clang-x86_64-qt6-base-debug-6.10.1-2-any.pkg/clang64/share/qt6/bin/cmake_automoc_parser.debug
Inspecting mingw-w64-clang-x86_64-qt6-base-debug-6.10.1-2-any.pkg/clang64/share/qt6/bin/cmake_automoc_parser.debug
File format: pe
Objects:
 - Error: invalid PE file
   caused by bad input invalid utf8 (22040925)
   caused by bad input invalid utf8 (22040925)
```
(The `object_debug` binary from the `symbolic` repo uses `goblin` to parse PE files).

I'm using `goblin 0.10.4` in this case.

This issue might be a consequence of https://github.com/m4b/goblin/issues/487.

kkent030315:
Thank you for the report. I'm interested in fixing this, let me take a deeper look into the problem.

PR #522, opened 2026-03-27
feat(pe): add `parse_imports` option to `ParseOptions`

supervacuus:
Parsing `PE` import tables is expensive: for each DLL, the parser does a full pass over the import lookup table (resolving every symbol `RVA` via a linear scan over sections), a second pass over the import address table, and then a third reshaping pass in `Import::parse`. For binaries with large or malformed import tables, this adds up fast, especially when the caller only needs headers, debug info, or exports.

This PR adds a `parse_imports` flag to `ParseOptions` (default: `true`) with a `with_parse_imports` builder method, allowing callers to skip import table parsing entirely. When set to `false`, `import_data`, `imports`, and `libraries` remain empty while everything else is parsed as usual.

This is a follow-up to #506 (#508).

### DIFF
--- a/src/pe/mod.rs
+++ b/src/pe/mod.rs
@@ -194,7 +194,10 @@ impl<'a> PE<'a> {
                 }
             }
             debug!("exports: {:#?}", exports);
-            if let Some(&import_table) = optional_header.data_directories.get_import_table() {
+            if let (true, Some(&import_table)) = (
+                opts.parse_imports,
+                optional_header.data_directories.get_import_table(),
+            ) {
                 let id = if is_64 {
                     import::ImportData::parse_with_opts::<u64>(
                         bytes,
--- a/src/pe/import.rs
+++ b/src/pe/import.rs
@@ -593,4 +593,34 @@ mod tests {
         assert_eq!(binary.imports[0].rva, 0x21B8);
         assert_eq!(binary.imports[0].size, 8);
     }
+
+    #[test]
+    fn skip_import_parsing_when_disabled() {
+        let opts = crate::pe::options::ParseOptions::default().with_parse_imports(false);
+        let binary = crate::pe::PE::parse_with_opts(WELL_FORMED_IMPORT, &opts)
+            .expect("Unable to parse binary");
+        assert!(binary.import_data.is_none());
+        assert!(binary.imports.is_empty());
+        assert!(binary.libraries.is_empty());
+        assert!(binary.header.optional_header.is_some());
+        assert!(!binary.sections.is_empty());
+    }
+
+    #[test]
+    fn skip_imports_preserves_debug_data() {
+        // Use a binary with debug directories to verify that skipping imports
+        // does not affect debug data parsing.
+        const MSVC_BIN: &[u8] =
+            include_bytes!("../../tests/bins/pe/debug_directories-msvc.exe.bin");
+        let opts = crate::pe::options::ParseOptions::default().with_parse_imports(false);
+        let binary =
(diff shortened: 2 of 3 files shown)


supervacuus:
Just wanted to ping if this kind of change is of interest.

I have also now added a new PR (https://github.com/m4b/goblin/pull/527) that fixes the underlying issue. It might still make sense to be able to turn off import parsing, but if https://github.com/m4b/goblin/pull/527 is accepted, I could understand if you didn't want to add another parser option. The resulting runtimes are now essentially the same for either PR.

m4b:
I'm ok with this i think, @kkent030315 is this something you think is fine?

src/pe/mod.rs
```diff
@@ -194,38 +194,40 @@ impl<'a> PE<'a> {
...
-                        opts,
-                    )?
-                };
-                debug!("import data {:#?}", id);
-                if is_64 {
-                    imports = import::Import::parse::<u64>(bytes, &id, &sections)?
-                } else {
-                    imports = import::Import::parse::<u32>(bytes, &id, &sections)?
+            if opts.parse_imports {
+                if let Some(&import_table) = optional_header.data_directories.get_import_table() {
```
kkent030315: can we merge it to one?
```suggestion
            if opts.parse_imports && let Some(&import_table) = optional_header.data_directories.get_import_table()
```

kkent030315 (approved):
LGTM. Thanks!

src/pe/mod.rs
supervacuus: `rust` is pinned at `1.85` in CI which [complains about the let chain](https://github.com/m4b/goblin/actions/runs/24664796028/job/72119566210?pr=522#step:6:19).

Otherwise fixed in: https://github.com/m4b/goblin/pull/522/commits/9f632aeff2ab2dea4a25b721c7ddf2b0cd6a8b9f

supervacuus: We can use

```rust
if let (true, Some(&import_table)) = (                                                                                                                              
    opts.parse_imports,                                                                                                                                             
    optional_header.data_directories.get_import_table(),                                                                                                            
) {                                                                                                                                                                 

}
```

if you prefer it to if-nesting. However, that means `get_import_table()` will be called unconditionally (which is not a costly invocation, but a change in behavior nonetheless).

Changed here: https://github.com/m4b/goblin/pull/522/commits/db951f2fbe540cd4bfbcedab5411e5d11b2e910c

supervacuus: Is the approach in https://github.com/m4b/goblin/commit/db951f2fbe540cd4bfbcedab5411e5d11b2e910c sufficient?

m4b: I think so

m4b:
I'm sorry @supervacuus there's a merge conflict with another pr that touched options, would you mind to rebase? I'll merge it asap as I want to cut a release tonight or tomorrow at latest. If don't hear I'll manually resolve it in editor though I'd prefer not

supervacuus:
> I'm sorry @supervacuus there's a merge conflict with another pr that touched options, would you mind to rebase? I'll merge it asap as I want to cut a release tonight or tomorrow at latest. If don't hear I'll manually resolve it in editor though I'd prefer not

I rebased and squashed my commits in https://github.com/m4b/goblin/pull/522/commits/0efa02fbb59f64d07cddd7382c5d62526f908598.

m4b (approved):
thank you!

m4b:
NB: backwards compatible

m4b:
released in 0.10.6

PR #524, opened 2026-04-01
PE: bound import descriptor loop by directory size

### LINKED ISSUE #523: PE: parser hangs while processing corkami's `manyimportsW7.exe`
Current version of goblin hangs on parsing [corkami's `manyimportsW7.exe` ](https://github.com/corkami/pocs/blob/master/PE/bin/manyimportsW7.exe), which is described as:
> W7-only binary use the `TLS AddressOfIndex` trick to clean its imports. On disk, the import table is full of bogus descriptors, which will be ignored on loading

Snippet to reproduce this behavior:
```rust
use std::fs;

use goblin::pe::options::ParseOptions;
use goblin::pe::options::ParseMode;
use goblin::pe;

fn main() {
    let path = "manyimportsW7.exe";
    let content = fs::read(path).unwrap();
    let mut parse_options = ParseOptions::default();
    parse_options.parse_mode = ParseMode::Permissive;

    let pe_file = pe::PE::parse_with_opts(&content, &parse_options).unwrap();
}
```
Other PE parsers, for example [`pedump`](https://github.com/zed-0xff/pedump/blob/1fb7442362491ba4d6ae67991f69acea80b2aace/lib/pedump.rb#L643) catch this trick:
```
[!] catched the 'imports terminator in TLS trick'
```

Taking into account, this sample is valid and somewhat popular with the community (as it is supported by other PE parsers, etc), it would be nice to catch fake imports usage in order not to load them all.

kkent030315:
Fixes #523

### DIFF
--- a/src/pe/import.rs
+++ b/src/pe/import.rs
@@ -452,8 +452,18 @@ impl<'a> ImportData<'a> {
         }
 
         debug!("import data offset {:#x}", offset);
+
+        let max_entries = if dd.size as usize >= SIZEOF_IMPORT_DIRECTORY_ENTRY {
+            dd.size as usize / SIZEOF_IMPORT_DIRECTORY_ENTRY - 1
+        } else {
+            0
+        };
         let mut import_data = Vec::new();
+        let mut entry_index = 0usize;
         loop {
+            if entry_index >= max_entries {
+                break;
+            }
             // Check bounds before reading to handle packed binaries where sections
             // may point to addresses beyond the file size
             if *offset + SIZEOF_IMPORT_DIRECTORY_ENTRY > bytes.len() {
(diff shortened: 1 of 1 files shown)


kkent030315:
To acautlly test this:

```bash
cargo run --example rdr -- "path/to/manyimportsW7.exe"
```

Tested locally.

m4b:
@kkent030315 friendly ping, is this something we want to merge, even as a temporary stopgap?

kkent030315:
> @kkent030315 friendly ping, is this something we want to merge, even as a temporary stopgap?

I think this one should be closed for now.

kkent030315:
Thanks @philipc for pointing this out :)

PR #525, opened 2026-04-07
Mach-O: Parse macOS version

CamJN:
Possible implementation for #514

### DIFF
--- a/src/mach/version.rs
+++ b/src/mach/version.rs
@@ -1,0 +1,132 @@
+/*-
+ * Copyright: see LICENSE file
+ */
+
+use crate::error::Error;
+use crate::mach::cputype::{CPU_TYPE_ARM64, CPU_TYPE_X86_64};
+use crate::mach::load_command::CommandVariant;
+use crate::mach::{Mach, MachO, SingleArch};
+use std::cmp::Ordering;
+use std::collections::VecDeque;
+use std::str::FromStr;
+use std::{env, fmt};
+
+#[derive(Eq, Debug)]
+pub struct Version {
+    pub major: u32,
+    pub minor: u32,
+    pub patch: u32,
+}
+
+impl fmt::Display for Version {
+    // This trait requires `fmt` with this exact signature.
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        // Write strictly the first element into the supplied output
+        // stream: `f`. Returns `fmt::Result` which indicates whether the
+        // operation succeeded or failed. Note that `write!` uses syntax which
+        // is very similar to `println!`.
+        write!(f, "{}.{}.{}", self.major, self.minor, self.patch)
+    }
+}
+
+impl Ord for Version {
+    fn cmp(&self, other: &Self) -> Ordering {
+        let mao = self.major.cmp(&other.major);
+        let mio = self.minor.cmp(&other.minor);
+        let pao = self.patch.cmp(&other.patch);
+        if mao == Ordering::Equal && mio == Ordering::Equal {
+            pao
+        } else if mao == Ordering::Equal {
+            mio
+        } else {
+            mao
+        }
+    }
+}
+
+impl PartialOrd for Version {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl PartialEq for Version {
+    fn eq(&self, other: &Self) -> bool {
+        self.major == other.major && self.minor == other.minor && self.patch == other.patch
+    }
+}
+
+impl FromStr for Version {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let mut parts = s
+            .trim()
+            .split('.')
+            .map(|p| p.parse::<u32>().unwrap())
+            .take(3)
+            .collect::<VecDeque<u32>>();
+
+        Ok(Self {
+            major: parts.pop_front().unwrap(),
+            minor: parts.pop_front().unwrap_or(0),
+            patch: parts.pop_front().unwrap_or(0),
+        })
+    }
+}
+
+impl From<u32> for Version {
+    fn from(packed: u32) -> Self {
+        // X.Y.Z is encoded in nibbles xxxx.yy.zz
+        // 12.6 = 0b0000_0000_0000_1100_0000_0110_0000_0000
+        let major = (packed & 0b1111_1111_1111_1111_0000_0000_0000_0000u32) >> 16;
+        let minor = (packed & 0b0000_0000_0000_0000_1111_1111_0000_0000u32) >> 8;
+        let patch = (packed & 0b0000_0000_0000_0000_0000_0000_1111_1111u32) >> 0;
+        Self {
+            major,
+            minor,
+            patch,
+        }
+    }
+}
+
+impl From<MachO<'_>> for Version {
+    fn from(b: MachO) -> Self {
+        let packed = b
+            .load_commands
+            .iter()
+            .find_map(|c| match c.command {
+                CommandVariant::VersionMinMacosx(v) => Some(v.version),
+                CommandVariant::BuildVersion(v) => Some(v.minos),
+                _ => None,
+            })
+            .unwrap();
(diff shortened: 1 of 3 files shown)


src/mach/version.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    impl From<Mach<'_>> for Version {
+        fn from(b: Mach) -> Self {
+            match b {
+                Mach::Binary(b) => Version::from(b),
+                Mach::Fat(f) => {
+                    match f
+                    .find(|r| {
+                        r.unwrap().cputype
```
m4b: can this unwrap panic (i.e., can a malformed or whatever binary cause this unwrap to panic?); if not please add documentation explaining why not

src/mach/version.rs
```diff
@@ -0,0 +1,138 @@
...
+                        == match env::var("CARGO_CFG_TARGET_ARCH").as_deref() {
+                            Ok("x86_64") => CPU_TYPE_X86_64,
+                            Ok("aarch64") => CPU_TYPE_ARM64,
+                            _ => panic!("unknown arch"),
```
m4b: this definitely seems like it can panic

src/mach/version.rs
```diff
@@ -0,0 +1,138 @@
...
+            Ok(Self {
+                major: parts.pop_front().unwrap(),
+                minor: parts.pop_front().unwrap_or(0),
+                patch: parts.pop_front().unwrap_or(0),
+            })
+        }
+    }
+
+    impl From<Mach<'_>> for Version {
+        fn from(b: Mach) -> Self {
```
m4b: so there are multiple unwraps in this function, which indicates to me the choice of `From` is not correct; `TryFrom` is probably a better choice, or if you need a parsing context, it would be more idiomatic in this crate to use scroll's `TryFromCtx`

src/mach/version.rs
```diff
@@ -0,0 +1,138 @@
...
+    impl PartialOrd for Version {
+        fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+            Some(self.cmp(other))
+        }
+    }
+
+    impl FromStr for Version {
+        type Err = Error;
+
+        fn from_str(s: &str) -> Result<Self, Self::Err> {
```
m4b, line 92: all these unwraps need to be map_err'd or documented why they cannot fail

src/mach/version.rs
```diff
@@ -0,0 +1,138 @@
...
+    fn from(b: MachO) -> Self {
+        let packed = b
+            .load_commands
+            .iter()
+            .find_map(|c| match c.command {
+                CommandVariant::VersionMinMacosx(v) => Some(v.version),
+                CommandVariant::BuildVersion(v) => Some(v.minos),
+                _ => None,
+            })
+            .unwrap();
```
m4b: ditto, unwrap again, needs documentation why it is infallible, or choice TryFrom/TryFromCtx, etc.

m4b (changes requested):
The unwraps need to be addressed, before can consider to merge. Additionally, it may be more idiomatic to have this be a `version` method on a `MachO`, and returns an `Optional<Version>`; I don't think we have anywhere else where we have a From method on MachO that returns some type, so it might be a little undiomatic in this crate to do that.

src/mach/version.rs
```diff
@@ -0,0 +1,148 @@
...
+    use std::collections::{VecDeque,HashMap};
+    use std::str::FromStr;
+    use std::fmt;
+}
+
+#[derive(Eq, Debug)]
+pub struct Version {
+    pub major: u32,
+    pub minor: u32,
+    pub patch: u32,
```
m4b, line 22: please document all of this, public things, especially new ones, should be doc'd

src/mach/version.rs
```diff
@@ -0,0 +1,148 @@
+/*-
+ * Copyright: see LICENSE file
+ */
+
+use crate::mach::MachO;
+use crate::mach::load_command::CommandVariant;
+
+if_std! {
+    use crate::error;
```
m4b, line 9: iirc, error is not under std; and I don't think cmp or str or fmt is either, they should be under core?

src/mach/version.rs
```diff
@@ -0,0 +1,148 @@
...
+    pub minor: u32,
+    pub patch: u32,
+}
+
+impl From<u32> for Version {
+    fn from(packed: u32) -> Self {
+        // X.Y.Z is encoded in nibbles xxxx.yy.zz
+        // 12.6 = 0b0000_0000_0000_1100_0000_0110_0000_0000
+        Self {
+            major: (packed & 0b1111_1111_1111_1111_0000_0000_0000_0000u32) >> 16,
```
m4b, line 30: nit: i think we use hex for other bitmasks in other parts of codebase, might want to be uniform with that

src/mach/version.rs
```diff
@@ -0,0 +1,148 @@
...
+                CommandVariant::VersionMinMacosx(v) => Some(v.version),
+                CommandVariant::BuildVersion(v) => Some(v.minos),
+                _ => None,
+            })
+            .map(Version::from)
+    }
+}
+
+impl PartialEq for Version {
+    fn eq(&self, other: &Self) -> bool {
```
m4b, line 51: you should be able to just `#[derive(PartialEq)`

src/mach/version.rs
```diff
@@ -0,0 +1,148 @@
...
+                mao
+            }
+        }
+    }
+
+    impl PartialOrd for Version {
+        fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+            Some(self.cmp(other))
+        }
+    }
```
m4b, line 87: ditto, you should be able to derive this (and it also doesn't require std)

src/mach/version.rs
```diff
@@ -0,0 +1,148 @@
...
+    impl FromStr for Version {
+        type Err = error::Error;
+
+        fn from_str(s: &str) -> Result<Self, Self::Err> {
+            let mut parts = s
+            .trim()
+            .split('.')
+            .map(|p| p.parse::<u32>().unwrap_or(0))
+            .take(3)
+            .collect::<VecDeque<u32>>();
```
m4b, line 98: this allocates a VecDequeu here just to pop the front; i think you can rewrite this to not allocate, and please do that. thank you for updating with fallible methods though.

src/mach/version.rs
```diff
@@ -0,0 +1,148 @@
...
+                Mach::Fat(f) => f.into_iter().map(|r| r.map(|s| match s {
+                    SingleArch::MachO(b) => b.version().ok_or_else(||error::Error::Malformed("Missing or corrupted version".to_string())),
+                    SingleArch::Archive(_) => Err(error::Error::Malformed("lib is an archive?".to_string())),
+                }).and_then(std::convert::identity)).collect(),
+            }
+        }
+    }
+
+    impl Mach<'_> {
+        pub fn versions(self) -> HashMap<CpuType, Version> {
```
m4b, line 127: needs docstring

In general, I don't necessarily like a method hanging off of Mach here in this file, as it's not local anymore to Mach. This also allocates a hashmap; I think it would be better to probably return an `impl Iterator<(CpuType, Version)>`; this way users can allocate if they want, etc.

src/mach/version.rs
```diff
@@ -0,0 +1,148 @@
+/*-
+ * Copyright: see LICENSE file
+ */
+
+use crate::mach::MachO;
+use crate::mach::load_command::CommandVariant;
+
+if_std! {
```
m4b, line 8: i think you can remove this entire block once you get rid of the allocations

m4b (changes requested):
almost there, just need to address some of the issues I flagged, thank you!

m4b:
@CamJN ping no pressure but if you get a chance would like to see feedback addressed and we can get this in!

CamJN:
I'm sorry, I've been absolutely slammed for months. I really picked a bad time to submit this originally. I do plan on fixing the things you asked for. I just need to have a few hours to do it right.

m4b:
@CamJN No worries and no pressure, was just doing PR triage on a lazy sunday :) whenever/if you have time, please do so at your own pace

PR #526, opened 2026-04-08
pe: wrap ResourceData parse in or_permissive_and_default (follow-up to #508)

CalebFenton:
Follow-up to #508. That PR added `or_permissive_and_default` inside `ResourceData::parse_with_opts` but the call site in `PE::parse()` still uses bare `?`:

```rust
// mod.rs line 367 — bare `?` means permissive mode doesn't help
let data = resource::ResourceData::parse_with_opts(
    bytes, resource_table, &sections, file_alignment, opts,
)?;
```

When `VersionInfo::parse` or `ManifestData::parse` fails inside that call (e.g. resource data RVAs that can't be mapped to file offsets), the error propagates through the bare `?` and kills the entire PE parse — even in `ParseMode::Permissive`.

This wraps it in `.map(Some).or_permissive_and_default(...)`, same as the debug_data, tls_data, load_config_data, and clr_data call sites nearby.

We hit this on packed PE malware (FSG, TELock) where the packer mangles the resource section but the rest of the PE is fine.

### DIFF
--- a/src/pe/mod.rs
+++ b/src/pe/mod.rs
@@ -40,7 +40,7 @@ use crate::pe::utils::pad;
 use crate::strtab;
 use options::ParseMode;
 
-use scroll::{Pwrite, ctx};
+use scroll::{ctx, Pwrite};
 
 use log::debug;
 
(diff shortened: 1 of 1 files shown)


m4b (approved):
Looks good, thanks for catching this!

m4b:
NB: backwards compatible

m4b:
released 0.10.6

PR #527, opened 2026-04-16
fix(pe): only map file-backed sections to offsets

supervacuus:
This is complementary to the `parse_imports` option change (https://github.com/m4b/goblin/pull/522). That change adds an escape hatch for callers who do not need import parsing. This PR otoh, solves the underlying issue that prompted this option (which could still make sense). The result is that, even with import parsing enabled in permissive mode, we get the same baseline runtime as if import parsing were disabled.

For on-disk PE parsing, sections with `SizeOfRawData == 0` are virtual-only. `goblin` currently still maps RVAs in such sections by falling back to `VirtualSize`. On affected files (in our case, `mingw`/`msys` debug companions), this causes unrelated bytes to be parsed as imports, and similarly affects other PE data directories.

This PR makes PE RVA resolution treat sections with `SizeOfRawData == 0` as not file-backed. Resource parsing is also made permissive-mode-friendly for the same class of virtual-only `.rsrc` directories, and the `malformed_resource_tree()` test is adjusted to assert parse failure rather than a specific panic path since the corrected mapping changes the failure point.

### DIFF
--- a/src/pe/utils.rs
+++ b/src/pe/utils.rs
@@ -51,9 +51,17 @@ fn section_read_size(section: &section_table::SectionTable, file_alignment: u32)
     // {
     //     readsize = min(readsize, (virtsize + 0xfff) & ~0xfff);
     // }
+    //
+    // Only bytes that are actually backed by the file may be mapped into offsets.
+    // Sections with `SizeOfRawData == 0` are virtual-only when parsing an on-disk PE,
+    // even if they have a non-zero `VirtualSize`.
+
+    let size_of_raw_data = section.size_of_raw_data as usize;
+    if size_of_raw_data == 0 {
+        return 0;
+    }
 
     let file_alignment = file_alignment as usize;
-    let size_of_raw_data = section.size_of_raw_data as usize;
     let virtual_size = section.virtual_size as usize;
     let read_size = {
         let read_size =
@@ -65,8 +73,6 @@ fn section_read_size(section: &section_table::SectionTable, file_alignment: u32)
 
     if virtual_size == 0 {
         read_size
-    } else if read_size == 0 {
-        virtual_size
     } else {
         cmp::min(read_size, round_size(virtual_size))
     }
@@ -261,4 +267,46 @@ where
 {
     debug_assert!(align != 0u8.into(), "Align must be non-zero");
     (value + align - 1u8.into()) & !(align - 1u8.into())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{find_offset, section_read_size};
+    use crate::pe::options::ParseOptions;
+    use crate::pe::section_table::SectionTable;
+
+    fn section(
+        virtual_address: u32,
+        virtual_size: u32,
+        size_of_raw_data: u32,
+        pointer_to_raw_data: u32,
+    ) -> SectionTable {
+        SectionTable {
+            virtual_address,
+            virtual_size,
+            size_of_raw_data,
+            pointer_to_raw_data,
+            ..SectionTable::default()
+        }
+    }
+
+    #[test]
+    fn virtual_only_sections_are_not_file_backed() {
+        let section = section(0x1fb000, 0x28590, 0, 0);
+        assert_eq!(section_read_size(&section, 0x200), 0);
+        assert_eq!(
+            find_offset(0x215580, &[section], 0x200, &ParseOptions::default()),
+            None,
+        );
+    }
+
+    #[test]
+    fn file_backed_sections_still_map_rvas() {
+        let section = section(0x1000, 0x600, 0x400, 0x200);
+        assert_eq!(section_read_size(&section, 0x200), 0x400);
+        assert_eq!(
+            find_offset(0x1123, &[section], 0x200, &ParseOptions::default()),
+            Some(0x323),
+        );
+    }
 }
--- a/src/pe/resource.rs
+++ b/src/pe/resource.rs
@@ -1967,9 +1967,8 @@ mod tests {
     }
 
     #[test]
-    #[should_panic = "Cycle detected in resource directory at offset"]
     fn malformed_resource_tree() {
-        let _ = crate::pe::PE::parse(MALFORMED_RESOURCE_TREE).unwrap();
+        assert!(crate::pe::PE::parse(MALFORMED_RESOURCE_TREE).is_err());
     }
 
     #[test]


src/pe/mod.rs
```diff
@@ -365,15 +365,21 @@ impl<'a> PE<'a> {
...
+                .map(Some)
+                .or_permissive_and_default(
+                    opts.parse_mode.is_permissive(),
+                    "Failed to parse resource data",
                 )?;
-                resource_data = Some(data);
-                debug!("resource_data data: {:#?}", data.version_info);
+                if let Some(ref data) = resource_data {
+                    debug!("resource_data data: {:#?}", data.version_info);
+                }
```
m4b: sorry I merged another (earlier) PR that had a similar change, would you mind to rebase?

src/pe/utils.rs
m4b, line 1: @kkent030315  could you review changes in this file? I generally defer to you on PE stuff :)

m4b:
@supervacuus just so I fully understand, the other PR that adds permissive parsing to the source directories that I've just merged (and caused a merge conflict with the similar changes in this PR), does this fix the underlying perf issue you were seeing, or does it require the RVA fixes as well?

supervacuus:
> does this fix the underlying perf issue you were seeing, or does it require the RVA fixes as well?

No, they are independent.

Both PRs can make sense independently. But I proposed the other PR due to performance issues this PR would resolve. It might still make sense to turn off imports parsing entirely if this PR lands, but having the option to turn it off would be less relevant for the downstream use cases in `symbolic` or `sentry-cli`.

supervacuus:
Oh, I am sorry... I thought you meant https://github.com/m4b/goblin/pull/522, but you were talking about https://github.com/m4b/goblin/pull/526, right?

No, while https://github.com/m4b/goblin/pull/526 also adds the `is_permissive()` check as I did in this PR, the essential change for the [performance fix is in `src/pe/utils.rs`](https://github.com/m4b/goblin/pull/527/changes#diff-2863ff71bf5f3add3a1bf0b981ddf6b713d1323dd5babd11350506f17b8eb900).

src/pe/utils.rs
```diff
@@ -65,8 +73,6 @@ fn section_read_size(section: &section_table::SectionTable, file_alignment: u32)
 
     if virtual_size == 0 {
         read_size
-    } else if read_size == 0 {
```
m4b, line 68: The only question I have is regressions in other cases that might happen because this check was removed; to be fair, I can't remember why it was added, and there's no comment explaining the logic here, so if this fixes a big bug for you, and all our tests pass, I'm biased towards just merging.

src/pe/mod.rs
```diff
@@ -379,7 +379,7 @@ impl<'a> PE<'a> {
             };
 
             if let Some(&resource_table) = optional_header.data_directories.get_resource_table() {
-                resource_data = resource::ResourceData::parse_with_opts(
+                resource_data = ResourceData::parse_with_opts(
```
m4b: nit: diff not strictly necessary

src/pe/mod.rs
```diff
@@ -391,7 +391,9 @@ impl<'a> PE<'a> {
                     opts.parse_mode.is_permissive(),
                     "Failed to parse resource data",
                 )?;
-                debug!("resource_data data: {:#?}", resource_data);
+                if let Some(ref data) = resource_data {
```
m4b: nit: it might be nice to log it's none but it's fine I guess

m4b (approved):
LGTM, thanks for your patience! I'd like @kkent030315 to review if possible the pe::utils stuff but seems ready to go to me otherwise

src/pe/mod.rs
supervacuus: I am happy to remove my changes in this file, since master already reflects them

src/pe/utils.rs
supervacuus: The actual fix is the early exit on `size_of_raw_data == 0`, which prevents this fallback to `virtual_size`. I see no other way in which `read_size` could be 0 now. That is why I removed the check, because it tests a case that should not exist once we exclude `size_of_raw_data == 0`. But maybe I misunderstood.

src/pe/mod.rs
supervacuus: I rebased and removed any of my changes to [src/pe/mod.rs](https://github.com/m4b/goblin/pull/527/files#diff-6303a321bf22fd69f034c1622e2b63c543fb1ceb6f3828e3dd81403e8fff4942)

src/pe/utils.rs
philipc: This check was added in #396. It was added to fix parsing of the file in https://github.com/ideeockus/goblin/blob/f504af9de4c86bdf697ceb6876f2683c0226886f/tests/bins/pe/exception_rva_mapping/amdcleanuputility.exe.upx_packed, which was giving the same error as #307. That error reappears with this PR regardless of whether this check is removed or not (the earlier early return is enough to cause it). I think the change in this PR is fine, and #396 should be fixed differently (use the permissive parsing option).

supervacuus:
Did you find a conclusion regarding this change and #522?

m4b:
@supervacuus we can merge this if it doesn't regress anything and fixes something for you.

supervacuus:
> @supervacuus we can merge this if it doesn't regress anything and fixes something for you.

I am fine with either solution, since in the downstream uses, we don't need the imports (yet). So having an option to turn off the import parser is sufficient.

I see it like this: the import parser option is the safer bet in terms of expected regressions, but also a much coarser solution. However, it might also make sense for uses other than the concrete performance hit we see downstream. 

This PR otoh, resolves the root cause of the actual performance issue and should likely land independently of the option from #522. As @philipc raised in https://github.com/m4b/goblin/pull/527#discussion_r3109977607, we at least know of one regression that can be rectified by the permissive parsing mode introduced since the fix.

m4b:
ok thanks for the summary; it sounds like for your usecases you'll probably use the options introduced in 522, but this is also a nice perf fix if someone uses the defaults (which is to parse the perf table).

It sounds like there isn't any danger to merge this in terms of regressions of default parser, so let's merge it; if we encounter something we can always revert since it's backwards compatible.

Thank you for your patience!

m4b:
NB: backwards compatible

m4b:
released in 0.10..6

PR #529, opened 2026-05-25
fix(elf): terminate NoteDataIterator after parse error to prevent infinite loop

### LINKED ISSUE #108: Elf: NoteIterator has an infinite loop
When parsing the multiarray.so binary mentioned in #106, the NoteIterator appears to infinitely loop if we do something like:

```rust
        if let Some(notes) = self.elf.iter_note_headers(self.bytes) {
            fmt_hdr(fmt, "Notes")?;
            writeln!(fmt, "")?;
            for (i, note) in notes.enumerate() {
                if let Ok(note) = note {
                    fmt_idx(fmt, i)?;
                    write!(fmt, " ")?;
                    fmt_str(fmt, note.name.trim_right_matches('\0'))?; // REMOVEME: hotfix for goblin 0.12 notes including \0
                    write!(fmt, " type: {} ", note.type_to_str())?;
                    for byte in note.desc {
                        write!(fmt, "{:x}", byte)?;
                    }
                    writeln!(fmt, "")?;
                } else {
                    writeln!(fmt, "BAD NOTE: {:?}", note);
                }
            }
            writeln!(fmt, "")?;
        }
```

SAY-5:
Fixes #108.

`NoteDataIterator::next` calls `gread_with(&mut self.offset, ...)`. When
parsing fails, the scroll crate does not advance `self.offset` (it only
advances on success). The error arm returned `Some(Err(e))` without
updating `self.offset`, so the very next call to `next` sees the same
offset, parses the same bytes, hits the same error, and returns another
`Some(Err(e))` -- indefinitely. Callers like `.count()` or a plain
`for` loop that ignore the error arm never see `None` and hang forever.

The fix: set `self.offset = self.size` before returning the error, which
causes the termination check `self.offset >= self.size` to fire on the
very next call and return `None`.

A new unit test `error_terminates_iterator` verifies:
1. Parsing truncated data yields `Some(Err(_))` on the first call.
2. The second call returns `None` instead of repeating the error.

All existing tests continue to pass (`cargo test`).

### DIFF
--- a/src/elf/note.rs
+++ b/src/elf/note.rs
@@ -114,7 +114,13 @@ if_alloc! {
                 debug!("NoteIterator - {:#x}", self.offset);
                 match self.data.gread_with(&mut self.offset, self.ctx) {
                     Ok(res) => Some(Ok(res)),
-                    Err(e) => Some(Err(e))
+                    Err(e) => {
+                        // Terminate the iterator on parse error; without this,
+                        // offset stays unchanged and the same position is parsed
+                        // again on the next call, causing an infinite loop.
+                        self.offset = self.size;
+                        Some(Err(e))
+                    }
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


m4b (approved):
Ok thank you; i wonder if we need to do a survey of other cases like this

PR #530, opened 2026-05-25
fix(pe): represent ordinal imports with name: None

### LINKED ISSUE #38: EASY PE: should Import.name be option?
```
            Import {
                name: "ORDINAL 0",
                dll: "WS2_32.dll",
                ordinal: 0,
                offset: 62264,
                rva: 0,
                size: 4
            },
```

this ("ORDINAL 0") is ~~basically~~ a stupid hack, but it makes working with Import much nicer. Alternatively, could make an enum, but, i dunno that just annoys me for some reason

wyf027:
## Summary

Fixes #38 by changing PE `Import.name` from `Cow<str>` to `Option<Cow<str>>`.

Ordinal-only imports no longer synthesize placeholder strings like `"ORDINAL 51398"`. Instead, `name` is `None` and the `ordinal` field remains authoritative.

## Breaking change

This is an API change for consumers that matched on the `"ORDINAL {n}"` string convention. They should now check `name.is_none()` and use `ordinal`.

## Test plan

- [x] `cargo test -p goblin --lib pe::import::tests`

### DIFF
--- a/src/pe/import.rs
+++ b/src/pe/import.rs
@@ -510,7 +510,7 @@ impl<'a> ImportData<'a> {
 #[derive(Debug)]
 /// A synthesized symbol import, the name is pre-indexed, and the binary offset is computed, as well as which dll it belongs to
 pub struct Import<'a> {
-    pub name: Cow<'a, str>,
+    pub name: Option<Cow<'a, str>>,
     pub dll: &'a str,
     pub ordinal: u16,
     pub offset: usize,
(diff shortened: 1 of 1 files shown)


m4b:
I guess this makes sense, I'd probably design it this way if I did it right now, but could you give some motivation for this PR? The ordinal is always "authoritative" afaik. Is there some usecase where you want to know if it truly didn't have an ordinal name (e.g., it was None)?

kkent030315:
@m4b this is BC right. so if we do this in future major bump, I believe we could do some major refactor for import table parser along with this?

m4b:
@kkent030315 yes, what other changes did you have in mind?

wyf027:
Resolved the rustfmt CI failure in commit bef45f7 by applying the formatter to src/pe/import.rs. Local verification: cargo fmt -- --check.

PR #532, opened 2026-05-25
fix(pe): skip indirect section names for PE images

### LINKED ISSUE #487: PE/COFF: Do not try to parse indirect section name in PE
Current version of goblin parses section names as indirect even on PE. The indirect section name is only a feature in COFF so we should avoid parsing it when it's PE and instead parse the name in UTF-8.

https://github.com/m4b/goblin/blob/baae69e976fedc8d8ad850d813a9ff0973de2f60/src/pe/section_table.rs#L113-L117

wyf027:
## Summary

- Add `parse_indirect_section_names` to `ParseOptions` (default: `false` for PE)
- Only resolve `/offset` and `//base64` COFF section name references when parsing COFF object files
- Enable indirect section name parsing for `Coff::parse` via `ParseOptions::coff()`

Fixes #487

## Test plan

- [x] `cargo test --lib pe::section_table::tests`
- [x] `cargo test --lib string_table_excludes_length`
- [x] New tests for PE literal `/1234567` names and COFF indirect parsing

### DIFF
--- a/src/pe/mod.rs
+++ b/src/pe/mod.rs
@@ -699,7 +699,7 @@ impl<'a> Coff<'a> {
         debug!("{:#?}", header);
         // TODO: maybe parse optional header, but it isn't present for Windows.
         *offset += header.size_of_optional_header as usize;
-        let sections = header.sections(bytes, offset)?;
+        let sections = header.sections_with_opts(bytes, offset, &crate::pe::options::ParseOptions::coff())?;
         let symbols = header.symbols(bytes)?;
         let strings = header.strings(bytes)?;
         Ok(Coff {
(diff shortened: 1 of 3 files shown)


kkent030315:
Thanks for the PR!

Relevant issue: #487 

I think we need some discussion whether we really need this? I like this, but don't know if we should go this direction.

wyf027:
Resolved the merge conflict with upstream master in commit db586c8 and applied rustfmt to the PE COFF section parsing line.

During the merge, upstream added `parse_imports` while this branch added `parse_indirect_section_names`; both options are preserved. I also updated the PE indirect-section-name integration test to match the branch semantics: PE images keep `/512` as the literal section name by default, while COFF-specific indirect parsing remains covered through `ParseOptions::coff()`.

Local verification:
- `cargo fmt -- --check`
- `cargo build`
- `cargo test`

PR #533, opened 2026-05-25
fix(mach): detect shebang scripts before Mach-O parsing

### LINKED ISSUE #494: Invalid magic number for many mach-o executables.
Running `Mach::parse` against all of the executables on my mac results in the following counts of errors:

```
  12 Invalid magic number: 0x2321202f
   8 Invalid magic number: 0x23212f53
  92 Invalid magic number: 0x23212f62
 233 Invalid magic number: 0x23212f75
```

Which leads me to believe that some of these magic numbers are just new not invalid, and might ought to be supported.

Here's my code, it's pretty simple so I don't think I did anything egregious. The errors above come from the 4th line from the end, where I print the error message from `Mach::parse`.

```rust
use crate::Mach::{Binary, Fat};
use clap::Parser;
use goblin::mach::{Mach, MachO, MultiArch, SingleArch};
use std::fs;
use std::path::Path;
use std::ffi::os_str::OsStr;

#[derive(Parser, Debug)]
#[command(version, about, long_about = None)]
struct Args {
    binaries: Vec<String>,
}

fn process_multi<'a>(multi: MultiArch<'a>) -> Vec<([u8; 16], [u8; 16])> {
    multi
        .into_iter()
        .filter_map(|res| res.ok())
        .filter_map(|s| match s {
            SingleArch::MachO(m) => Some(m),
            _ => {eprintln!("Passed archive not binary"); None}
        })
        .flat_map(|m| process_macho(m).into_iter())
        .collect()
}

fn process_macho<'a>(m: MachO<'a>) -> Vec<([u8; 16], [u8; 16])> {
    m.segments
        .sections()
        .flatten()
        .filter_map(|res| res.ok())
        .map(|s| (s.0.segname, s.0.sectname))
        .collect()
}

fn main() {
    let args = Args::parse();

    args.binaries
        .iter()
        .map(Path::new)
        .flat_map(fs::read)
        .flat_map(|bytes| match Mach::parse(&bytes) {
            Ok(Fat(m)) => process_multi(m),
            Ok(Binary(b)) => process_macho(b),
            Err(err) => { eprintln!("{}", err); Vec::new() },
        })
        .for_each(|t| unsafe { println!("{},{}",OsStr::from_encoded_bytes_unchecked(&t.0).display(),OsStr::from_encoded_bytes_unchecked(&t.1).display()) });
}
```

wyf027:
## Summary

- When scanning macOS directories, many non-Mach-O files are shell scripts whose first four bytes look like invalid Mach magic (e.g. `0x23212f75` for `#!/u`).
- Detect inputs starting with `#!` early in `Mach::parse` and return a clear `Malformed` error instead of `Invalid magic number`.

Fixes #494

## Test plan

- [x] `cargo test --lib mach::test::parse_rejects_shebang_scripts`
- [x] Existing Mach-O fat binary/archive parse tests still pass

### DIFF
--- a/src/mach/mod.rs
+++ b/src/mach/mod.rs
@@ -549,6 +549,11 @@ impl<'a> Mach<'a> {
             let error = error::Error::Malformed("size is smaller than a magical number".into());
             return Err(error);
         }
+        if bytes.starts_with(b"#!") {
+            return Err(error::Error::Malformed(
+                "input appears to be a script with a shebang (#!), not a Mach-O binary".into(),
+            ));
+        }
         let magic = peek(&bytes, 0)?;
         match magic {
             fat::FAT_MAGIC => {
@@ -567,6 +572,7 @@ impl<'a> Mach<'a> {
 #[cfg(test)]
 mod test {
     use super::{Mach, SingleArch};
+    use crate::error::Error;
 
     #[test]
     fn parse_multi_arch_of_macho_binaries() {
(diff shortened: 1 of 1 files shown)


m4b (approved):
thank you

m4b:
NB: backwards compatible

PR #534, opened 2026-05-25
fix(elf): detect imports via SHN_UNDEF instead of st_value (fixes #419)

### LINKED ISSUE #419: `is_import` is wrong on MIPS ELF

wyf027:
## Summary

- Change `Sym::is_import()` to treat symbols with `st_shndx == SHN_UNDEF` as imports instead of relying on `st_value == 0`.
- Update the standalone `is_import()` helper to take `shndx` rather than `st_value`.
- Add unit tests covering MIPS-style undefined globals with non-zero `st_value`.

Fixes #419.

On MIPS ELF, undefined dynamic symbols (e.g. `memcpy@GLIBC_2.0`) can have a non-zero `st_value` while still being imports. The previous heuristic incorrectly classified them as defined exports.

As suggested by @philipc in the issue, checking `st_shndx == SHN_UNDEF` is the reliable indicator.

## Test plan

- [x] `cargo test is_import`
- [x] `cargo test --lib`

### DIFF
--- a/src/elf/sym.rs
+++ b/src/elf/sym.rs
@@ -94,9 +94,9 @@ pub fn st_visibility(other: u8) -> u8 {
 
 /// Is this information defining an import?
 #[inline]
-pub fn is_import(info: u8, value: u64) -> bool {
+pub fn is_import(info: u8, shndx: u16) -> bool {
     let bind = st_bind(info);
-    bind == STB_GLOBAL && value == 0
+    (bind == STB_GLOBAL || bind == STB_WEAK) && shndx == crate::elf::section_header::SHN_UNDEF as u16
 }
 
 /// Convenience function to get the &'static str type from the symbols `st_info`.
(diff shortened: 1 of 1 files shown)


wyf027:
Resolved the rustfmt CI failure in commit 2b42a39 by applying the formatter to src/elf/sym.rs. Local verification: cargo fmt -- --check.

PR #538, opened 2026-05-26
pe: validate offset and size in exception data

Dav1dde:
It's possible that `pe.exception_data.functions()` panics because of an out of bounds error. Size and offset are never validated against the total available data.

This adds a check in the constructor which validates size and offset to make the panic in `functions()` impossible.

We did observe a few of these panics in the wild: https://sentry-s4s2.sentry.io/share/issue/acd5bd4bf9c5417c809ba5f7bbccd747/


m4b (approved):
Ok, looks good to me, seems reasonable! cc @kkent030315

PR #541, opened 2026-06-17
pe: skip raw writes for virtual-only sections

mq1n:
## Summary

- treat PE sections with `SizeOfRawData == 0` as virtual-only on disk
- preserve virtual-only section table entries while skipping nonexistent raw bytes during PE rewrite
- add synthetic regression coverage for the writer path without requiring an executable fixture

## Root cause

Some PE files, including the local Free Pascal sample used to reproduce this, contain a `.bss` section with a nonzero virtual size but no raw on-disk data. The writer previously expanded that virtual-only section into zero bytes and wrote them at `PointerToRawData == 0`, which could corrupt the DOS/PE headers.

## Validation

- `cargo fmt --check`
- `cargo test`
- focused virtual-only section regression tests
- touched-file strict Clippy filter
- local `freepascal_cli.exe` rewrite remained byte-identical by SHA256

## Notes

This is a narrow PE writer fix and does not special-case Free Pascal binaries. The regression test uses synthetic section data instead of adding a binary executable fixture.

### DIFF
--- a/src/pe/mod.rs
+++ b/src/pe/mod.rs
@@ -458,7 +458,13 @@ impl<'a> PE<'a> {
         );
 
         for section in &self.sections {
-            let section_data = section.data(&self.bytes)?.ok_or_else(|| {
+            if section.size_of_raw_data == 0 {
+                // Virtual-only sections, such as .bss, have no on-disk bytes to write.
+                bytes.gwrite_with(section, offset, ctx)?;
+                continue;
+            }
+
+            let section_data = section.data(self.bytes)?.ok_or_else(|| {
                 error::Error::Malformed(format!(
                     "Section data `{}` is malformed",
                     section.name().unwrap_or("unknown name")
@@ -865,7 +871,7 @@ mod tests {
 
     #[test]
     fn string_table_excludes_length() {
-        let coff = Coff::parse(&&COFF_FILE_SINGLE_STRING_IN_STRING_TABLE[..]).unwrap();
+        let coff = Coff::parse(&COFF_FILE_SINGLE_STRING_IN_STRING_TABLE[..]).unwrap();
         let string_table = coff.strings.unwrap().to_vec().unwrap();
 
         assert!(string_table == vec!["ExitProcess"]);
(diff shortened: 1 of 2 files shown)


m4b:
@kkent030315 if you could review would appreciate :)

src/pe/mod.rs
```diff
@@ -947,4 +953,79 @@ mod tests {
...
+            load_config_data: None,
+            certificates: Vec::new(),
+            resource_data: None,
+            clr_data: None,
+        };
+
+        let mut output = vec![HEADER_SENTINEL; SECTION_TABLE_OFFSET + 0x80];
+        let mut section_table_offset = SECTION_TABLE_OFFSET;
+
+        pe.write_sections(&mut output, &mut section_table_offset, None, scroll::LE)
```
m4b, line 1014: it's also generally a good idea for a test to "round trip", e.g, you write a PE file, and then we can parse it back again

src/pe/section_table.rs
```diff
@@ -100,28 +100,36 @@ impl SectionTable {
...
 
     pub fn data<'b>(&self, pe_bytes: &'b [u8]) -> error::Result<Option<Cow<'b, [u8]>>> {
+        // Sections with no raw data are virtual-only in an on-disk PE.
+        if self.size_of_raw_data == 0 {
+            return Ok(None);
+        }
+
         let section_start: usize = self.pointer_to_raw_data.try_into().map_err(|_| {
-            Error::Malformed(format!("Virtual address cannot fit in platform `usize`"))
+            Error::Malformed("Virtual address cannot fit in platform `usize`".to_string())
```
m4b, line 109: generally this is nice but it is technically diff noise (and not related to your PR)

src/pe/section_table.rs
```diff
@@ -344,15 +352,30 @@ mod tests {
...
-            let mut section = SectionTable::default();
-            section.pointer_to_raw_data = 4;
-            section.size_of_raw_data = 4;
-            section.virtual_size = 4;
+            let section = SectionTable {
+                pointer_to_raw_data: 4,
+                size_of_raw_data: 4,
+                virtual_size: 4,
+                ..SectionTable::default()
+            };
```
m4b, line 360: also nice, but diff noise

src/pe/section_table.rs
```diff
@@ -100,28 +100,36 @@ impl SectionTable {
...
-            Error::Malformed(format!("Virtual address cannot fit in platform `usize`"))
+            Error::Malformed("Virtual address cannot fit in platform `usize`".to_string())
         })?;
 
         // assert!(self.virtual_size <= self.size_of_raw_data);
         // if vsize > size_of_raw_data, the section is zero padded.
         let section_end: usize = section_start
             + usize::try_from(self.size_of_raw_data).map_err(|_| {
-                Error::Malformed(format!("Virtual size cannot fit in platform `usize`"))
+                Error::Malformed("Virtual size cannot fit in platform `usize`".to_string())
```
m4b, line 116: ditto

src/pe/section_table.rs
```diff
@@ -100,28 +100,36 @@ impl SectionTable {
...
             })?;
 
-        let original_bytes = pe_bytes.get(section_start..section_end).map(Cow::Borrowed);
+        let original_bytes = match pe_bytes.get(section_start..section_end) {
+            Some(bytes) => bytes,
+            None => return Ok(None),
```
m4b, line 121: I think this changes the semantics; you now return none if we can't get the original bytes, but before it would return `Ok(original_bytes)` if the `is_some()` check did not pass. Is this intentional or refactor artifact? Could you give some motivation about why this change was made?

m4b (commented):
ah ok this is for the writer

